PR #5, opened 2023-03-30
Maintain the Current OCI Version Label

Zindswini:
I would like to bring this change into Bazzite, I'm experiencing the discover update loop.

Hopefully I'm interpreting what is needed correctly.

Thanks!

image


KyleGospo:
Much appreciated, thank you!

PR #11, opened 2023-05-23
Enable virtual keyboard at login

EyeCantCU:
Useful for devices with touchscreens like the Steam Deck. Provides a mode of input at login when a wired keyboard isn't available


KyleGospo:
Thank you very much for this!

Issue #13, opened 2023-05-26
Nvidia kernel arguments should be sidestepped on non Nvidia hardware


EyeCantCU:
Using the default kernel arguments the Steam Deck fails to properly load amdgpu

Removing Nvidia related arguments as well as nomodeset addresses this allowing amdgpu to load successfully

KyleGospo:
May also be worth separating the nvidia builds into their own -nvidia variant, if only to save space in AMD configurations.

EyeCantCU:
I can certainly do that sometime this week :)

PR #16, opened 2023-06-08
feat: Separate Nvidia images

EyeCantCU:
Builds separate images for Nvidia so that users on non-Nvidia hardware don't have to adjust kargs

Resolves: https://github.com/ublue-os/bazzite/issues/13


EyeCantCU:
Marked as a draft as this would require migration for current users

PR #17, opened 2023-06-18
Overhaul images

EyeCantCU:
These changes effectively make the structure similar to bluefin.

bazzite now serves as a base for desktop configurations and now bazzite-deck has been introduced to add changes specific to the Steam Deck and Steam Deck-alike configurations. Also:

- Introduce images built from main for non-Nvidia systems
- Consolidate structure (remove duplicates, merge Containerfiles)

Also resolves: https://github.com/ublue-os/bazzite/issues/13


KyleGospo:
Python3-hid is built in our copr repo, which says there should be a Fedora 38 build but just in case [I went ahead and kicked off one](https://copr.fedorainfracloud.org/coprs/kylegospo/bazzite/build/6093919/). Can we revert the part of this PR that removes the btrfs package due to this dependency? That's going to be needed for proper microSD card support on the deck.

EyeCantCU:
Yeah, I can certainly do that

PR #25, opened 2023-07-03
feat(steamos-update): Initial support for Universal Blue updater

EyeCantCU:
This implements ublue-update in steamos-update, replacing the usage of rpm-ostree. This will allow for not only system updates, but flatpak updates and distrobox updates as well


EyeCantCU:
Ideally, checking for update availability is something I'm going to be able to move over to ublue-update instead of leaving it here

PR #28, opened 2023-07-05
Add Waydroid

KyleGospo:
**Not a blocker for 1.0**

Add Waydroid for Android gaming on Bazzite. For the Deck variant we'll need a compositor for the desktop since it's x11 only, but it should work OOTB in Wayland KDE and Gamescope thanks to Valve adding support for wayland clients.


KyleGospo:
https://github.com/casualsnek/waydroid_script

Something to consider

EyeCantCU:
> https://github.com/casualsnek/waydroid_script
> 
> Something to consider

This would be awesome to incorporate

KyleGospo:
One decision to make, do we use weston or kwin as the nested wayland compositor in X11 desktop? Weston seems more common and would support a potential GNOME build but kwin is of course built in for us ATM

sekalengrengginang:
> https://github.com/casualsnek/waydroid_script
> 
> Something to consider

afaik there is issue with waydroid script if you tried to add arm64 android apps support

EyeCantCU:
> > https://github.com/casualsnek/waydroid_script
> > Something to consider
> 
> afaik there is issue with waydroid script if you tried to add arm64 android apps support

I can test for that and see if that's the case. Thank you for mentioning this

EyeCantCU:
> One decision to make, do we use weston or kwin as the nested wayland compositor in X11 desktop? Weston seems more common and would support a potential GNOME build but kwin is of course built in for us ATM

I was actually thinking about how we would handle Gnome builds this morning. I think we should use weston for this

KyleGospo:
Moved this to the non-Nvidia section since Waydroid cannot make use of Nvidia GPUs at all.

KyleGospo:
Waydroid and related packages are on the main branch now and usable, this PR will instead be for helper scripts to get it working on X11 and for work in Yafti to make it easy to setup.

KyleGospo:
Moved the just command over too so this PR has no reason to exist, will open a new one when we can start on a real Yafti screen.

PR #39, opened 2023-07-15
feat: Implement GNOME images

EyeCantCU:
These changes provide initial support for GNOME Bazzite images


EyeCantCU:
Caveats: The Deck Gnome images use Wayland; not X11 which could be problematic for input. The just files haven't been updated yet and neither has the yafti configuration so I'd imagine quite a bit of that is no-op at the moment

EyeCantCU:
All aforementioned issues should be worked out

KyleGospo:
> All aforementioned issues should be worked out

What are your thoughts on getting these to start building? If it's not ready for a 1.0 we can always just not advertise these

KyleGospo (approved):

EyeCantCU (commented):
Everything here looks fantastic and ready to go!

KyleGospo:
LGTM

PR #41, opened 2023-07-16
feat: Implement sentry's kernel-fsync

EyeCantCU:
Includes support for the Steam Deck, Surface lineup, and Asus devices alongside numerous other useful patches


EyeCantCU:
We're going to use akmods for this

PR #47, opened 2023-07-18
Update yafti.yml with way too many options

nicknamenamenick:
I added way too much stuff and got a little carried away.  In a way I probably should have thought about storage usage and use cases for this image.  I do not mind if most of these are not accepted, but if you see any here that may be interesting then here are my suggestions.

Other applications I did not add for the following reasons:

- Desktop Files Creator (GUI to create .desktop files), felt it was a little out of scope but if anyone wants to look into for Utilities it could be fine?

- Audacity, replaced it with Tenacity

- Brasero (CD ripping and burning tool), no Flatpak. There are alternative applications on Flathub but I have no idea how well they work.  I also don't know how many people still have CD players in their PCs.

- Piper, needs ratbagd.

- Peazip, Ark is preinstalled, but there are situations where this might be needed.

- PowerISO, proprietary and not many scenarios where you need to mess with ISO files, but there are some reasonable instances where this would come in handy and be easier than any command line tools.

- Vivaldi Web Browser, no Flatpak.  Distrobox could work, and I only would suggest this because it is relatively popular.  No reason to go through the hassle of figuring out a container for this when most people would opt for the available web browsers on Flathub.

### DIFF
--- a/system_files/desktop/usr/share/ublue-os/firstboot/yafti.yml
+++ b/system_files/desktop/usr/share/ublue-os/firstboot/yafti.yml
@@ -22,7 +22,7 @@ screens:
           condition:
             run: distrobox list | grep -v bazzite-arch
           packages:
-          - Install Bazzite Arch: just --unstable install-bazzite-arch
+          - Install Bazzite Arch: just install-bazzite-arch
           - Export Steam: distrobox-enter -n bazzite-arch -- '  distrobox-export --app steam'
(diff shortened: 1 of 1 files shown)


KyleGospo (approved):

KyleGospo:
Due to this being made in-between some major changes to Yafti I'm closing this, but I've merged a majority of these changes in https://github.com/ublue-os/bazzite/commit/abed6f86df14c3c4abb2bb370018f2f1333f84d4

Thanks!

PR #76, opened 2023-07-27
fix(power-button-handler): Rename 'inhibit' string to 'systemd-inhibit'

EyeCantCU:
subprocess.call() improperly passes the inhibit function otherwise preventing the handler from ever starting in the first place


KyleGospo:
🐐

PR #80, opened 2023-07-27
feat(desktop): Add Easy Effects to installable utilities

EyeCantCU:
Easy Effects provides an equalizer, compressor, limiter, and reverb tool for the manipulation of audio streams


KyleGospo:
This is great, but we may want to pull it from the deck image due to it not working in game mode

KyleGospo:
Could instead offer a parabolic EQ via pipewire for the deck builds

EyeCantCU:
That sounds good to me :)

KyleGospo (approved):

PR #81, opened 2023-07-27
feat: Set SteamOS BTRFS mount flags for all BTRFS partitions

EyeCantCU:
Adds a script to just and entry to yafti setting the mount flags used by SteamOS BTRFS for all BTRFS drives in fstab

### DIFF
--- a/system_files/desktop/usr/share/ublue-os/just/custom.just
+++ b/system_files/desktop/usr/share/ublue-os/just/custom.just
@@ -46,6 +46,9 @@ enable-wallpaper-engine:
   plasmapkg2 -i /tmp/wallpaper-engine-kde-plugin/plugin
   rm -rf /tmp/wallpaper-engine-kde-plugin
 
+set-btrfs-flags:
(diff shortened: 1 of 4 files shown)


system_files/desktop/usr/share/ublue-os/just/custom.just
```diff
@@ -96,6 +96,9 @@ resize-zram:
     just --unstable resize-zram
   fi
 
+set-btrfs-flags:
+  sudo sed -i 's/compress=zstd:1/defaults,nofail,x-systemd.growfs,noatime,lazytime,commit=120,compress-force=zstd:1,space_cache=v2,discard=async/g' /etc/fstab
```
KyleGospo, line 100: Good for SteamOS, but we don't need some of these. I recommend:
`sudo sed -i 's/compress=zstd:1/noatime,lazytime,commit=120,compress-force=zstd:1,space_cache=v2,discard=async/g' /etc/fstab`

EyeCantCU: Awesome. Sounds good to me. I'll adjust them

PR #85, opened 2023-07-28
chore: Remove duplicate f3 installation from Containerfile

EyeCantCU:
We have this marked as a dependency of jupiter-hw-support-btrfs here: https://github.com/ublue-os/bazzite/blob/71279c26b2d3583dc79f623d7503ec4af5c95da9/spec_files/jupiter-hw-support/jupiter-hw-support-btrfs.spec#L31


KyleGospo:
:flushed:

PR #123, opened 2023-08-08
feat: User first-run script

KyleGospo:
Removes must-run parts from Yafti and runs them for every user's first login.


EyeCantCU (approved):
This is so much better! :)

PR #143, opened 2023-08-14
chore(yafti): Rename to 'Bazzite Portal'

EyeCantCU:
Provides clearer naming for users


KyleGospo:
How about "Bazzite Configurator" or similar and I'll rename it in the GNOME extension as well

PR #148, opened 2023-08-15
fix: Enable waydroid for non-nvidia targets

Linwer:
Stumbled upon waydroid-container.service being disabled. I think the intention here was to not have the service enabled for Nvidia users and to enable it for everyone else, or maybe I'm missing something


KyleGospo:
Disabled by-default because not everyone's going to use it (and it has issues that break it with SELinux enabled).

Eventually yafti will enable this for you.

PR #151, opened 2023-08-16
feat: added boilr .desktop entry

Linwer:
GNOME users have long forgotten what a "~/Desktop" folder is, so I added a desktop entry for BoilR based on the Flatpak .desktop entry.

### DIFF
--- a/system_files/deck/shared/usr/share/ublue-os/just/custom.just
+++ b/system_files/deck/shared/usr/share/ublue-os/just/custom.just
@@ -66,6 +66,26 @@ get-boilr:
     jq -r ".assets[] | select(.name | test(\"linux_BoilR\")) | .browser_download_url") \
     -O ~/Desktop/BoilR
   chmod +x ~/Desktop/BoilR
+  for px in 64 128 256;\
+  do \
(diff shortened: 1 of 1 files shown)


KyleGospo:
This looks great, only request I have would be to put Boilr in ~/Applications and mkdir -p ~/Applications. We can also drop the creation of the desktop icon this way.

Linwer:
Idk, that would mean breaking the existing workflow that KDE users (or GNOME users with the desktop icons extension) have. Having the executable in ~/Desktop doesn't and won't really bother anyone, even if inelegant.

KyleGospo:
I'm also fine copying the .desktop file to the desktop, you can check for Kinoite before doing that as we do for others in the same file

Linwer:
something like this then?

KyleGospo:
Thank you!

system_files/deck/shared/usr/share/ublue-os/just/custom.just
```diff
@@ -61,11 +61,34 @@ get-greenlight:
   chmod +x ~/Desktop/Greenlight.AppImage
 
 get-boilr:
-  wget \
+  curl \
```
szescxz, line 64: Missed an `-L` here.

PR #156, opened 2023-08-16
chore(ci): Always use the latest version of isogenerator

EyeCantCU:
This will ensure we are always using the latest version of isogenerator


castrojo:
I would be careful with this, we pin on main for a reason. 😄

PR #162, opened 2023-08-18
feat: Add support for ROG Ally

EyeCantCU:
Adds support for the Asus ROG Ally

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -362,7 +362,13 @@ RUN wget https://copr.fedorainfracloud.org/coprs/lukenukem/asus-linux/repo/fedor
 RUN rpm-ostree override replace \
     --experimental \
     --from repo=copr:copr.fedorainfracloud.org:lukenukem:asus-kernel \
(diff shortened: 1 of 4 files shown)


Containerfile
KyleGospo, line 1: also remove vpower, & powerbuttond
sdgyrodsu can stay iirc, it supports ally

EyeCantCU: Done :)

KyleGospo:
Spoke with @bsherman about this, my understanding is we'll eventually be able to build kmods for this kernel in ublue-os/akmods and then we can get this shipped @EyeCantCU.

EyeCantCU:
> Spoke with @bsherman about this, my understanding is we'll eventually be able to build kmods for this kernel in ublue-os/akmods and then we can get this shipped @EyeCantCU.

Awesome! That would be perfect

KyleGospo:
https://github.com/ublue-os/asus

Follow here for further updates

PR #188, opened 2023-08-25
(fix) Make batterylimit.service work properly

RodoMa92:
It seems that you cannot use input redirection directly when running the command. Pass the command to bash instead seems to make it work correctly.


KyleGospo:
<3

PR #197, opened 2023-08-27
fix(desktop): boot scripts slowdown

tulilirockz:
Some scripts slow down the boot process when their services are enabled. The only fix for now is just to disable them, but with this PR it checks whether or not they already ran and skips them if so.
The problems is that if for some reason someone changes their hardware, those should be manually ran again by the user, so maybe there should be an option in Yafti or something like that for re-enabling these scripts.

### DIFF
--- a/system_files/desktop/shared/usr/lib/systemd/system/bazzite-hardware-setup.service
+++ b/system_files/desktop/shared/usr/lib/systemd/system/bazzite-hardware-setup.service
@@ -7,6 +7,7 @@ Before=systemd-user-sessions.service jupiter-biosupdate.service jupiter-controll
 Type=oneshot
 RemainAfterExit=yes
 ExecStart=/usr/bin/bazzite-hardware-setup
+ConditionPathExists=!/etc/bazzite/hardware_setup_done
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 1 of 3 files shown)


EyeCantCU:
We should check for these files in the services instead. It'll ensure the scripts never get triggered. One thing to keep in mind with the Flatpak script, the reason we don't disable this is because it's not always going to be successful for a number of reasons. If it fails, this file will get created and it'll never run again. I don't feel like that's optimal at all. It also prevents new Flatpaks we add in the future from ever being installed without user intervention

tulilirockz:
Done! That should be better now

system_files/desktop/shared/usr/lib/systemd/system/bazzite-hardware-setup.service
```diff
@@ -7,6 +7,7 @@ Before=systemd-user-sessions.service jupiter-biosupdate.service jupiter-controll
 Type=oneshot
 RemainAfterExit=yes
 ExecStart=/usr/bin/bazzite-hardware-setup
+ConditionPathExists=!/etc/bazzite/hardware_setup_done
```
EyeCantCU, line 10: I've been brainstorming a better place to put this but honestly I can't think of anywhere else it would belong

EyeCantCU (approved):

system_files/desktop/shared/usr/lib/systemd/system/bazzite-hardware-setup.service
tulilirockz: If someone eventually makes a standard "universal blue" configuration directory, we can move that check file onto it. For now I guess this would be the best?

PR #202, opened 2023-08-27
feat: Sign Bazzite images in yafti

EyeCantCU:
This adds a just script that signs the current image and adds an entry for it in yafti which gets hidden afterwards.


KyleGospo (approved):

akdev1l (approved):

EyeCantCU:
Deck image failures are due to dependency issues and not the content of this PR

PR #231, opened 2023-08-30
Added a help section to the justfile

dnkmmr69420:
Modify it to your needs

### Guidelines
In this project we follow [Semantic PRs][1] and [Conventional Commits][2] to describe features and fixes we made. It would be nice if you did too. Not sure what this all means? Here are some [examples][3]. Did you already use this? Awesome, thanks again!

[1]: https://github.com/Ezard/semantic-prs
[2]: https://www.conventionalcommits.org/en/v1.0.0/#summary
[3]: https://www.conventionalcommits.org/en/v1.0.0/#examples

### DIFF
--- a/system_files/desktop/shared/usr/share/ublue-os/just/custom.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/custom.just
@@ -1,7 +1,26 @@
+help:
+  echo "install-bazzite-arch: Creates a bazzite-arch distrobox container"
+  echo "configure-waydroid: Sets up waydroid"
+  echo "install-corectl: Installs corectl"
+  echo "install-nix: Installs the nix package manager using determinate systems installer"
+  echo "remove-nix: Uninstalls the nix package manager"
+  echo "enable-auto-login: Enables auto login"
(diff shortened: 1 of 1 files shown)


dnkmmr69420:
but how will people to know what commands are available that does not involve looking at the source code?

bigpod98:
Isnt there just list or something like that that lists just commands

bigpod98:
Its just --list but ok

castrojo:
`just` by itself shows the individual help snippets for each one.

image

dnkmmr69420:
I never knew that existed. I think there should be something on the very top because if you just type `just` it automatically removes arch-bazzite and reinstalls the container. The first thing should be like this

```
all:
      echo 'type 'justfile --list' for help
```

castrojo:
Mine works like it's supposed to, there's probably something specific in the bazzite custom one that is overriding it that we should probably fix.

KyleGospo:
Thanks, I used your work here to implement the same concept in this way:
https://github.com/ublue-os/bazzite/commit/74e9506a77414202fb534508e2cda3d252b8c88d

PR #236, opened 2023-08-30
fix(install-bazzite-arch): Use sh instead to get output in yafti

EyeCantCU:
Provides output in yafti... at the expense of the code looking ugly


KyleGospo:
Looks beautiful to me

EyeCantCU:
Looks && \
beautiful &&\
to && \
me

;)

PR #238, opened 2023-08-31
fix(deck): Fix stuttering volume in microphone

RodoMa92:
~~This still needs an additional manual fix using alsamixer, pushing up microphone volume to 100, then running sudo alsactl store -F once to permanently store the configuration of the audio card, otherwise by default the microphone hardware volume will lower it again back to 52 (which is barely audible even cranked to 100 from the audio panel menu).~~
Fixed by adding the saved state as a default configuration file.

To my hears this feels almost identical to SteamOS now.


KyleGospo:
Incredible work as always, thank you!

PR #241, opened 2023-08-31
feat(deck): Use winehq-staging instead of Fedora's wine-core.

KyleGospo:


KyleGospo:
@EyeCantCU Weird exit code here, you seen this before?
Also need to add winetricks back in before this lands

EyeCantCU:
> @EyeCantCU Weird exit code here, you seen this before?
> Also need to add winetricks back in before this lands

I actually have. A week or two ago I tried to layer Tabby (a terminal emulator) in my image and saw this. Tried working around it for a while but eventually gave up because I wasn't able to get a hold of anything useful

KyleGospo:
Very strange

PR #243, opened 2023-08-31
feat: Add option to install Fleek when installing Nix

KyleGospo:


KyleGospo:
@EyeCantCU can I get a review from you on this one when you get a moment?
Much appreciated!

system_files/desktop/shared/usr/share/ublue-os/just/custom.just
```diff
@@ -48,6 +48,12 @@ install-nix:
 remove-nix:
   sudo /nix/nix-installer uninstall
 
+install-fleek:
+  #!/usr/bin/env bash
+  mkdir -p $HOME/.config/nix
```
EyeCantCU, line 54: I recall $HOME not working when invoking the script with sudo -A just. I think we had to retrieve this some other way (like $SUDO_USER)

But otherwise everything here looks great. I can look more into the prior just to be sure

EyeCantCU: Ah. Found it

```export HOME=$(getent passwd ${SUDO_USER:-$USER} | cut -d: -f6)```

EyeCantCU (approved):

KyleGospo:
Thank you!

PR #244, opened 2023-08-31
Added GModCEFCodecFix  to justfile

dnkmmr69420:
It is this here: https://github.com/solsticegamestudios/GModCEFCodecFix I put it into the justfile


fiftydinar:
It would be nice if this is not hardcoded to a fixed date release

EyeCantCU (approved):
Looks great to me. Thank you so much for the submission

EyeCantCU:
I see what you mean. We can actually work around that

dnkmmr69420:
> It would be nice if this is not hardcoded to a fixed date release

we can easily edit the justfile again at least. maybe we can find a better way of doing this like having it source a file on a github repo that has the actual link then it will use it as a variable.

EyeCantCU (approved):
Yup. Got it worked out. Just uses curl to find the link to the latest file and throws the output to wget

PR #247, opened 2023-09-01
chore(greenlight): Bump version to 2.0.0-beta12

EyeCantCU:
As soon as 2.0.0 gets a release, we can just fetch the latest version from GitHub, but until then...


EyeCantCU:
Just to clarify, the reason we already don't just fetch the latest release is because the latest release is v1.2.0, which lacks a lot of the work that's been done lately (I.E. Steam Deck support)

PR #253, opened 2023-09-02
fix(just): Wait for any existing transactions to complete before invoking rpm-ostree

EyeCantCU:
Prevents rpm-ostree from failing when another transaction is in process by waiting for the current transaction to complete


EyeCantCU:
Ublue Update needs a release

KyleGospo (approved):

PR #272, opened 2023-09-07
added an option to enable flatpak theming with accent colors in the justfile

dnkmmr69420:

### DIFF
--- a/system_files/deck/shared/usr/share/ublue-os/just/custom.just
+++ b/system_files/deck/shared/usr/share/ublue-os/just/custom.just
@@ -360,3 +360,6 @@ _get-image:
     url = protocol + url
 
   print(url)
+
+enable-flatpak-theming:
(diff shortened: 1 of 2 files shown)


KyleGospo:
Is this needed for anything in Bazzite? Currently Gradience adds these rules when a theme is applied via Bazzite Portal.

EyeCantCU:
> Is this needed for anything in Bazzite? Currently Gradience adds these rules when a theme is applied via Bazzite Portal.

Could possibly be useful if someone doesn't apply one of the default themes, or sticks with Adwaita (and enables the dark mode for instance). I am curious why this is done on the user level and not system level though

KyleGospo:
That's fair, if that's the case let's also do
`~/.icons:ro`
`~/.themes:ro`

KyleGospo:
Addressed in latest commit, thanks!

PR #280, opened 2023-09-09
feat(hardware-setup): Enable amdgpu on supported systems

EyeCantCU:
Enable amdgpu on systems using the radeon driver. The arguments for Southern Islands and Sea Islands do not conflict so just enable both

Do not block this based on image flavor as some people may be using mixed GPU systems that have both an AMD and Nvidia card

### DIFF
--- a/system_files/desktop/shared/usr/bin/bazzite-hardware-setup
+++ b/system_files/desktop/shared/usr/bin/bazzite-hardware-setup
@@ -25,6 +25,7 @@ fi
 
 # GLOBAL
 SYS_ID="$(cat /sys/devices/virtual/dmi/id/product_name)"
+GPU_ID=$(lspci -k | grep -A 3 -E "(VGA|3D)")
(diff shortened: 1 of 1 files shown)


KyleGospo:
Think this would result in a loop since subsequent runs would have AMDGPU applied with the kargs in place, better if we just check for Radeon, then apply and forget since they can't hurt anything when applied.

EyeCantCU:
> Think this would result in a loop since subsequent runs would have AMDGPU applied with the kargs in place, better if we just check for Radeon, then apply and forget since they can't hurt anything when applied.

It should just confirm that amdgpu is in use and skip over it, but you're right. These shouldn't do any harm. I was just thinking for systems that go from main -> nvidia after changing cards, it would be nice to clean things up, but they can just sit there without disturbing anything. I'll remove the amdgpu check

KyleGospo (approved):

PR #287, opened 2023-09-11
Desktop cleanup

KyleGospo:
Since we now recommend HTPCs use the Deck image, we get to clean up the desktop images and bring them closer to upstream Fedora.

This issue will also cover some cleanup/fixes for Distrobox steam that may be needed.


dnkmmr69420:
yeah it needs to be closer to upstream fedora

PR #316, opened 2023-09-18
feat: Added virtualization setup in the justfile

dnkmmr69420:
Not much else to say


KyleGospo:
This is good, but I think we should move this to ublue-os/config, worth asking in the dev channel in Discord.

KyleGospo:
Merged via https://github.com/ublue-os/bazzite/pull/399

PR #323, opened 2023-09-19
feat(just): add waydroid-related utilities

tulilirockz:
Hello! I was setting up waydroid in my laptop and I thought that adding these waydroid utilities in just could be a nice addition :)


KyleGospo (approved):
This is awesome, thank you so much!

PR #327, opened 2023-09-19
fix(deck): Revert to copying Steam RPMs from GHCR hosted images

EyeCantCU:


dnkmmr69420:
`bazzite-deck` still failed to build

PR #330, opened 2023-09-19
chore: Install steam separately without problematic repos

KyleGospo:


EyeCantCU:
Since it looks like we need to install all dependencies for Steam anyway, we could walk through everything the spec does and throw it onto the image without installing the package. Not the cleanest approach though

Containerfile
```diff
@@ -341,18 +341,55 @@ RUN rpm-ostree install \
...
+        systemd-libs.i686 \
+        libva.i686 \
+        libvdpau.i686 \
+        libdbusmenu-gtk3.i686 \
+        libatomic.i686 \
+        pipewire-alsa.i686 && \
+    sed -i '0,/enabled=0/s//enabled=1/' /etc/yum.repos.d/rpmfusion-nonfree-steam.repo && \
+    sed -i '0,/enabled=1/s//enabled=0/' /etc/yum.repos.d/rpmfusion-nonfree.repo && \
+    sed -i '0,/enabled=1/s//enabled=0/' /etc/yum.repos.d/rpmfusion-nonfree-updates.repo && \
+    sed -i '0,/enabled=1/s//enabled=0/' /etc/yum.repos.d/rpmfusion-nonfree-updates.repo && \
```
szescxz, line 373: Duplicated 😂

PR #339, opened 2023-09-21
feat(just,yafti): webapp manager + openrazer and oversteer in distrobox

tulilirockz:
This PR adds the LinuxMint desktop manager to Yafti and the justfiles! These actions should set it up properly (i've tested them in my machine and they seem fine!)


KyleGospo:
I very much appreciate you doing this with distrobox vs just layering it.
My intention is to do the same for utilities like oversteer, openrazer, and solaar. Give me a bit to set up the shared infra for that and I'll land this.

tulilirockz:
@KyleGospo Hi! I've seen that you added the proper udev rules to ublue-os/config and this repo, so I decide to update this PR to include those addressed programs into one distrobox container, those being OpenRazer and Oversteer! :> I don't know how to integrate solaar into it so that's as far as I went

KyleGospo:
Pulled this into https://github.com/ublue-os/bazzite/pull/342

Thanks again!

PR #345, opened 2023-09-22
feat: Add image info for auto-signing via ublue-update

EyeCantCU:
We use image-info.sh to generate information about the given image at the time it's built. This replaces our current implementation of sourcing identifiers written to /etc/default/bazzite


KyleGospo:
Need to update this file as well:

https://github.com/KyleGospo/gamescope-session/blob/main/usr/share/gamescope-session/gamescope-session-script

KyleGospo:
COPY makes a new layer, might want to just make that part of the image and lose the copy command if you feel that won't cause issues.

EyeCantCU:
> Need to update this file as well:
> 
> https://github.com/KyleGospo/gamescope-session/blob/main/usr/share/gamescope-session/gamescope-session-script
Need to update this file as well:

Got it :) 

https://github.com/KyleGospo/gamescope-session/blob/main/usr/share/gamescope-session/gamescope-session-script

KyleGospo (approved):

PR #350, opened 2023-09-23
feat(kde): add xwaylandvideobridge on bazzite portal

sekalengrengginang:
add xwaylandvideobridge on bazzite portal for KDE desktop version


KyleGospo:
Addressed in https://github.com/ublue-os/bazzite/commit/db4bd53f7cba2fcaa59709ac1ccd9e1c42abb1eb

I'll add to Yafti soon, preferably in a KDE-only section so nobody on GNOME sees it.

PR #352, opened 2023-09-23
feat: Enable builds without kmods

EyeCantCU:


EyeCantCU:
I need to go through and add a check for nokmods to every place we check image_flavor for main or alternatively exempt nvidia instead of checking for main

PR #421, opened 2023-10-12
chore(f39): Restore kmods removed from main

EyeCantCU:
This restores kmods removed from Fedora 39


EyeCantCU:
Depends on:

https://github.com/ublue-os/akmods/pull/79

PR #424, opened 2023-10-12
feat: Looking glass shm OOB auto generation and support

HikariKnight:
This adds the config for systemd to to generate the looking-glass shm file if libvirt requests it and also provide a just command to set the correct selinux file context for `/dev/shm/looking-glass`.

this will result in any libvirt VM started with
```xml
<shmem name="looking-glass">
      <model type="ivshmem-plain"/>
      <size unit="M">X</size>
</shmem>
```
inside the xml to auto create /dev/shm/looking-glass when needed.
the user will just have to run `just selinux-looking-glass` once on their system to permanently enable it.

If you do not know about looking-glass it is a kvm framebuffer relay for gpu passthrough.
Users will still have to download and compile the client with the same release as the host they installed in the VM from https://looking-glass.io/ for it to work, this will have to be compiled in a fedora docker distrobox (the fedora toolbox distrobox does not work for this) and then copy the binary to the host `~/.local/bin` for now to avoid randomly breaking peoples looking-glass VMs whenever looking-glass updates (as the host application in the VM has no auto update function)

### DIFF
--- a/system_files/desktop/shared/usr/etc/tmpfiles.d/10-looking-glass.conf
+++ b/system_files/desktop/shared/usr/etc/tmpfiles.d/10-looking-glass.conf
@@ -1,0 +1,2 @@
+# Type Path               Mode UID  GID Age Argument
+f /dev/shm/looking-glass 0660 1000 qemu -
--- a/system_files/desktop/shared/usr/share/ublue-os/just/60-custom.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/60-custom.just
@@ -326,3 +326,7 @@ disable-watchdog:
     WATCHDOG_KARGS="$WATCHDOG_KARGS --append-if-missing=modprobe.blacklist=sp5100_tco"
   fi
   rpm-ostree kargs $WATCHDOG_KARGS
+
+# Add SELinux file context for default looking-glass shm file
+selinux-looking-glass:
+  sudo semanage fcontext -a -t svirt_tmpfs_t /dev/shm/looking-glass
--- a/system_files/desktop/shared/usr/share/ublue-os/just/60-custom.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/60-custom.just
@@ -327,6 +327,6 @@ disable-watchdog:
   fi
   rpm-ostree kargs $WATCHDOG_KARGS
 
-# Add SELinux file context for default looking-glass shm file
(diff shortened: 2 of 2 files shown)


KyleGospo (approved):

KyleGospo:
Regarding your last point, is the issue merely a lack of packages? I'm open to getting a copr of this going if that's the case to avoid the building step

HikariKnight:
> Regarding your last point, is the issue merely a lack of packages? I'm open to getting a copr of this going if that's the case to avoid the building step

a copr would be very nice, however if the client (which is on the host) ever updates and the host application (in the guest VM) ever gets out of sync with versioning, the user might be without any access to their VM without an any alternative way in (like vnc, not the built in spice vnc) to update the host application.

the best alternative i would think of is to somehow implement a warning to the user that looking-glass will be updated to the newest release whenever there is an update? so that they can get the VM ready first :thinking:

gnif:
Please see: https://looking-glass.io/docs/B7-rc1/install_libvirt/#ivshmem
This is a substandard configuration.

> the user might be without any access to their VM without an any alternative way in (like vnc, not the built in spice vnc) to update the host application.

LG has supported the SPICE protocol for VGA fallback since the release of B6.

HikariKnight:
> This is a substandard configuration. Please see: https://looking-glass.io/docs/B7-rc1/install_libvirt/#ivshmem
> 
> > the user might be without any access to their VM without an any alternative way in (like vnc, not the built in spice vnc) to update the host application.
> 
> LG has supported the SPICE protocol for VGA fallback since the release of B6.
> 
> Please also note that the Looking Glass Team will not provide support for people using Bazzite with LG, LG is not considered ready for packaging or including into any third party distros or products at this time. Users coming to the LG support channels that are using Bazzite will be referred back to your support channels for help.
> 
> Btw, I do think it's cool that you have added LG support at all here :)

no problem @gnif i try my best already in the bazzite discord to help those with VM and passthrough issues, however we do not compile or ship the binary, we try to follow the documentation as closely as possible and we based everything on B6 "stable" specifically to standardize so all we do is add the SELinux rules and the shm tmpfile so that the LG binary __will work__, the user still has to manually compile the binary themselves following upstream documentation so we do not actually ship LG.

Also  based on the info you provided in the forum that the kernel module is not LG version specific, which means we can actually consider including the kernel module for those that want to use it (since kernel modules has to be included with the image with ostree), that way all the user needs would be to compile the binary in order to use LG.

PR #441, opened 2023-10-16
fix: removed unneeded parts of the nix installation on the justfile

dnkmmr69420:
fix(deck): removed unneeded part of the justfile for the nix installation
    fix(desktop): removed unneeded part of the justfile for the nix installation


KyleGospo:
Thanks!

dnkmmr69420:
you're welcome

PR #452, opened 2023-10-18
Feat: Add Virtual Surround 7.1 setup to just

HikariKnight:
Added a basic setup using the HeSuVi convolver file for Control Room 1 from https://github.com/ShanonPearce/ASH-Listening-Set just so there is something in the config.

If the user wants something like DTS, Atmos or OpenAL, they will have to hunt down the convolver wav files themselves as I am unsure about the legalities behind those files, so I figured it is better to just play it safe.

Also fixes missing comments above the virtual channels commands.


EyeCantCU (approved):
This is incredible! Thank you

PR #459, opened 2023-10-22
fix: prevent vfio initramfs args from conflicting with bazzite-hardware-setup

HikariKnight:
Noticed this morning that bazzite-hardware-setup nuked the vfio arguments from initramfs.
After some testing I found out that since `rpm-ostree initramfs --enable` is run, i could just add the vfio arguments to `/etc/dracut.conf.d/vfio.conf` and rebuild initramfs once, thereby preventing any future conflict.

However it will also reduce bazzite hw_version by 1 to re-trigger bazzite-hardware-setup so that it will re-add any missing initramfs arguments.

I initially had a variable where i used perl to check what the current initramfs args are and then include them in the `rpm-ostree initramfs --enable` command as this would avoid having to rebuild initramfs 2 times, however after some consideration i decided this might not be reliable down the line and instead opted to have initramfs rebuild 2 times instead as that would be a lot safer (and it would be a 1 time thing anyway)

### DIFF
--- a/system_files/desktop/shared/usr/share/ublue-os/just/60-custom.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/60-custom.just
@@ -272,7 +272,6 @@ enable-vfio:
   INITRAMFS_ARGS=$(rpm-ostree status | grep -m 1 Initramfs | perl -pe "s/\s+(Initramfs: regenerate|Initramfs:)//" | perl -pe "s/ \'/ --arg='/")
   if [[ ${VIRT_TEST} == *kvm.report_ignored_msrs* ]]; then
       echo 'add_drivers+=" vfio vfio_iommu_type1 vfio-pci "' | sudo tee /etc/dracut.conf.d/vfio.conf
-      echo $(($(cat /etc/bazzite/hws_version)-1)) | sudo tee /etc/bazzite/hws_version
       rpm-ostree initramfs --enable $INITRAMFS_ARGS
     if [[ ${CPU_VENDOR} == "AuthenticAMD" ]]; then
       VENDOR_KARG="amd_iommu=on"
--- a/system_files/desktop/shared/usr/share/ublue-os/just/60-custom.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/60-custom.just
@@ -270,10 +270,9 @@ enable-vfio:
   CPU_VENDOR=$(grep "vendor_id" "/proc/cpuinfo" | uniq | awk -F": " '{ print $2 }')
   VENDOR_KARG="unset"
   if [[ ${VIRT_TEST} == *kvm.report_ignored_msrs* ]]; then
-    rpm-ostree initramfs \
-      --enable \
-      --arg="--add-drivers" \
-      --arg="vfio vfio_iommu_type1 vfio-pci"
+      echo 'add_drivers+=" vfio vfio_iommu_type1 vfio-pci "' | sudo tee /etc/dracut.conf.d/vfio.conf
+      echo $(($(cat /etc/bazzite/hws_version)-1)) | sudo tee /etc/bazzite/hws_version
+      rpm-ostree initramfs --enable $INITRAMFS_ARGS
     if [[ ${CPU_VENDOR} == "AuthenticAMD" ]]; then
       VENDOR_KARG="amd_iommu=on"
     elif [[ ${CPU_VENDOR} == "GenuineIntel" ]]; then
--- a/system_files/desktop/shared/usr/share/ublue-os/just/60-custom.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/60-custom.just
@@ -269,6 +269,7 @@ enable-vfio:
   VIRT_TEST=$(rpm-ostree kargs)
   CPU_VENDOR=$(grep "vendor_id" "/proc/cpuinfo" | uniq | awk -F": " '{ print $2 }')
   VENDOR_KARG="unset"
+  INITRAMFS_ARGS=$(rpm-ostree status | grep -m 1 Initramfs | perl -pe "s/\s+(Initramfs: regenerate|Initramfs:)//" | perl -pe "s/ \'/ --arg='/")
   if [[ ${VIRT_TEST} == *kvm.report_ignored_msrs* ]]; then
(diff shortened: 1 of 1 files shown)


KyleGospo:
You can remove the initramfs line and simple delete that file rather than decrement.

fiftydinar:
> You can remove the initramfs line and simple delete that file rather than decrement.
> 
> Apologies for breaking this use case inadvertently.

This.

This is one of the advantages of initramfs generation. When you delete some karg from hardware-setup you previously included, users won't get that karg removed. While here, you delete the modprobe/dracut-conf file + initramfs line & that's it. It is really seamless.

HikariKnight:
> Apologies for breaking this use case inadvertently.
This is ok, at least it was something easy to catch

> > You can remove the initramfs line and simple delete that file rather than decrement.
> > Apologies for breaking this use case inadvertently.
> 
> This.
> 
> This is one of the advantages of initramfs generation. When you delete some karg from hardware-setup you previously included, users won't get that karg removed. While here, you delete the modprobe/dracut-conf file + initramfs line & that's it. It is really seamless.

> You can remove the initramfs line and simple delete that file rather than decrement.

problem is that initramfs has to be rebuilt after adding the vfio driver, deleting the hw_version file will not trigger a new rebuild as that is only triggered in bazzite-hardware-setup if an initramfs argument is missing

i can make it work with a single rebuild using this hacky original commit
```bash
  INITRAMFS_ARGS=$(rpm-ostree status | grep -m 1 Initramfs | perl -pe "s/\s+(Initramfs: regenerate|Initramfs:)//" | perl -pe "s/ \'/ --arg='/")
  if [[ ${VIRT_TEST} == *kvm.report_ignored_msrs* ]]; then
    rpm-ostree initramfs \
      --enable \
      --arg="--add-drivers" \
      --arg="vfio vfio_iommu_type1 vfio-pci"
      echo 'add_drivers+=" vfio vfio_iommu_type1 vfio-pci "' | sudo tee /etc/dracut.conf.d/vfio.conf
      echo $(($(cat /etc/bazzite/hws_version)-1)) | sudo tee /etc/bazzite/hws_version
      rpm-ostree initramfs --enable $INITRAMFS_ARGS
```
but it is not really future proof (or safe?) as it will just guess the proper initramfs arguments based on the quoted initramfs args reported by `rpm-ostree status`, as `rpm-ostree initramfs` only reports the raw string.

The old method used `rpm-ostree initramfs --arg="--add-driver" --arg="vfio vfio_iommu_type1 vfio-pci"` however this would overwrite the initramfs args done by bazzite-hardware-setup, causing it to revert it on the next hw_version bump due to missing initramfs args.

I can just edit bazzite-hardware-setup to trigger an initramfs rebuild if hw_version is not present, that might be a better solution overall?

HikariKnight:
Something like this

HikariKnight:
Reverted the commits to the original solution. It will query `rpm-ostree status` which shows every argument separately inside single quotes and try fetch the initramfs arguments bazzite-hardware-setup used from there and apply them to the rebuild command, it will work until a more permanent solution for users to add things to initramfs is decided upon.

In most cases this will do 1 single rebuild of initramfs, however in some instances if the user has queued up a `regenerate` with `rpm-ostree initramfs --enable` it will trigger another initramfs rebuild the next time hw_version is updated.

EyeCantCU:
What if the `just` shortcut just created a file bazzite hardware setup looks for, then hardware setup gets retriggered and handles it? The file will exist in the future so every time initramfs is updated, it'll know to update it during hardware setup as well

HikariKnight:
> What if the `just` shortcut just created a file bazzite hardware setup looks for, then hardware setup gets retriggered and handles it? The file will exist in the future so every time initramfs is updated, it'll know to update it during hardware setup as well

this is a similar solution that i have been thinking about which would work for now, however if any future just commands want to add to the initramfs lines without using a file in `/etc/dracut.conf.d` it will get overwritten, but that is something that can be looked at when we need that.

will be easy to just write a check for `/etc/bazzite/rebuild_initramfs` and then delete it and rebuild initramfs, the main issue is that the changes would then not be ready without rebooting after rebuilding the initramfs, so we would probably have to trigger a reboot if it is a triggered rebuild.

i should have some free time this weekend to look at this again.

EyeCantCU:
> > What if the `just` shortcut just created a file bazzite hardware setup looks for, then hardware setup gets retriggered and handles it? The file will exist in the future so every time initramfs is updated, it'll know to update it during hardware setup as well
> 
> this is a similar solution that i have been thinking about which would work for now, however if any future just commands want to add to the initramfs lines without using a file in `/etc/dracut.conf.d` it will get overwritten, but that is something that can be looked at when we need that.
> 
> will be easy to just write a check for `/etc/bazzite/rebuild_initramfs` and then delete it and rebuild initramfs, the main issue is that the changes would then not be ready without rebooting after rebuilding the initramfs, so we would probably have to trigger a reboot if it is a triggered rebuild.
> 
> i should have some free time this weekend to look at this again.

We could write all needed initramfs changes to a file in /etc/bazzite and trigger a rebuild when the file hash changes or when necessary without other changes being overwritten

HikariKnight:
> > > What if the `just` shortcut just created a file bazzite hardware setup looks for, then hardware setup gets retriggered and handles it? The file will exist in the future so every time initramfs is updated, it'll know to update it during hardware setup as well
> > 
> > 
> > this is a similar solution that i have been thinking about which would work for now, however if any future just commands want to add to the initramfs lines without using a file in `/etc/dracut.conf.d` it will get overwritten, but that is something that can be looked at when we need that.
> > will be easy to just write a check for `/etc/bazzite/rebuild_initramfs` and then delete it and rebuild initramfs, the main issue is that the changes would then not be ready without rebooting after rebuilding the initramfs, so we would probably have to trigger a reboot if it is a triggered rebuild.
> > i should have some free time this weekend to look at this again.
> 
> We could write all needed initramfs changes to a file in /etc/bazzite and trigger a rebuild when the file hash changes or when necessary without other changes being overwritten

sometimes though in the case of vfio we dont need to add any initramfs commands directly, just need to trigger a rebuild to include new configured modules, so a way to force a rebuild will be needed either way.
I will start working on this in the morning.

HikariKnight:
Closing this as the fix will be baked in with the reworked initramfs management since it is what kicked off this anyway

PR #488, opened 2023-10-30
chore: Add cleaner fedora distrobox check.

ashylily:
Replaces the identical Fedora distrobox checks with a hidden recipe making things a bit cleaner and making it easier to add new recipes that use the distrobox.


KyleGospo:
This is excellent, thank you

PR #525, opened 2023-11-09
feat: Always layer Steam

KyleGospo:
Drops the `bazzite-arch` pattern in favor of all images having steam layered into them. A lot has changed since our 1.0 release and while bazzite-arch has it's place, I think this makes more sense for the majority of our users now.

Additionally opens us up to supporting HDR via gamescope and vk_hdr_layer on desktop images.

Note that we will continue to maintain bazzite-arch as it continues to be a useful tool for numerous other distros, and can still be optionally used on bazzite if desired - IE: Mesa git, AMDGPU-Pro.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -162,7 +162,7 @@ RUN rpm-ostree install \
         vk_hdr_layer.x86_64 \
         vk_hdr_layer.i686 \
         gperftools-libs.i686 && \
-    if [[ "${IMAGE_FLAVOR}" != "surface-nvidia" ]]; then \
+    if [[ ! "${IMAGE_FLAVOR}" =~ "surface" ]]; then \
         rpm-ostree install \
             obs-vkcapture.x86_64 \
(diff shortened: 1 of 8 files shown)


Eeyoresyndrome:
My issues with the steam flatpak were mostly: it didn't include udev rules for controllers. The additional time it hangs waiting for network access is sorta just a QOL thing for me.

Anyway I'm for this change. -bazzite AMD desktop user.

castrojo:
We include the udev rules already so that shouldn't be an issue?

EyeCantCU (approved):
🐐

bsherman:
I've looked through this and it seems pretty good to me. Happy to defer to @EyeCantCU (who has already approved) as I'm not as familiar with the details of bazzite builds.

EyeCantCU:
Everything here looks great to me. Can confirm it works as it should over in unstable as well

EyeCantCU (approved):

PR #534, opened 2023-11-12
feat: Alma linux containers on justfile

dnkmmr69420:


dnkmmr69420:
now [here](https://github.com/ublue-os/bazzite/pull/538)

Issue #533, opened 2023-11-12
Open for comment: DX Images (Developer Experience)


KyleGospo:
Similarly to Bluefin-DX, we'd like to ship DX images which come preconfigured with common game development and debugging tools. If you're a developer we'd love to hear what you might find useful in an image like Bazzite.

EyeCantCU:
Suggestions:
* godot: https://godotengine.org/
* android-tools: Provides ADB, needed for Android game development

Utilities that would likely be helpful:
* Blender
* GIMP
* InkScape
* Audacity

Things that would be cool, but totally not necessary/probably out of scope:
* Zsh - with oh-my-zsh configured with the half life theme
* chatgpt-shell-cli symlinked to GLaDOS/glados
* Nerd fonts - broader symbol support
* bat, eza, fd - robust alternatives to cat, ls, and find
* virt-manager - for image testing

badlydrawnface:
That would seem interesting! Getting JetBrains IDEs installed through a copr is a bit of a hassle and there doesn't seem to be a way to make JetBrains Toolbox run on a specific Distrobox container (obviously excluding the AUR package), because of how it installs, it can only install on the host, where there won't be any (other than probably Python) SDKs installed.

EyeCantCU:
> That would seem interesting! Getting JetBrains IDEs installed through a copr is a bit of a hassle and there doesn't seem to be a way to make JetBrains Toolbox run on a specific Distrobox container (obviously excluding the AUR package), because of how it installs, it can only install on the host, where there won't be any (other than probably Python) SDKs installed.

The challenge here is, you can't throw all of them into one distrobox image (nor do people need all of them... usually) as the size is too big for GHCR unless you do it after the fact

One solution would be using distrobox assemble for each JetBrains IDE, keeping the environment completely separate from one another, installing everything while setting the image up on the host

Another alternative is one distrobox instance that provides everything needed by each IDE with just scripts that install whatever IDE you need

castrojo:
Jetbrains has it's own "toolbox" that people have been using and so far the feedback has been positive: https://universal-blue.org/images/bluefin/developer-experience/#jetbrains

EyeCantCU:
> Jetbrains has it's own "toolbox" that people have been using and so far the feedback has been positive: https://universal-blue.org/images/bluefin/developer-experience/#jetbrains

[...]

PR #565, opened 2023-11-19
feat: Add DX Images

KyleGospo:
https://github.com/ublue-os/bazzite/issues/533

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -515,21 +515,11 @@ COPY --from=cgr.dev/chainguard/flux:latest /usr/bin/flux /usr/bin/flux
 COPY --from=cgr.dev/chainguard/helm:latest /usr/bin/helm /usr/bin/helm
 COPY --from=cgr.dev/chainguard/ko:latest /usr/bin/ko /usr/bin/ko
 COPY --from=cgr.dev/chainguard/minio-client:latest /usr/bin/mc /usr/bin/mc
-COPY --from=cgr.dev/chainguard/kubectl:latest /usr/bin/kubectl /usr/bin/kubectl
-
-RUN curl -Lo ./kind "https://github.com/kubernetes-sigs/kind/releases/latest/download/kind-$(uname)-amd64" && \
-    chmod +x ./kind && \
-    mv ./kind /usr/bin/kind
 
 # Install DevPod
 RUN rpm-ostree install https://github.com/loft-sh/devpod/releases/download/v0.3.7/DevPod_linux_x86_64.rpm && \
     wget https://github.com/loft-sh/devpod/releases/download/v0.3.7/devpod-linux-amd64 -O /tmp/devpod && \
     install -c -m 0755 /tmp/devpod /usr/bin
-
-# Install kns/kctx and add completions for Bash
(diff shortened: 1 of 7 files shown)


Containerfile
```diff
@@ -495,3 +495,43 @@ RUN /tmp/image-info.sh && \
...
+
+FROM bazzite as bazzite-dx
+
+ARG IMAGE_NAME="${IMAGE_NAME}"
+ARG IMAGE_VENDOR="${IMAGE_VENDOR}"
+ARG IMAGE_FLAVOR="${IMAGE_FLAVOR}"
+ARG BASE_IMAGE_NAME="${BASE_IMAGE_NAME}"
+ARG FEDORA_MAJOR_VERSION="${FEDORA_MAJOR_VERSION}"
+
+COPY system_files/deck/shared system_files/deck/${BASE_IMAGE_NAME} /
```
EyeCantCU, line 507: Instead of copying Deck files, we could move from this being an image target to an option so that DX desktop and DX Deck images can exist separately by adding another build flag that distinguishes itself between standard images and DX images

Containerfile
```diff
@@ -495,3 +495,43 @@ RUN /tmp/image-info.sh && \
...
+    wget https://github.com/docker/compose/releases/latest/download/docker-compose-linux-x86_64 -O /tmp/docker-compose && \
+    install -c -m 0755 /tmp/docker-compose /usr/bin
+
+COPY --from=cgr.dev/chainguard/flux:latest /usr/bin/flux /usr/bin/flux
+COPY --from=cgr.dev/chainguard/helm:latest /usr/bin/helm /usr/bin/helm
+COPY --from=cgr.dev/chainguard/ko:latest /usr/bin/ko /usr/bin/ko
+COPY --from=cgr.dev/chainguard/minio-client:latest /usr/bin/mc /usr/bin/mc
+
+# Install DevPod
+RUN rpm-ostree install https://github.com/loft-sh/devpod/releases/download/v0.3.7/DevPod_linux_x86_64.rpm && \
```
EyeCantCU, line 520: I believe I read that this is going to be provided as an AppImage only going into the future with only the CLI being provided as an RPM. We could install the CLI, and then have a dx entry in the portal that provides downloading the AppImage as an option if that sounds all right to you?

castrojo: The plan was to COPR this one with the intent of seeing if someone can help us get it in fedora.

EyeCantCU: I don't mind looking into throwing it on COPR. I think that's the best bet as it would be nice for it to just be there out of the box

markob94:
Any progress on DX version ?

KyleGospo:
> Any progress on DX version ?

Still planned, we'll be returning to this soon

FlyinPancake:
Are you guys waiting on some external blockers, or do you need some manpower on this? I can provide the later one :D

FlyinPancake:
For the time being I built a "DX" image with bluebuild no nvidia builds yet since I use it on Intel and AMD GPUs however I can add the if someone needs it <https://github.com/FlyinPancake/cirrus-os>

PR #567, opened 2023-11-20
feat: add just syntax checker

ArtikusHG:
Integrate our new [just syntax checker](https://github.com/ublue-os/just-action) into the build action.


EyeCantCU (approved):
Thank you for the contribution!

PR #572, opened 2023-11-21
feat: Preinstall GOverlay

nicknamenamenick:
GOverlay (https://github.com/benjamimgois/goverlay)

It's a nice GUI for MangoHud, vkBasalt, and ReplaySorcery though.  In my testing in layering it to Bazzite, it complains MangoHud is missing, but it worked in my testing so I'm not sure why it does that. 
**EDIT:** Known issue in the latest stable.

It could also be placed in the Bazzite Portal as well instead of being preinstalled, and I can make a new PR for that.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -163,6 +163,7 @@ RUN rpm-ostree install \
         vk_hdr_layer.x86_64 \
         vk_hdr_layer.i686 \
         gperftools-libs.i686 && \
+        goverlay \
     if [[ ! "${IMAGE_FLAVOR}" =~ "surface" ]]; then \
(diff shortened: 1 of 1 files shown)


EyeCantCU:
It would be ideal to figure out why it claims MangoHUD isn't installed before including this, although I do think it is a great addition

nicknamenamenick:
> It would be ideal to figure out why it claims MangoHUD isn't installed before including this, although I do think it is a great addition

Seems to be an issue with the [current stable release](https://github.com/benjamimgois/goverlay/issues/73)

PR #595, opened 2023-12-02
chore: Remove documentation link in Bazzite Portal

nicknamenamenick:
The Forums already contain the documentation.  The user no longer has to scroll down to click the button that reboots their device.


EyeCantCU (approved):
Thank you!

PR #623, opened 2023-12-19
fix: add missing shebang to selinux-looking-glass

HikariKnight:
My previous pull request got merged a few seconds before a commit adding the missing shebang to the recipe got pushed, my bad.
This corrects it, still getting used to how just operates.


EyeCantCU:
Thank you!

PR #641, opened 2024-01-02
fix(ci): build Release ISO condition

duhow:
Boolean output becomes a `string`, so needs to be checked against string value.
This should fix the :x: invalid merges to `main`.


duhow:
Also relates to https://github.com/google-github-actions/release-please-action/issues/912 .

EyeCantCU (approved):
Thank you so much for this! I really appreciate the fix

PR #669, opened 2024-01-10
feat: Bring fsync kernel to stable thanks to newly finished Secure Boot support, along with months of changes that have been held back by it!

KyleGospo:
Now that @EyeCantCU 's kernel signing is working, this PR will ship all the enhancements we've made in the testing branch.

Note that this will require everyone using Secure Boot to have added the akmods mokutil key -- most should have already due to degraded functionality without it.


EyeCantCU (approved):
🎉

bsherman:
So, this means, if you use SecureBoot and _don't_ have the ublue MOK, the kernel won't boot at all?

EyeCantCU:
> So, this means, if you use SecureBoot and _don't_ have the ublue MOK, the kernel won't boot at all?
> 
> Or, will the ublue signature be added to the pre-existing signature, so the user would have an experience similar to current without the ublue MOK, only our kmods would fail to load?

We use the same key pair here as what's used in akmods. If people had the key enrolled before, it will continue to function as it was. If not, they already would have had to have had secure boot disabled. It shouldn't impact anyone

KyleGospo:
Unfortunate requirement of using the fsync kernel since it's not signed by fedora. That being said we'll update our documentation and further telegraph this change.

bsherman:
Certainly good you can sign the kernel and make this possible at all! Probably just good to have well documented. I tend to think a potential total failure to boot is a bit harsher than "degraded experience". :-)

KyleGospo:
> Certainly good you can sign the kernel and make this possible at all! Probably just good to have well documented. I tend to think a potential total failure to boot is a bit harsher than "degraded experience". :-)

Yeah, at the very least we know anyone using features like displaylink or utilizing a Nvidia graphics card has already enrolled that key.

I think what I will do is put an additional week buffer on this (this change was mentioned in the last release at the top), and add a message to the terminal during that time warning people about it.

If this bites anyone after that at least it's as easy as turning off secure boot for them.

KyleGospo:
I should also mention that they will still have their previous deployment as well, so they're not completely dead. Ideally they would boot into the previous deployment that then has that warning in the terminal.

bsherman:
> I should also mention that they will still have their previous deployment as well, so they're not completely dead. Ideally they would boot into the previous deployment that then has that warning in the terminal.

Awesome point and also great to document!  

atomic systems saving the day again!

m2Giles:
New installs need to have the key. Maybe kickstart should also have the enroll key like nvidia.

Already without the key display link was broken and you would have an error showing.

Maybe for the next week-month have a service that checks if key was enrolled with `mokutil --list-enrolled`. If not stage enrolled key. UX for that might be hard on a deck though.

KyleGospo:
> New installs need to have the key. Maybe kickstart should also have the enroll key like nvidia.
> 
> Already without the key display link was broken and you would have an error showing.
> 
> Maybe for the next week-month have a service that checks if key was enrolled with `mokutil --list-enrolled`. If not stage enrolled key. UX for that might be hard on a deck though.
> 
> ```
> if ! $(mokutil --list-enrolled | grep -qv "Universal Blue"); then
>         echo "Key not Enrolled"
>         mokutil --import /etc/pki/akmods/certs/akmods-ublue.der
>         exit 0
> fi
> echo "Key already Enrolled"
> ```
> 
> Though you probably need to do something to pass something to mokutil for the password on import.

Good news is the deck doesn't ship secure boot enabled at all, so no need to worry about UX there.

system_files/deck/shared/usr/share/ublue-os/just/60-custom.just
```diff
@@ -8,6 +8,12 @@ toggle-autologin:
       sudo touch $DESKTOP_AUTOLOGIN
     fi
 
+# Install Bazzite's Steam Game Mode Startup Video
+install-gamemode-video:
+    mkdir -p $HOME/.steam/root/config/uioverrides/movies/deck_startup.webm
```
SuperRiderTH: Shouldn't this just be 

`mkdir -p $HOME/.steam/root/config/uioverrides/movies/` ?

HikariKnight (approved):

PR #699, opened 2024-01-20
feat: Clean up just files

KyleGospo:
@HikariKnight 

Good start here, this does the split we talked about which should make the actual cleanup treatment like ublue-os/config got easier.


HikariKnight:
yes it should.

PR #710, opened 2024-01-25
feat(ci): Introduce a stable tag

EyeCantCU:
The "stable" and "latest" tags are equivalent. Given "testing" and "unstable" tags are used, users may likewise figure that the current stable release uses the "stable" tag


KyleGospo:
Bridge this soon, just going to update the readme to point to these tags

KyleGospo (approved):

PR #720, opened 2024-01-29
fix(steamos-update): Properly read the return code of the upgrade command

mbouron:


EyeCantCU:
Thank you so much for catching this!

EyeCantCU (approved):

PR #729, opened 2024-02-01
fix: added glibc-headers and glibc-devel to override list

noelmiller:
This is to fix build issues I found this morning where glibc-headers and glibc-devel have dependency conflicts. I found this issue in my custom image which shares most of Bazzite’s Containerfile (minus a few things: https://github.com/noelmiller/isengard/actions/runs/7739690630/job/21103077663#step:6:1425)

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -174,6 +174,12 @@ RUN rpm-ostree override replace \
         || true && \
     rpm-ostree override replace \
     --experimental \
+    --from repo=updates-archive \
+        glibc-headers \
+        glibc-devel \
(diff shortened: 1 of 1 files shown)


Containerfile
```diff
@@ -174,6 +174,12 @@ RUN rpm-ostree override replace \
         || true && \
     rpm-ostree override replace \
     --experimental \
+    --from repo=updates-archive \
```
EyeCantCU, line 177: Unsure if we should be pinning to updates archive here

PR #755, opened 2024-02-09
feat(gamescope): add env var for 720p patch

Sterophonick:
This patch allows for an environment variable, `GAMESCOPE_ENABLE_720P_RESTRICT`, to be set to re-renable the 720p restriction that Steam sets, ideal for setups such as mine on Steam Deck that have scripts to auto-adjust game config files based on what's visible to `XWAYLAND1`, also for games like DOOM Eternal where setting `+r_mode -2` forces the resolution to the highest available. Disabled by default, can be set in `/etc/environment`.


KyleGospo:
Thanks for this, excellent change!

PR #765, opened 2024-02-12
chore(just): cleanup and standardize just recipes

HikariKnight:
# Justfile cleanup and refactor
## Splitting justfiles into multiple files
By default all files are shared across `desktop` and `deck` images
- `80-bazzite.just` contains general bazzite system stuff (terminal, grub, gamemode video, etc)
- `81-bazzite-fixes.just` contains fixes/patches/workarounds to common things (ex: tcmalloc)
- `82-bazzite-apps.just` generic `install-` or `setup-` commands for apps that only provide install/uninstall functions
- `82-bazzite-$APPNAME.just` dedicated just file for more complex setup/configure scripts (does more than install/uninstall), currently `sunshine` and `waydroid`.
- `83-bazzite-audio.just` justfile dedicated to audio stuff (ex: virtual surround and virtual audio channels)
- `84-bazzite-virt.just` recipes for virtualization setup and configuration
- `85-bazzite-image.just` image specific recipes (deck vs desktop)
- `90-bazzite-de.just` DE specific recipes (kde vs gnome vs other)

## Numbering
This does not matter much but i figured there could be some logic to the madness anyway and build on what @KyleGospo had already done.
numbers 80 to 99 are reserved for image use (eg: bazzite, bluefin, etc) and follows the logic above, but each number is assigned to a "scope"
- 80 for the system
- 81 for fixes
- 82 for apps
- 83 for audio/visual (what kind of visual is TBD)
- 84 for virtualization and containerization (not application related)
- 85 for hardcoded stuff (ex: image specific)
- 86-89 is unused
- 90 for hardcoded DE stuff.
- 91-99 is unused

# Refactoring
A lot of the just code will be refactored to include dialogs, however i am not sure if all the `install-foo` applications need an uninstall action :thinking: 
Primarily we just need to split things up and get things more organized.


KyleGospo:
No idea how this got marked as ready, fixed.

HikariKnight:
Unsure if there is anything else here that will be needed or changed unless anyone has any suggestions while reviewing

HikariKnight:
Added the missing addition of `configure-powerprofile` to let people set power-save as the default tuneD profile when the desktop starts.
We can add `performance` profile to the list, but after discussion the value it will provide is negligible. The user should instead switch to it when they need it.

We only recommend balanced, as we do not have any ability to set the profile back to balanced when the user switches back to gamemode.

as such there is a highlight to ask people to only do performance and audio bug reports using the balanced profile
image

KyleGospo (approved):

KyleGospo:
This looks amazing dude, can't thank you enough

PR #773, opened 2024-02-15
feat(waydroid): add automount of shared waydroid directory to waydroid-launcher

aarron-lee:
Updates waydroid-launcher to create a `WaydroidShare` folder in the home directory and mounts it in waydroid.

This lets you share files directly  with waydroid without copying over the entire file.

Note that waydroid itself will see the shared folder as read-only, it will not be able to write to the directory.

waydroid documentation for setting up a shared folder:

https://docs.waydro.id/faq/setting-up-a-shared-folder


aarron-lee:
will reopen later, noticing an issue when restarting waydroid

PR #782, opened 2024-02-17
fix(deck): Correct rotation for Lenovo p14s gen2 in the deck image

HikariKnight:
For some reason the desktop gets rotated to the right on this laptop when on the deck image, this laptop has a landscape display and does not match any of our checks. Meaning it gets rotated by valve by accident?
This will correct it back to normal rotation.

reported by Crush in discord.


HikariKnight:
Crush tried running `bazzite-rotation-fix` directly but it did not trigger `kscreen-doctor output.1.rotation.normal`
running the command directly did fix the screen rotation on the laptop though, marking as draft until i can investigate further

HikariKnight:
noticed Crush had run a downloaded version of `bazzite-rotation-fix` and that didnt work but running `/usr/libexec/bazzite-rotation-fix` on the system worked.

so this is now confirmed to fix it now.

PR #792, opened 2024-02-22
Add rclone to the bazzite image for mounting cloud storage

dylanmtaylor:
This will allow users to mount storage, for instance:

rcloud mount onedrive:/ /mnt/OneDrive

This does not work in distrobox or docker as you can't do a FUSE mount from within a container.


KyleGospo:
Out of scope for this project, but that should be easy to layer for those that need it.

PR #868, opened 2024-03-09
fix: custom initramfs args actually get applied

m2Giles:
rpm-ostree initramfs is very temperamental.


HikariKnight (approved):
It was very temperamental when you showed me this, yes
It worked when i told it to apply my vfio args back when that was not a dracut.conf file

EyeCantCU (approved):
LGTM

PR #884, opened 2024-03-15
chore: add "console" keyword for Ptyxis

nicknamenamenick:


noelmiller (approved):

noelmiller:
We are going to implement this change in a different way, but thank you for contributing the idea :)

PR #904, opened 2024-03-22
feat(deck): Add ujust command to disable swipe...

nicknamenamenick:
...gestures (Thanks [Aru](https://github.com/aarron-lee)!)

Taken from: https://github.com/aarron-lee/legion-go-tricks/blob/main/bazzite-disable-swipe-gestures.sh

edit: i will get to this eventually unless someone beats me to the punch.  may wait until after 40 releases.


noelmiller (changes requested):
This might be better as a toggle rather than a pure disabling of the feature. If someone wants to turn it back on, they would have to edit a config file in order to do that. @HikariKnight maybe this something you can help with?

nicknamenamenick:
> This might be better as a toggle rather than just disabling the feature. If someone wants to turn it back on, they would have to edit a config file in order to do that. @HikariKnight maybe this something you can help with?

I agree, it is.  Although, the `-deck` images never got the same standardization treatment.  Hikari worked on the desktop images and upstream, but planned touching `ujust` commands on `-deck` images in the future.  

Mostly due to documentation from other proejct (HHD, certain Decky plugins that we use, etc.) and Youtube guides that show the older commands.  However, that doesn't mean this specific command couldn't benefit from a toggle since it's new.

HikariKnight:
> This might be better as a toggle rather than just disabling the feature. If someone wants to turn it back on, they would have to edit a config file in order to do that. @HikariKnight maybe this something you can help with?

make the toggle check if the file exists. if it exists, remove it, if it doesn't exist, add it. :)
and yes i agree it should be a toggle.

PR #905, opened 2024-03-22
chore: Remove hl2linux-selinux package

nicknamenamenick:
This package does not resolve the issue as intended, and Team Fortress 2 will eventually not have this issue anyways once the 64-bit executable lands in a future update. 

Left 4 Dead 2, and perhaps other Source 1 games, will still have this issue, but the fix can be found [here](https://universal-blue.discourse.group/docs?topic=574).


noelmiller (approved):
Confirm with @KyleGospo, but LGTM.

PR #908, opened 2024-03-23
feat: remove custom rpm-ostree initramfs arg support

HikariKnight:
After longer testing with @m2Giles, `rpm-ostree initramfs` is too finiky to properly support with custom arg support, dracut.conf files work better without stepping on bazzites initramfs args, and is less messy.

@KyleGospo maybe it would also be an idea to move bazzites initramfs args to something like `/usr/lib/dracut/dracut.conf.d/bazzite.conf` in the future, or is it using rpm-ostree initramfs directly so that it can be "hardwired" based on hardware and adjusted on the fly for the hardware?


KyleGospo:
Change looks good to me and I am open to making that move. I'll merge this into the next testing branch build to play with.

HikariKnight:
yeah i just opened the PR as a draft so we could start the discussion as me and M2 (but mostly them) found several issues where custom arg support is flaky as `rpm-ostree initramfs` only allows you to use `--arg=-I blah` once, if you provide another `-I` arg it will fail while using a `dracut.conf` file in `/etc/dracut.conf.d` or `/usr/lib/dracut/dracut.conf.d/` with the includes, it does not care.

We concluded in the end that its better to drop the custom args and instead try standardize with users using dracut files instead, that way only the system is allowed to use `rpm-ostree initramfs` or even consider moving our initramfs stuff to dracut and just tell people to write dracut configs instead of using `rpm-ostree initramfs` since we have the `rebuild` file we check for, which could be moved to something like `/etc/initramfs` or something now since i want to try make a similar system for bluefin, just need to remove the bazzite specific stuff.

I also believe @m2Giles said that if we went full dracut.conf files, we could potentially utilize `initramfs-etc` instead which would be faster i believe? correct me if i remember wrong.
so it is worth playing around with in testing (especially now that i got my nvidia test VM up and running)

fiftydinar:
> I also believe @m2Giles said that if we went full dracut.conf files, we could potentially utilize `initramfs-etc` instead which would be faster i believe? correct me if i remember wrong. so it is worth playing around with in testing (especially now that i got my nvidia test VM up and running)

Bazzite pytorch boot-screen does not reside in `/etc`, hence why initramfs-etc is not suitable. Initramfs is needed to update the custom boot screen to actually work.

m2Giles:
Initramfs-etc only seems to work for file includes from /etc so likely not suitable if drivers or modules need to be included.

So far I'm having success with no args on rpm-ostree initramfs and adding everything via config files in /etc/dracut.conf.d

fiftydinar:
> Initramfs-etc only seems to work for file includes from /etc so likely not suitable if drivers or modules need to be included.
> 
> So far I'm having success with no args on rpm-ostree initramfs and adding everything via config files in /etc/dracut.conf.d

I think it's better to use `/usr/lib/dracut/dracut.conf.d` for Bazzite config & `/etc/dracut/dracut.conf.d` for user config. User can reset the config easily by copying the config template from `/usr/etc/dracut/dracut.conf.d`

HikariKnight:
would be `/etc/dracut.conf.d/` for the user for reference as fiftydinar says obviously

KyleGospo:
I think what we need to do from here, ideally, is to use initramfs-etc, and look into a solution similar to what we do for kernel signing, where a post-build step opens the initramfs image, adds to it, and writes it back to the image.

fiftydinar:
@KyleGospo 

Sounds like a great idea!

I would maybe try to make it available universally too, like kernel-signer is.
I would definitely utilize it instead of my BlueBuild initramfs-etc module!

HikariKnight:
@KyleGospo 
I think it sounds good, @fiftydinar is the one who has played around with initramfs-etc so they know more about it than me.

So essentially what we would need is just add our current stuff to initramfs-etc and also make it easy for users to add more stuff to it (like vfio, apple keyboard fn mode, modprobe files, etc) and make a simple way for them to signal the system that there is new things to add, unless it is all automatic. :thinking:

m2Giles:
Can initramfs-etc handle adding dracut modules and drivers? I thought it only does files in /etc and is mutually exclusive with host generated initramfs

fiftydinar:
> Can initramfs-etc handle adding dracut modules and drivers? I thought it only does files in /etc and is mutually exclusive with host generated initramfs

That's something that needs to be tested.
In my module, I assume that including dracut modules from `/etc` will work.

fiftydinar:
I can discuss about this more in a separate issue, since this issue is targeted  to remove custom initramfs arg support, as @HikariKnight said.

Here's the thread about this on BlueBuild side:
https://github.com/blue-build/modules/issues/101

m2Giles:
For tracking /etc/dracut.conf.d/*, I think we can use systemd-path units to watch the directory and if any changes are made, create the rebuild flag in /etc/bazzite/initramfs.

For the default bazzite args, we need to get them out of being called by ostree initramfs since users are likely going to overwrite them by accident. We could add them to /usr/lib/dracut.conf.d/. Or if there are specific versions, use a symlink to /etc/dracut.conf.d/.

PR #924, opened 2024-03-29
feat: update image label with actual kernel version

bsherman:
Sets a variable with for the specific kernel being installed and uses that both to select the desired AKMODS layer and set the ostree.linux image label as appropriate.


KyleGospo:
Awesome surprise, thank you!

KyleGospo (approved):

PR #936, opened 2024-04-02
feat: move to upstream iso generator and add flatpak support

HikariKnight:
opening on behalf of @noelmiller due to github being weird


bsherman (approved):
worth a try

PR #945, opened 2024-04-05
Fedora 40

KyleGospo:


.github/workflows/build.yml
```diff
@@ -50,9 +51,9 @@ jobs:
         base_image_flavor: [main, asus, framework, surface]
         base_name: [bazzite, bazzite-deck, bazzite-nvidia]
         base_image_name: [kinoite, silverblue]
-        major_version: [39]
+        major_version: [40]
```
dylanmtaylor, line 54: Why not ship 39 and 40 concurrently?

.github/workflows/build_iso.yml
```diff
@@ -2,73 +2,157 @@ name: Build ISOs
...
+          echo "tag=${TAG}" >> $GITHUB_OUTPUT
+
       - name: Set EXTRA_BOOT_PARAMS
+        id: generate-extra-params
+        shell: bash
         run: |
           EXTRA_BOOT_PARAMS=""
-          if [[ "${{ matrix.image_name }}" == "bazzite-deck" || "${{ matrix.image_name }}" == "bazzite-deck-gnome" || "${{ matrix.image_name }}" == "bazzite-deck-budgie" ]]; then
+          if [[ "${{ matrix.image_name }}" =~ "deck" ]]; then
               EXTRA_BOOT_PARAMS="inst.resolution=1280x800"
```
dylanmtaylor, line 66: I wonder how we'd handle this if for instance a steamdeck 2 came out and had a higher resolution... is there a way to force it to match the native resolution of the display without hard-coding it?

system_files/desktop/shared/usr/libexec/bazzite-flatpak-manager
```diff
@@ -22,24 +22,6 @@ if [[ -f $VER_FILE && $VER = $VER_RAN ]]; then
...
 
-# Opt out of and remove Fedora's flatpak repo
-if grep -qz 'fedora' <<< $(flatpak remotes); then
-  /usr/lib/fedora-third-party/fedora-third-party-opt-out
-  /usr/bin/fedora-third-party disable
-  flatpak remote-delete fedora --force
-fi
-
-# Ensure Flathub is enabled
-flatpak remote-add --if-not-exists --system flathub /usr/etc/flatpak/remotes.d/flathub.flatpakrepo
```
dylanmtaylor, line 33: I think enabling Flathub by default was a nice touch. Any reason not to do this anymore?

FlyinPancake: AFAIK fedora 39+ ships with unfiltered flathub by default

fiftydinar: @dylanmtaylor 
Because new ISO action will enable FlatHub by default.

ISO is the main supported install method, while rebase is unsupported, so there is no reason for this to exist.

@FlyinPancake 

It ships unfiltered flathub, but does not enable it by default, unless you agree to the prompt in Gnome Software.

.github/workflows/build_iso.yml
FlyinPancake: While this is a valid question, this is not in the scope of this PR

Containerfile
```diff
@@ -44,19 +44,19 @@ RUN wget https://raw.githubusercontent.com/ublue-os/COPR-command/main/copr -O /u
...
+#    --experimental \
+#    --from repo=copr:copr.fedorainfracloud.org:sentry:kernel-fsync \
+#        kernel \
+#        kernel-core \
+#        kernel-modules \
+#        kernel-modules-core \
+#        kernel-modules-extra \
+#        kernel-uki-virt \
+#        kernel-headers \
+#        kernel-devel
```
FlyinPancake, line 59: `sentry:kernel-fsync` builds Fedora 40 images. Is there any reason to not include them?

Containerfile
```diff
@@ -107,24 +107,25 @@ RUN if [[ "${IMAGE_FLAVOR}" =~ "surface" ]]; then \
...
+#        /tmp/akmods-rpms/kmods/*evdi*.rpm \
+#        /tmp/akmods-rpms/kmods/*zenergy*.rpm \
+#        /tmp/akmods-rpms/kmods/*ayaneo-platform*.rpm \
+#        /tmp/akmods-rpms/kmods/*ayn-platform*.rpm \
+#        /tmp/akmods-rpms/kmods/*bmi160*.rpm \
+#        /tmp/akmods-rpms/kmods/*bmi260*.rpm \
+#        /tmp/akmods-rpms/kmods/*bmi323*.rpm \
+#        /tmp/akmods-rpms/kmods/*rtl8814au*.rpm \
+#        /tmp/akmods-rpms/kmods/*ryzen-smu*.rpm && \
+#    sed -i 's@enabled=1@enabled=0@g' /etc/yum.repos.d/negativo17-fedora-multimedia.repo
```
FlyinPancake, line 128: `ublue` builds `akmods` with F40 so I believe it should be added back

.github/workflows/build.yml
KyleGospo: > Why not ship 39 and 40 concurrently?

With the amount of package we have to manage it's a huge effort to maintain both, and gaming with older packages/Mesa is a bit silly.

PR #947, opened 2024-04-05
chore: Add patch for Ayn Loki refresh rates in gamescope

citral23:
This needs gamescope-session-plus edit in device-quirks to works, removing the panel being forced as external

If not it does nothing, so no harm done.


KyleGospo:
Thank you!

PR #953, opened 2024-04-07
feat: Enable USB Controller Polling Rate Patch

badlydrawnface:
Enabled some files for a patch in the kernel from Nobara that can increase a PlayStation controller's polling rate to 1000hz. Should a justfile entry be made? It would probably only enable/disable and restart, so probably not...

This can be very useful for Rocket League players.


fiftydinar:
Fsync kernel already has USB polling patch built-in, which Bazzite uses.

Is this needed to additionally enable this functionality?

badlydrawnface:
Yes, these are a few files that need to be included on the root of the OS. The patch itself is already in the kernel, but these files just enable it to work. Sorry for the misleading title.

KyleGospo:
This patch works via karg, so this is more of a helper. Im@ open to testing it.

PR #984, opened 2024-04-16
fix(ayaneo slide): flipping upside down in desktop mode

hijae:
ayaneo slide rotation fix. Same as Mini Pro, Loki max and AIR
IMG_0852 중간


KyleGospo:
Thank you!

PR #990, opened 2024-04-18
fix(ds-inhibit): autoload hid-playstation to avoid hook conflicts

Sterophonick:
ds-inhibit on upstream has issues with the first connect of PlayStation controllers, where the touchpads are not hooked correctly and therefore don't correctly interact with steam input. work around by using `modules-load.d` to autoload `hid-playstation` on boot.


Sterophonick:
ignore this please omg i hate

PR #1020, opened 2024-04-26
fix: bazzite-user-setup should only run when needed

bsherman:
I noticed that `bazzite-user-setup` is fully running on every reboot. 

The logic in the script makes it pretty clear this is not intended unless some versions change, etc. This fixes the bug.


bsherman:
For reference, I found this bug because I couldn't change the monospace font with GNOME Tweaks on bazzite. So I thought there must be some service force resetting the font. Thus I found this script, but this script doesn't set font.

So then I though maybe it was bazzites use of dconf defaults rather than gschema overrides which was causing a problem. But in fact, I think it's actually a bug in GNOME Tweaks, it doesn't seem to be able to change the monospace font anymore... I've verified the bug on both bazzite and my other silverblue F40 image... 

Seems to be a known bug after chatting in discord about it. 

https://gitlab.gnome.org/GNOME/gnome-tweaks/-/issues/489



This comment is purely commentary and not directly related to the fix itself. :-)

system_files/desktop/shared/usr/libexec/bazzite-user-setup
```diff
@@ -33,7 +33,7 @@ USER_SETUP_FEDORA_VER_RAN=$(cat $USER_SETUP_FEDORA_VER_FILE)
 USER_SETUP_IMAGE_VER_RAN=$(cat $USER_SETUP_IMAGE_VER_FILE)
 
 # Run script if updated
-if [[ -f $USER_SETUP_VER_FILE && -f $USER_SETUP_FEDORA_VER_RAN && -f $USER_SETUP_FEDORA_VER_FILE ]]; then
+if [[ -f $USER_SETUP_VER_FILE && -f $USER_SETUP_FEDORA_VER_FILE && -f $USER_SETUP_FEDORA_VER_FILE ]]; then
```
mulderje, line 36: Looks like you doubled up on the `USER_SETUP_FEDORA_VER_FILE` check. Maybe should be the below against the three files or just drop one of the checks if not needed? 

Also perhaps move the below three `cat` lines into the first if after the file exist checks?

```bash
USER_SETUP_VER_RAN=$(cat $USER_SETUP_VER_FILE)
USER_SETUP_FEDORA_VER_RAN=$(cat $USER_SETUP_FEDORA_VER_FILE)
USER_SETUP_IMAGE_VER_RAN=$(cat $USER_SETUP_IMAGE_VER_FILE)
```

```suggestion
if [[ -f $USER_SETUP_VER_FILE && -f $USER_SETUP_FEDORA_VER_FILE && -f $USER_SETUP_IMAGE_VER_FILE ]]; then
```

bsherman: Good catch! Thank you! Since this already got merged, you want to PR just the one line change?

I'll approve.

mulderje: No problem! #1023

PR #1023, opened 2024-04-26
fix(bazzite-user-setup): follow up from #1020 to include all file checks

mulderje:
Also pushes the file `cat` runs until after the first if


mulderje:
Follow up from #1020 

cc: @bsherman

KyleGospo:
Thank you!

PR #1038, opened 2024-04-28
fix: /etc/fstab adjustment condition

BWBellairs:
Don't look for compress=zstd as user may want to set this. Instead, set a flag so filesystem options are set on first boot instead.


KyleGospo (approved):

KyleGospo:
Great improvement, thanks!

PR #1069, opened 2024-05-03
fix(kde): Remove `QT_IM_MODULE` and `GTK_IM_MODULE` for IBus in wayland

chenx-dust:
To enable IBus in wayland session, it is necessary to remove environment variable  `QT_IM_MODULE` and `GTK_IM_MODULE`. The proper method to make ibus work is setting virtual keyboard to "IBus Wayland".


KyleGospo:
Thank you!

PR #1074, opened 2024-05-05
Update README.md

WesfunOfficial:
The idea of this commit is to emphasize the differences between the bazzite and bazzite-nvidia rebase requests. Emphasizing that the former will give you access to NVK but that it is prone to bugs and not to submit issues with Ublue/Bazzite but rather Mesa who is the maintainers of the driver at the time of writing. Hyperlink was added to clearly direct users of where to submit said bugs if need be

The latter emphasizing that the -nvidia requests come with NVIDIAs Proprietary drivers, not NVK with the rebase examples


WesfunOfficial:
@KyleGospo Could i have you look at this and see if everything is good?

KyleGospo:
Looks good, but let me merge next Mesa update. NVK still isn't quite ready. Thanks for the PR!

PR #1078, opened 2024-05-06
feat(nvidia): Default to x11 again

nicknamenamenick:
Maybe soon the proprietary Nvidia drivers can run a Wayland session without exploding into giblets.


KyleGospo:
Ambitious of you to think that of them

PR #1082, opened 2024-05-06
feat: mesa 24.1

aaron-trout:
Description of changes:

- Upgrade to mesa 24.1
- Remove Broadcom fix backport since it's part of 24.1 release
- 'nouveau-experimental' vulkan driver is now just 'nouveau'
- disable Intel RayTracing on anything but x86_64 (it's not supported anywhere else)
- Few dependency definitions / updates to fix build issues
  - python3-pycparser
  - cbindgen
  - version bump for wayland-protocols
  - ~restrict proc-macro2 crate to `< 1.0.80` (fixes `error[E0658]: use of unstable library feature 'proc_macro_byte_character'`)~
    - ~also restricting syn crate to `< 2.0.59` because of dependency conflict on proc-macro2~
  - add 'paste' crate
- add new dri libs to defined output files

This is a draft for now using the mesa 24.1 release candidate builds. Once the 24.1.0 final drops, I will update the PR and mark as ready for review. 

COPR Build: https://copr.fedorainfracloud.org/coprs/trouter/bazzite-multilib/package/mesa/


aaron-trout:
So the COPR [build works for x86_64](https://download.copr.fedorainfracloud.org/results/trouter/bazzite-multilib/fedora-40-x86_64/07420350-mesa/builder-live.log.gz) but [is failing on i686](https://download.copr.fedorainfracloud.org/results/trouter/bazzite-multilib/fedora-40-i386/07420350-mesa/builder-live.log.gz), at the point of installing rust dependencies. 

```
Updating and loading repositories:
 Copr repository                        100% | 101.3 KiB/s |   1.5 KiB |  00m00s
 local                                  100% | 126.3 KiB/s |   3.8 KiB |  00m00s
Repositories loaded.
Failed to resolve the transaction:
Problem 1: nothing provides requested (crate(proc-macro2) >= 1.0.56 with crate(proc-macro2) < 1.0.80)
 Problem 2: nothing provides requested (crate(syn/clone-impls) >= 2.0.15 with crate(syn/clone-impls) < 2.0.59)

Copr build error: Build failed
```

I'm not too sure why that would work on one arch but not the other; I thought they were not arch specific (and indeed in the working 64 bit build they show as `noarch` when being installed)

Any pointers would be helpful; I know close to zero about rust :smiley:

KyleGospo:
Been in the middle of a move but I will check this all out soon, thank you so much for doing it!

aaron-trout:
> Been in the middle of a move but I will check this all out soon, thank you so much for doing it!

No prob! Seems like the rust package woes are resolved now that Rust 1.78 is available in the Fedora 40 package repos so I am updating the PR to remove some of the version restrictions I added there. 

What's the preferred workflow on this repo w.r.t. git commits? Should I squash all my commits on this branch into a single one before review+merge?

aaron-trout:
Ok, so the build is now working: https://copr.fedorainfracloud.org/coprs/trouter/bazzite-multilib/build/7436938/

Here's a summary of what I have learned:

- rust package proc-macro2 added [Literal::byte_character](https://github.com/dtolnay/proc-macro2/commit/5ac12c94ef188c16a048461d65af2f245641dcac) and [Literal::c_string](https://github.com/dtolnay/proc-macro2/commit/70a804be6d9b21d8707a1c349759f64332429c35) in 1.0.80
- These are some sort of experimental/unsupported thing in rust-lang until the next release of the language which will be 1.79
- If you look at those commits linked above, those features are supposed to be automatically disabled on rustc < 1.79

```rust
   if rustc < 79 {
        println!("cargo:rustc-cfg=no_literal_byte_character");
    }
```

However, when we try to build mesa in COPR we get the warnings about using "unstable library feature". This is the same as described here: https://users.rust-lang.org/t/cannot-compile-simple-tokio-example/109956

I don't know exactly why that is happening, because it looks like this should just work automatically. So, I tried to downgrade `proc-macro2` to 1.0.80. I managed to get that working for the x86_64 build but for some reason the build dependencies failed to install on the i386 version. It seems like the Fedora package repos don't keep historical versions so you cannot easily pin to an older version.

Next topic is meson subprojects. Looks like mesa actually defines a bunch of these rust dependencies directly in the meson build system: https://gitlab.freedesktop.org/mesa/mesa/-/blob/406dda70e7c9baa59c975eb64025e7c3b210c3bc/subprojects/proc-macro2.wrap#L2

So I added the `--wrap-mode=default` and `--force-fallback-for=proc-macro2,syn` flags to the meson build in `mesa.spec` which means we directly fetch the rust crate and install from there, rather than using the Fedora system packages. This now makes the build green, but I'm not sure if it's acceptable to _not_ use the Fedora packages 🤷

Lastly the `%rewrite_wrap_file`... I'm not sure what purpose this currently serves, maybe you have some context on that @KyleGospo? From what I understand so far, the wrap files are only relevant if we are _not_ using system packages, and relying on those meson subprojects instead. Which I don't think is the case, because it completely refused to download anything mentioned in the wrap files until I added `--wrap-mode=default` (It would give an error that automatic wrap-based downloading is disabled or something like that). So, maybe we can just remove that thing?

I built an image based off `bazzite-gnome` which installs Mesa 24.1 from my COPR repo. If anybody wants to blindly trust me and run that image on their PC, it's at `eu.gcr.io/trouter-public/bazzite-mesa-test:latest`. Here's the Containerfile I used:

```
FROM ghcr.io/ublue-os/bazzite-gnome:40
ARG FEDORA_MAJOR_VERSION="${FEDORA_MAJOR_VERSION:-40}"

RUN rpm-ostree override remove mesa-va-drivers-freeworld mesa-vdpau-drivers-freeworld && \
    curl -Lo /usr/bin/copr https://raw.githubusercontent.com/ublue-os/COPR-command/main/copr && \
    chmod +x /usr/bin/copr && \
    curl -Lo /etc/yum.repos.d/_copr_trouter-bazzite-multilib.repo https://copr.fedorainfracloud.org/coprs/trouter/bazzite-multilib/repo/fedora-"${FEDORA_MAJOR_VERSION}"/trouter-bazzite-multilib-fedora-"${FEDORA_MAJOR_VERSION}".repo?arch=x86_64 && \
    rpm-ostree override replace \
    --experimental \
    --from repo=copr:copr.fedorainfracloud.org:trouter:bazzite-multilib \
    mesa-filesystem \
    mesa-libxatracker \
    mesa-libglapi \
    mesa-dri-drivers \
    mesa-libgbm \
    mesa-libEGL \
    mesa-libEGL-devel \
    mesa-vulkan-drivers \
    mesa-libGL \
    mesa-libOSMesa \
    mesa-va-drivers \
    mesa-vdpau-drivers && \
    ostree container commit
```

aaron-trout:
```bash
$ glxinfo | grep -i mesa | grep version
OpenGL core profile version string: 4.3 (Core Profile) Mesa 24.1.0-rc3
OpenGL version string: 4.3 (Compatibility Profile) Mesa 24.1.0-rc3
OpenGL ES profile version string: OpenGL ES 3.2 Mesa 24.1.0-rc3
$ vulkaninfo | grep -i mesa
VK_LAYER_MESA_device_select (Linux device selection layer) Vulkan version 1.3.211, layer version 1:
        driverID                                             = DRIVER_ID_MESA_NVK
        driverInfo                                           = Mesa 24.1.0-rc3
        driverID                                             = DRIVER_ID_MESA_LLVMPIPE
        driverInfo                                           = Mesa 24.1.0-rc3 (LLVM 18.1.1)
```

🥳

aaron-trout:
RC4 build: https://copr.fedorainfracloud.org/coprs/trouter/bazzite-multilib/build/7448569/

KyleGospo:
@aaron-trout Thanks for all the work on this! Can't wait to get stable NVK out to people

aaron-trout:
Ah, you beat me to it I was just trying to update the branch for 24.1 final and my push was rejected 😅

There was still some commented out stuff in here, and I think because of https://github.com/ublue-os/bazzite/blob/main/spec_files/mesa/mesa-28923.patch we can now remove the workaround I had related to `proc-macro2`. 

I'll test it out and raise another PR to tidy that up.

PR #1094, opened 2024-05-08
chore(gnome): remove MUTTER_DEBUG_FORCE_KMS_MODE

trgeiger:
This is no longer needed for VRR in v40+ and it breaks direct scanout.


KyleGospo:
Thanks!

Issue #959, opened 2024-04-08
Duperemove service


PVermeer:
### Describe the bug

`sqlite3` Is not installed on image for:
https://github.com/ublue-os/bazzite/blob/aa648f57c04ddb21df1b39fbe3b3b0bf240112a2/system_files/desktop/shared/usr/lib/systemd/system/btrfs-dedup%40.service#L13

```bash
apr 08 14:42:46 desktop-bazzite cp[46202]: cp: cannot stat '/var/home/.duperemove.hash': No such file or directory
apr 08 14:43:05 desktop-bazzite compsize[46204]: Processed 593523 files, 3216161 regular extents (3620099 refs), 246036 inline.
apr 08 14:43:05 desktop-bazzite compsize[46204]: Type       Perc     Disk Usage   Uncompressed Referenced
apr 08 14:43:05 desktop-bazzite compsize[46204]: TOTAL       87%      635G         730G         769G
apr 08 14:43:05 desktop-bazzite compsize[46204]: none       100%      521G         521G         525G
apr 08 14:43:05 desktop-bazzite compsize[46204]: zstd        54%      113G         208G         244G
apr 08 14:43:05 desktop-bazzite compsize[46204]: prealloc   100%      259M         259M          59M
apr 08 14:43:05 desktop-bazzite sh[46890]: sh: line 1: sqlite3: command not found
apr 08 14:43:05 desktop-bazzite sh[46889]: sh: line 1: [: : integer expression expected
apr 08 14:43:05 desktop-bazzite duperemove[46892]: Adding exclude pattern: /var/home/.duperemove.hash
apr 08 14:43:05 desktop-bazzite duperemove[46892]: Adding exclude pattern: /var/home/@swapfile/swapfile
apr 08 14:43:05 desktop-bazzite duperemove[46892]: process_extents: unable to get extent
apr 08 14:43:05 desktop-bazzite duperemove[46892]: file /var/home/pimv/.config/Code/GPUCache/data_0 changed
apr 08 14:43:05 desktop-bazzite duperemove[46892]: process_extents: unable to get extent
apr 08 14:43:05 desktop-bazzite duperemove[46892]: file /var/home/pimv/.config/Code/GPUCache/data_1 changed
apr 08 14:43:05 desktop-bazzite duperemove[46892]: process_extents: unable to get extent
apr 08 14:43:05 desktop-bazzite duperemove[46892]: file /var/home/pimv/.config/Code/DawnCache/data_1 changed
[...]

PR #1099, opened 2024-05-09
fix(desktop): sqlite column error in btrfs-dedup service

PVermeer:
I think since I reported #959 `sqlite` still errors in the `btrfs-dedup@.service`:

https://github.com/ublue-os/bazzite/blob/b5f4b6d179993d39c5f607958a91f2a330aa3ac6/system_files/desktop/shared/usr/lib/systemd/system/btrfs-dedup%40.service#L13

```sh
ei 09 13:39:31 desktop-bazzite systemd[1]: Starting btrfs-dedup@var-home.service - Btrfs deduplication on /var/home...
mei 09 13:39:59 desktop-bazzite compsize[18933]: Processed 725459 files, 4077370 regular extents (4581627 refs), 322845 inline.
mei 09 13:39:59 desktop-bazzite compsize[18933]: Type       Perc     Disk Usage   Uncompressed Referenced
mei 09 13:39:59 desktop-bazzite compsize[18933]: TOTAL       85%      727G         847G         895G
mei 09 13:39:59 desktop-bazzite compsize[18933]: none       100%      573G         573G         579G
mei 09 13:39:59 desktop-bazzite compsize[18933]: zstd        56%      154G         274G         315G
mei 09 13:39:59 desktop-bazzite compsize[18933]: prealloc   100%      288M         288M          82M
mei 09 13:39:59 desktop-bazzite sh[19131]: Error: in prepare, no such column: version_major
mei 09 13:39:59 desktop-bazzite sh[19131]:   SELECT keyval FROM config where keyname=version_major
mei 09 13:39:59 desktop-bazzite sh[19131]:                             error here ---^
mei 09 13:39:59 desktop-bazzite sh[19130]: sh: line 1: [: : integer expression expected
mei 09 13:39:59 desktop-bazzite duperemove[19134]: Adding exclude pattern: /var/home/.duperemove.hash
mei 09 13:39:59 desktop-bazzite duperemove[19134]: Adding exclude pattern: /var/home/@swapfile/swapfile
mei 09 13:39:59 desktop-bazzite duperemove[19134]: process_extents: unable to get extent
mei 09 13:39:59 desktop-bazzite duperemove[19134]: file /var/home/pimv/.config/Code/GPUCache/data_0 changed
mei 09 13:39:59 desktop-bazzite duperemove[19134]: process_extents: unable to get extent
```
This result in removal of the hash file for `duperemove` and it starts from scratch every time and this takes about 4 hours on my main system. After applying this fix, it does it in about 3-5 minutes on a second run.

I tested a bit in the terminal with different shell commands that should do the same and found it behaves differently when it runs in the service. I don't know why this is... If anyone knows I would like to know.

Anyways, I found that `'version_major'` would loose its surrounding `'` in any case when also using `'` to capture the `sh -c` command so `"` has to be used for this. I also found that when using `sh -c` in a service, a nested `"` has to be escaped with `\`.

I tested the command in the service by increasing the `-ge 4` expression and `||` echoing some string. The expression works fine with this fix.

### DIFF
--- a/system_files/desktop/shared/usr/lib/systemd/system/btrfs-dedup@.service
+++ b/system_files/desktop/shared/usr/lib/systemd/system/btrfs-dedup@.service
@@ -10,7 +10,7 @@ ExecCondition=sh -c '[ "$(stat -f -c "%%T" "$1")" = btrfs ]' _ %f/
 ExecStartPre=-find %f/ -mindepth 1 -maxdepth 1 -mtime +14 -name .duperemove.hash -exec rm -f '{}' \;
 ExecStartPre=-cp -a %f/.duperemove.hash duperemove.hash
 ExecStartPre=-compsize %f/
-ExecStart=-sh -c '[ "$(sqlite3 -readonly duperemove.hash "SELECT keyval FROM config where keyname=''version_major''")" -ge 4 ] || rm -f duperemove.hash'
+ExecStart=sh -c "[ $(sqlite3 -readonly duperemove.hash \"SELECT keyval FROM config where keyname == 'version_major'\") -ge 4 ] || rm -f duperemove.hash"
 ExecStart=duperemove -r -d -h -q --hashfile=duperemove.hash --skip-zeroes --exclude="%f/.duperemove.hash" --exclude="%f/@swapfile/swapfile" %f/
 ExecStartPost=-compsize %f/
 ExecStopPost=-cp -a duperemove.hash %f/.duperemove.hash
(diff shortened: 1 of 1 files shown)


PVermeer:
I also have a question about this line:
https://github.com/ublue-os/bazzite/blob/b5f4b6d179993d39c5f607958a91f2a330aa3ac6/system_files/desktop/shared/usr/lib/systemd/system/btrfs-dedup%40.service#L10

It seems that it would delete the hash file after 14 days that the service has not run.
If I go on holiday for two weeks or not use my handheld for a while it would start from scratch again and take 4 hours again to check every file.

KyleGospo:
Agreed, we can extend that.

PR #1105, opened 2024-05-11
feat: Trigger CEC onboot on Bluetooth device connection

AkazaRenn:
Added a new udev rule which is triggered when a Bluetooth device is connected. It calls `/usr/bin/cec-control onboot`, same as `cec-onboot.service`, should set the TV's input source to Bazzite. Useful when you just grab the controller and turn it on to start gaming.


KyleGospo:
This is really cool, thank you

Issue #1114, opened 2024-05-12
Bazzite Portal should probably have an option to setup Nvidia kernel arguments


AugustDeer:
I spent a ton of time troubleshooting why nothing could use my Nvidia card, before realizing I needed to run `ujust nvidia kargs`. I assume there's a reason this isn't configured automatically in setup, but it should at least be in the Bazzite Portal so it's clear it's an essential setup step.

KyleGospo:
This is automatically set by the hardware setup script already, and you should not have needed to do this.

PR #1115, opened 2024-05-13
feat(nvidia): Set Nvidia kernel arguments in Bazzite Portal

AugustDeer:
Allows setting necessary kernel arguments in the Bazzite Portal, since it's a necessary step to use Nvidia graphics.

Resolves ublue-os/bazzite#1114


KyleGospo:
Thanks but we prefer this be done in bazzite-hardware-setup as it is now.

AugustDeer:
Oh, is this meant to be done automatically? Because it wasn't on my machine.

PR #1128, opened 2024-05-16
feat: homebrew on image

m2Giles:


m2Giles:
Before merging this, instead of using an overlayfs we could just copy over the brew install from the image into place. While it will be a duplicate on image, it would avoid possible headaches of the user running brew update. We could also not run the brew update command, merely put the brew tree in place and they will download the portable ruby in the mutable location.

m2Giles:
I transitioned it to the copy method. This is much smaller and means that brew will behave exactly as expected as if you installed it yourself.

bketelsen:
nice!!

PR #1165, opened 2024-05-23
chore(mesa): remove rpmbuild/meson workarounds in mesa.spec

aaron-trout:
Fixing up some issues leftover from #1082 

- Reverting back to Fedora packages for `proc-macro2` and `syn` (the recent addition of `spec_files/mesa/mesa-28923.patch` should fix that issue)
- Reintroduce `%rewrite_wrap_file` - that does seem to fix an issue where meson cannot find the `syn` package


aaron-trout:
COPR build off this commit: https://copr.fedorainfracloud.org/coprs/trouter/bazzite-multilib/build/7482721/

PR #1171, opened 2024-05-25
feat: add fightcade to yafti

noelmiller:
Fightcade is a great platform for playing old fighting games. You can also play with other people online through their platform.


HikariKnight (approved):
LGTM

PR #1180, opened 2024-05-26
fix(user-setup): Improve language-sensible templates setup

Zeglius:
Use 'xdg-user-dirs' to obtain a localized path to templates directory (with '$HOME/Templates' as fallback. Should fix any related issues with Gnome nautilus not detecting templates using other language that english.


KyleGospo:
Thank you!

PR #1183, opened 2024-05-26
feat(waydroid): pass arguments for waydroid-launcher through to waydroid

hoshsadiq:
This PR ensures waydroid-launcher is able to pass through the arguments given to `waydroid-launcher`. This means waydroid-launcher can be used for more than just launching the full UI, such as running a single app. Thus, this means additional desktop entries could be created running individual apps, which can then be added to Game Mode.

This also fixes a small bug in `waydroid-fix-controller`. The globbing for some reason doesn't work properly, thus it doesn't seem to work well. This uses `tee` instead to pipe the output using sudo.


EyeCantCU (approved):
Thank you!

PR #1184, opened 2024-05-27
fix(ujust): sync arch packages before installing opentabletdriver

retchohrips:
Currently, running `ujust install-opentabletdriver` when you don't already have an Arch distrobox will create one and try to install opentabletdriver. This will fail because the repos are out of sync and it can't locate the files needed. This should fix that by running `paru -Syu --noconfirm` before trying to install opentabletdriver.


KyleGospo:
Thanks!

Issue #1204, opened 2024-06-06
sunshine-workaround.service bad unit file


mjpaden:
### Describe the bug

When I do a systemctl list-units, sunshine-workaround.service shows that it did not load due to a "bad-setting" (Reason: Unit sunshine-workaround.service has a bad unit file setting.) I think the second ExecStartPre line is missing an endquote.

### What did you expect to happen?

All built-in services to be loaded do so without error.

coquizen:
I have experienced the same error. I even tried running `systemctl edit sunshine-workaround.service` to override but even with an empty `ExecStartPre=` prior to the quote balance fixed line doesn't work.

PR #1226, opened 2024-06-14
fix(sunshine): fix systemd file for missing closing quote

coquizen:
fix(sunshine): sunshine-workaround missing closing quote in ExecStartPre line preventing proper execution. See #1204


HikariKnight (approved):
thanks!

PR #1242, opened 2024-06-18
feat: provide script to detect known cards requiring drm workaround

HikariKnight:
This covers for weird things like:
* Having sddm login screen enabled
* Having multiple users (if point 1 is true)
* The gamescope-session user is not uid 1000
* Having VULKAN_ADAPTER defined through `export-gpu` (if none is defined, then it will check the gpu marked as `Default: yes`)
* Ignores gpus used for VFIO as switcherooctl ignores any card bound to the vfio-pci driver

The script exits with exit code 0 if the card requires DRM workaround (currently if a polaris card is detected) and exit code 1 if nothing needs to be done.

NOTE: this is still missing RX 450-460 cards as i do not know the output for those. but they should hopefully get caught by the Ellemeres and Lexa Pro regex checks

### DIFF
--- a/system_files/desktop/shared/usr/libexec/hwsupport/gamescope-drm-workaround
+++ b/system_files/desktop/shared/usr/libexec/hwsupport/gamescope-drm-workaround
@@ -12,10 +12,8 @@ FINAL_RE="$POLARIS_RE"
 # Get the users homedirs so we can check for export-gpu configs
 HOMES=$(grep "/home" /etc/passwd | awk -F ":" '{ print $6 }')
 
-# Use switcherooctl to find all GPUs and conveniently does not list gpus used by vfio-pci <3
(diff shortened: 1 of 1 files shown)


HikariKnight:
After talking to you @KyleGospo i have now modified this to just `exit 0` if any card matches if `VULKAN_ADAPTER` is not defined
If `VULKAN_ADAPTER` is set through an env config in the users home then it will only check that GPU.
This means that should a user have a polaris card they use for GPU passthrough or use it for other tasks, they can just `export-gpu` to get back to he newer backend.

This also makes it possible to upstream our workaround to ChimeraOS as the requirement for switcherooctl is no longer present.

KyleGospo:
See: https://github.com/KyleGospo/gamescope-session/commit/b90d0a69c0a8c2f7d3df1988944ffbe29c2e1d1e

PR #1255, opened 2024-06-22
fix(wine):Match Fedora wine's configuration

trbocode:
Wine is in wow64 mode which seems to often not work and seemed to cause some people trouble with lutris, by default  Fedora sets wine and wine preloader to be 32 bit executables

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -433,8 +433,8 @@ RUN rpm-ostree install \
         libobs_glcapture.i686 \
         mangohud.x86_64 \
         mangohud.i686 && \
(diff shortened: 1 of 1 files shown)


KyleGospo:
This would break any 64 bit executable, no?

trbocode:
No, since it works in Fedora and the wine64 executable is the one for the WOW64 mode (As in, compiled with 64 bit libs), which is experimental

Current implementation often crashes with missing files errors, which seem to be common on 64 bit wine

PR #1257, opened 2024-06-22
feat: generalize certain SYS_IDs

HikariKnight:
This one is dedicated to the ROG Ally RC71L and its many silent revisions with weird name changes that make no sense.

`/usr/libexec/hwsupport/sysid` will in most cases just return the value of `cat /sys/devices/virtual/dmi/id/product_name`
however it will Trunctate or return generalized SYS_IDs for some hardware we do not need to differentiate between like the `ROG Ally RC71L` and `ROG Ally RC71L_RC71L` and `ROG Ally RC71L_Action`
So when ASUS silently make a revision to the ROG Ally again, we dont need to update the scripts to account for that (unless there is an actual hardware change we need to account for, then we can just check `/sys/devices/virtual/dmi/id/product_name` directly)

TLDR; we are now ready for ASUS to silently release the new Rog Ally revision named `ROG Ally RC71L_RC71L_Action_UNGABUNGA_X`


KyleGospo:
Leaving a note here to also change this in the GameScope session

HikariKnight:
Yeah only trunctate/generalize the SYS_ID where the hardware changes between the revision does not matter (see ROG Ally).
doubt it will be useful for the ayaneo and AOK devices, but the door is open to expand it as needed.

PR #1261, opened 2024-06-23
feat(fcitx5): Add advanced Chinese input method `fcitx5-chinese-addons`

p3nj:
## Main Purpose
Add `fcitx5-chinese-addons` package under fcitx5-mozc to enable advanced input method for Chinese users. (Shuangpin, Wubi..)

## Why
Fcitx5 is a widely used input method framework among Chinese users. It supports various input methods such as Shuangpin and Wubi, which significantly speed up the typing process. Shuangpin allows users to input Chinese characters with just two keystrokes, while Wubi uses stroke-based inputs.

To enable these input methods, an additional Fcitx5 package needs to be layered, similar to `fcitx5-mozc`
image

Once the package is layered, the package should function correctly.
image

## File Changed
+ [Containerfile:488]

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -485,6 +485,7 @@ RUN if grep -q "kinoite" <<< "${BASE_IMAGE_NAME}"; then \
             rom-properties-kf6 \
             joystickwake \
             fcitx5-mozc \
+            fcitx5-chinese-addons \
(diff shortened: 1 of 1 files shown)


KyleGospo:
Thank you!

PR #1270, opened 2024-06-24
Update cosign.pub

wangzheng2008-Lenovo:


HikariKnight:
Pretty sure this was opened in error

PR #1279, opened 2024-06-25
feat(tdpfix): add detection for multiple cards

HikariKnight:
Will now fix the TDP on systems with multiple cards installed


misteryolo:
May i ask, when does this fix come out?

HikariKnight:
> May i ask, when does this fix come out?

as always, whenever the next successful build for `main` happens as there is currently a package skew :upside_down_face: 
https://github.com/ublue-os/bazzite/actions/workflows/build.yml?query=branch%3Amain

rather have the package skew happen in github than on your system...

PR #1299, opened 2024-06-30
fix: Pre-Install Exhibit on existing installations

nicknamenamenick:


KyleGospo:
I'm going to say as a policy we no longer touch flatpaks going forward.

New installs of course are fine to customize, but let's leave existing up to the end user.

Issue #1306, opened 2024-07-01
Documentation refers to "ujust restore-original-terminal" instead of "_ujust restore-original-terminal"


moritz-john:
### Describe the bug

When you start the Ptyxis Terminal you sometimes see this tip:
```

   Welcome to Bazzite 󰊴

[...]

  󰋼 Rather use KDE Konsole? You can restore the application icon for it with
  ujust restore-original-terminal . Be sure to also change the default terminal
  in System Settings.
```

It says to use `ujust restore-original-terminal` to switch to KDE Konsole.

But when entering `ujust restore-original-terminal` this happens:
```
user@DESKTOP-endeavour:~$ ujust restore-original-terminal
error: Justfile does not contain recipe `restore-original-terminal`.
Did you mean `_restore-original-terminal`?
user@DESKTOP-endeavour:~$
```

### What did you expect to happen?

I expected the advertised `ujust restore-original-terminal` command to work and not return an error.

### Extra information or context

https://github.com/ublue-os/bazzite/blob/main/README.md 
> [Ptyxis Terminal](https://gitlab.gnome.org/chergert/ptyxis) used as the default in all images. This terminal is specifically designed for the container workflow you'll use in Bazzite. If you wish to return to stock, run `ujust restore-original-terminal`

Here the command is also wrongly documented (`ujust restore-original-terminal` instead of `_ujust restore-original-terminal`)

moritz-john:
I think the `ujust restore-original-terminal` command got changed here: https://github.com/ublue-os/bazzite/commit/0c6ad9480c8eb6ac71c757dcc8fb58c090814feb but the associated documentation/tips were never updated.

nicknamenamenick:
Thanks for pointing this out.  It should be fixed in the next build hopefully.

moritz-john:
Nice! I think the same documentation change needs to be made here: https://github.com/ublue-os/bazzite?tab=readme-ov-file#about--features

> * [Ptyxis Terminal](https://gitlab.gnome.org/chergert/ptyxis) used as the default in all images. This terminal is specifically designed for the container workflow you'll use in Bazzite. If you wish to return to stock, run `ujust restore-original-terminal`

and in `README-zh-cn.md` + `README-BR.md` + `README-FR.md` + `README-SPA.md` & `README-NL.md` according to this quick search: https://github.com/search?q=repo%3Aublue-os%2Fbazzite%20ujust%20restore-original-terminal&type=code

[...]

PR #1307, opened 2024-07-01
chore(motd): updated command to unhide konsole

nicknamenamenick:
Fixes https://github.com/ublue-os/bazzite/issues/1306.


KyleGospo:
Let's drop this MOTD IMO, that was the intent behind making the ujust hidden

PR #1336, opened 2024-07-06
feat(ujust): Add install-resolve

Zeglius:


system_files/desktop/shared/usr/share/ublue-os/just/82-bazzite-apps.just
```diff
@@ -157,6 +157,46 @@ install-opentabletdriver:
     systemctl enable --user --now arch-opentabletdriver.service && \
     distrobox enter -n arch -- bash -c 'distrobox-export --app otd-gui'
 
+# Install/update Davinci Resolve, a close source video editing utility
```
tristann21, line 160:
```suggestion
# Install/update Davinci Resolve, a closed-source video editing utility
```

tristann21 (commented):
change "close source" to "closed-source"

system_files/desktop/shared/usr/share/ublue-os/just/82-bazzite-apps.just
```diff
@@ -157,6 +157,48 @@ install-opentabletdriver:
...
+    if podman container inspect davincibox >/dev/null 2>&1; then 
+      echo "Removing existing davincibox container"
+      podman stop --ignore davincibox && podman rm davincibox
+    fi
+    DOWNLOADDIR=$(xdg-user-dir DOWNLOAD || echo ${HOME})
+    echo "${bold}This script requires you to download DaVinci Resolve installer manually from their website and place it in ${HOME} or ${DOWNLOADDIR}${n}"
+    echo "${bold}https://www.blackmagicdesign.com/event/davinciresolvedownload${n}"
+    # Download davincibox setup script
+    SETUPSCRIPT=/tmp/davincibox_setup.sh
+    wget -O $SETUPSCRIPT https://raw.githubusercontent.com/zelikos/davincibox/main/setup.sh
```
p5, line 175: From a supply-chain point of view, it would be good if this script was pinned to a specific commit since it's not managed by ublue

p5 (approved):
(Am not able to run the script as I'm not at a Linux machine)

The script looks good to me!  Nice one!

Let's just get another pair of eyes from @KyleGospo

fiftydinar:
I don't think that there should be an immediate opt-out destructive operation like destroying the existing Resolve container, I think that there should be a prompt, which would ask you if you want to install Davinci Resolve or not, with status shown if it's installed or not.

Zeglius:
> I don't think that there should be an immediate opt-out destructive operation like destroying the existing Resolve container, I think that there should be a prompt, which would ask you if you want to install Davinci Resolve or not, with status shown if it's installed or not.

Added some prompts to install/uninstall at commit 0c2a83a
Not only that, but seems simply removing the container will leave some leftovers at HOME. So using [davincibox scripts](https://github.com/zelikos/davincibox/blob/58d52028e8a51ed868303941eee253e68b3355df/system_files/usr/bin/add-davinci-launcher#L13) to handle that was also needed.
Fixed that with commit 98500bf.

KyleGospo (approved):

PR #1337, opened 2024-07-06
feat: Container indicator for fish shell

v7rzlhb3xg:
Modified default fish shell prompt to indicate if a user is inside a container with a box icon and the container name, similar to how bash looks inside a distrobox container. Not sure if there was a better way to do this, but it works!

Here's a comparison of the stock fish shell vs when I'm in a container I made for LLM stuff.

Screenshot from 2024-07-06 15-24-32
Screenshot from 2024-07-06 15-24-40


KyleGospo:
This is very cool, thank you!

KyleGospo (approved):

castrojo (approved):

PR #1352, opened 2024-07-11
feat(ujust): add install-zed

Zeglius:


KyleGospo:
This is great, but let's get it in https://github.com/ublue-os/config
I'd rather Bazzite stick to apps that have a direct use case for gaming, and more generic stuff like Zed be usable everywhere and maintained in one place.

PR #1367, opened 2024-07-15
feat: Install `wqy-zenhei-fonts` for GoldSrc engine to display Chinese fonts

p3nj:
## Main Purpose
Add `wqy-zenhei-fonts` package under `google-noto-cjk-fonts` to let GoldSrc engine display Chinese characters. 

## Why
Old GoldSrc engines (Half-Life, Counter-Strike 1.6, Condition Zero) and potentially some Source 1 games, are using `wqy-zenhei` to display Chinese characters. On default installation, the package is not included resulting in blank characters being displayed.
image

image

It's the well-known solution being mentioned multiple times on [Source-1 repo](https://github.com/ValveSoftware/Source-1-Games/issues?q=wqy-zenhei) and somewhere on the internet that I'm not able to find discussion threads anymore.

After layering the package `wqy-zenhei-fonts` the results are shown below.
image

image

## Side note
It also fixes not just Chinese but also special symbols.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -322,6 +322,7 @@ RUN rpm-ostree install \
         xwiimote-ng \
         twitter-twemoji-fonts \
         google-noto-sans-cjk-fonts \
+        wqy-zenhei-fonts \
         lato-fonts \
         fira-code-fonts \
         nerd-fonts \


KyleGospo (approved):

m2Giles (approved):

KyleGospo:
Thank you!

szescxz:
Can we consider making this fix optional (e.g. as a `ujust` command)? Unless stock SteamOS is also including this font by default.
Just kind of surprised that Gaming Mode UI is also picking this font in CJK languages.

KyleGospo:
@szescxz @p3nj This is reverted now, let's figure out how to fix this without the regression

p3nj:
I can make the `ujust` script, is the gaming mode UI the HTPC image?
Currently, I'm only using the desktop version. can you provide more info about the Gaming Mode UI? Thanks a bunch.

szescxz:
> can you provide more info about the Gaming Mode UI?

@p3nj Using the `deck` variant here, although I think you should be able to reproduce it with Steam's Big Picture mode. It's just another Chromium-based stuff anyway so not really something mysterious here.

p3nj:
wqy_zenhei_fonts
non-wqy_zenhei_fonts

szescxz:
If you are expecting something obvious like tofus as in your original issue, that's not the case here.
The difference is exactly the font selection - WenQuanYi is getting picked up if installed, instead of Noto.
An obvious difference would be `将` in your second test message: Japanese glyph is being used with Noto, instead of Chinese glyph with WenQuanYi. _Glyph issues would be another story, though._

As for why font selection matters here:
1. At least for the `deck` variants, Bazzite's behavior should stay close to SteamOS as much as possible, unless there's a good reason not to. We could blame Valve for not hardcoding Noto as one of the fallback fonts in their CSS, but still.
2. Might be just me, but with WenQuanYi all CJK stuff gets a lighter font weight which is even more terrible for display.

Also, since your fix only applies to games with specific engines, it's better to put it in the FAQ ([example](https://github.com/ublue-os/bazzite/blob/57f51690da6b84e66bd56c7d6eb379b91446546a/system_files/desktop/shared/usr/share/ublue-os/docs/Gaming/Common%20Gaming%20Issues.md?plain=1#L20)) or [`ujust`](https://github.com/ublue-os/bazzite/blob/main/system_files/desktop/shared/usr/share/ublue-os/just/81-bazzite-fixes.just). Since Bazzite is `rpm-ostree`-based, removing an unwanted fix could be more difficult than adding one you need: https://universal-blue.discourse.group/t/removing-base-package-using-rpm-ostree/2029

p3nj:
Thanks for the detailed explanation. 
As you mentioned, Bazzite's behavior should be close to SteamOS unless there is a significant reason otherwise. Including this in the FAQ or `ujust` sounds like a good idea.

Issue #1289, opened 2024-06-28
Controller does not turn off


pdien:
### Describe the bug

Whether in Game Mode or the desktop, the connected controller will not go to sleep. Auto-sleep function is set to 15 minutes and it still stays on past that time, draining the battery. HOME + △/Y also does not turn the controller off.

### What did you expect to happen?

Controller should be turning off.

### Output of `rpm-ostree status`

```shell
State: idle
Deployments:
● ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-deck:stable
                   Digest: sha256:b48da348e0f942a18e833fb14f5483d70b291b41f34089c390148ad53a36dec4
                  Version: 40.20240625.0 (2024-06-27T02:01:04Z)
          LayeredPackages: sunshine

  ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-deck:stable
                   Digest: sha256:b48da348e0f942a18e833fb14f5483d70b291b41f34089c390148ad53a36dec4
                  Version: 40.20240625.0 (2024-06-27T02:01:04Z)
          LayeredPackages: sunshine
```

### Hardware

![image](https://github.com/ublue-os/bazzite/assets/22202244/4452735b-2626-4933-a3f6-57506b4300d5)

### Extra information or context

Controllers used: DualSense, Dualshock 4, Xbox One, Xbox Series X.
Turning off bluetooth does turn them off, but this is not the ideal use case.

thekk1:
The workaround #1390 will not handle this issue.
Especial not the part with the shortcut HOME + △/Y.

thekk1:
I can confirm this behavior on my machine UM790pro.

pdien:
This might be a general Linux problem, as suggested by S1ngl3-x in #1040. I installed a fresh Ubuntu image, installed Steam, connected a DualSense controller and the HOME + △/Y shortcut still doesn't work.

thekk1:
This issue is tracked at steam under [#8678](https://github.com/ValveSoftware/steam-for-linux/issues/8678)

thekk1:
Who does not want to wait for the fix #1416 can try the script from here:
https://github.com/thekk1/bluetooth-ds4-ds5-workaround

Issue #1119, opened 2024-05-13
Xbox controller via xbox dongle not working after success of waking up PC


herbalizer404:
### Describe the bug

Xbox controller via xbox dongle not working after success of waking up PC.
I press xbox button, it wakes up the PC, xbox controller is well connected to dongle with button light not flashing as expected. But when I reach the gamemode controller not working, I cant move in the game mode UI.

### What did you expect to happen?

Xbox controller to work before and after suspend/resume.
It works before suspend, but not after resuming.

### Output of `rpm-ostree status`

```shell
State: idle
Deployments:
● ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-deck:40-stable-20240502
                   Digest: sha256:c02ece6452dd8b2017f1e3b687f8107cccd0d8a7453399fd9b642df9cdb5cca9
                  Version: 40.20240502.0 (2024-05-02T23:42:25Z)
          LayeredPackages: sunshine

  ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-deck:40-stable-20240505
                   Digest: sha256:349ddc3efa844187ca3b7f1ae16c880b00e18aaabf995d876ee6bf36554d22c0
                  Version: 40.20240505.0 (2024-05-05T05:45:58Z)
          LayeredPackages: sunshine
```

### Hardware

AMD ryzen 7600
AMD 7900 GRE
B650I Lightning WiFi

Issue #1040, opened 2024-04-28
Duelsense controller doesn't work after system or controller gos to sleep


Decker-01:
### Describe the bug

After the system goes into sleep mode or the Dualsense switches off, I cannot reconnect the Dualsense properly. The controller is not recognized correctly and does not adopt the LED settings that I have set up. Inputs in the UI are not possible. I do not have a similar problem with the Bluetooth mouse. When i restart the system, everythings works fine.

### What did you expect to happen?

The controller should be recognized after sleep mode or switching off. 

### Output of `rpm-ostree status`

```shell
State: idle
Deployments:
● ostree-unverified-registry:ghcr.io/ublue-os/bazzite-deck:latest
                   Digest: sha256:67bb1f9dd52a420f080fd67e71a60bb6f732c12bacab7d27476dfbdca601c757
                  Version: 40.20240427.0 (2024-04-27T22:36:59Z)
            LocalPackages: lact-0.5.4-0.x86_64
                Initramfs: '"-I /etc/crypttab /etc/modprobe.d/amdgpu.conf /etc/modprobe.d/deck-blacklist.conf" ' 

  ostree-unverified-registry:ghcr.io/ublue-os/bazzite-deck:latest
                   Digest: sha256:25e6c62e647e7bef2a461f24d0c7fea72a86fb95edb66bae00f350e42364bf55
                  Version: nightly (2024-04-26T17:04:40Z)
            LocalPackages: lact-0.5.3-0.x86_64
                Initramfs: '"-I /etc/crypttab /etc/modprobe.d/amdgpu.conf /etc/modprobe.d/deck-blacklist.conf" '
```

### Hardware

![grafik](https://github.com/ublue-os/bazzite/assets/166182283/aad12196-2112-48e6-bb7f-3de78693b133)

PS5 Dualsense Controller

S1ngl3-x:
I remember this to be general linux problem. I've just disabled auto sleep of the controller in steam.

waldhari81:
Same issue for me, xbox controller is gone after sleep mode.  Reboot fixes it.  I tried other controllers, the same issue.

thekk1:
> Same issue for me, xbox controller is gone after sleep mode. Reboot fixes it. I tried other controllers, the same issue.

This is not what this issue is about. We talk about turning the controller off, not the system.

Issue #993, opened 2024-04-19
Legion Go suspend not working while game is running


gyorgys:
### Describe the bug

While running Dishonored in Game Mode, pressing the power button will not suspend the system. The system will ostensibly try to go to suspend (screen blanks, power LED blinks) but it wakes up immediately.

No problem sleeping while in Steam UI

### What did you expect to happen?

Pressing the power button will should suspend the system.

### Output of `rpm-ostree status`

```shell
State: idle
Deployments:
● ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-deck:stable
                   Digest: sha256:b3e8d6e3cb7ed88de6dfe20dcf1f82a38eff42e02226d68897ea051a90ec7fa7
                  Version: 39.20240418.0 (2024-04-18T20:08:16Z)
                Initramfs: '"-I /etc/crypttab /etc/modprobe.d/amdgpu.conf /etc/modprobe.d/deck-blacklist.conf"' 

  ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-deck:stable
                   Digest: sha256:b3e8d6e3cb7ed88de6dfe20dcf1f82a38eff42e02226d68897ea051a90ec7fa7
                  Version: 39.20240418.0 (2024-04-18T20:08:16Z)
                Initramfs: '"-I /etc/crypttab /etc/modprobe.d/amdgpu.conf /etc/modprobe.d/deck-blacklist.conf"'
```

### Hardware

83E1

Lenovo Legion Go

### Extra information or context

I have been playing Dishonored before updating and suspend worked fine

Shaunathanz:
Weird question but can you try disconnecting the True Strike controllers and then try to suspend it? I've been having suspend issues since the last update and it doesn't seem to be game specific for me, but the controllers being connected (specifically the left one) seems to be the deciding factor when the issue is occurring.

gyorgys:
The issue went away with the 4/19 build

gyorgys:
The issue went away with the 4/19 build

gyorgys:
The issue is back on 4/19 build with a different game (CP77). I detached the controllers per @Shaunathanz suggestion and were able to suspend that way

PR #1390, opened 2024-07-21
feat: Apply bluetooth controller stuck on sleep workaround

Zeglius:
This solves the issue with controllers being kept powered on during suspend.
See ~~issue #1289~~, issue #1040 and potentialy issue #1119


KyleGospo:
Any regressions for this? Just trying to avoid breaking Bluetooth if we can, we had previous similar concepts to this that caused devices not to sleep.

Zeglius (commented):
> Any regressions for this? Just trying to avoid breaking Bluetooth if we can, we had previous similar concepts to this that caused devices not to sleep. 

Devices as in controllers/any bluetooth connected device or the PC/handheld? If its the later, we can add a timeout to prevent the service blocking the suspension for too long:
```
TimeoutSec=10s
```
Other than potentially issue #993 (which Im not sure if is exclusive to legion go), would need some point of reference for having bluetooth get restarted impeding suspension of the device

thekk1 (approved):

thekk1:
#1289 can not be affected by this workaround because the issue has nothing to do with the sleep mode.

thekk1 (commented):
It should be:
"Disconnect all bluetooth devices"
Because this is what you do.

system_files/desktop/shared/usr/lib/systemd/system/bazzite-bluetooth-sleep-workaround.service
thekk1, line 1: This should be changed to:
Disconnect all bluetooth devices before sleep

Zeglius: Solved commit e2590682e7ec4fca467eb14fa9b2b81e415d03b5

thekk1 (approved):

KyleGospo:
Thanks! Merging into testing because we've been bit by bluetooth changes before. Will get on Main if no issues are found.

KyleGospo (approved):

system_files/desktop/shared/usr/libexec/bazzite-bluetooth-sleep-workaround
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/bash
+
+devices=$(bluetoothctl devices Connected | awk '{print $2}')
+
+# shellcheck disable=SC2068
+for dev in $devices; do
+    [ -n "$dev" ] &&  bluetoothctl disconnect "$dev"
```
HikariKnight, line 7: Just adding this for reference if it will be needed.
when i had to turn off my windows VMs bluetooth adapter using a script at boot i had to do it something like
```bash
echo -e "select 00:E0:4C:77:99:88\npower off\n" | bluetoothctl
```
otherwise it would not reliably work.

not asking for a change, just adding it as a comment in case it ends up not activating reliably then this is something that can be tried :slightly_smiling_face:

HikariKnight (approved):

system_files/desktop/shared/usr/libexec/bazzite-bluetooth-sleep-workaround
HikariKnight: actually @Zeglius it will need some changes as this will fail on devices that would happen to have multiple bluetooth controllers.

see https://github.com/ublue-os/bazzite/pull/1416#issuecomment-2254628297

Zeglius: Solved with commit b4c19487f1ed00e128997f91caae52e8ec769066

HikariKnight (approved):

PR #1439, opened 2024-07-30
feat(rechunk): Introduce image rechunking to lower update sizes

antheas:
Adds image rechunking to the bazzite build. May or may not run as it is. Before being moved to testing, the action and docker image used should be pinned.

Especially since ref points to a local filesystem dir, which the upload action might not like.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,6 +29,12 @@ on:
       - ".github/workflows/build_iso.yml"
   merge_group:
   workflow_dispatch:
+    # TODO: Enable this when the prev tag is enabled in rechunk
+    # inputs:
+    #   fresh-rechunk:
+    #     description: 'Clear rechunk plan'
+    #     type: boolean
+    #     default: false
 env:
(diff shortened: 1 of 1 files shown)


HikariKnight (approved):
LETS GO!

KyleGospo (approved):

HikariKnight (approved):

KyleGospo (approved):

HikariKnight (approved):

HikariKnight (approved):

KyleGospo (approved):

antheas:
This is as close as I can get it today. Note that the changes had to be a bit more major.

`rechunk` uses podman mount which requires root, and the buildah action ran in rootless, so it had to be switched out. 

Then, for the other actions to work, incl. secure boot and pushing, the raw oci-dir image needs to be pulled into podman and tagged.

HikariKnight (approved):

HikariKnight (approved):

PR #1459, opened 2024-08-04
feat: Add btrfs-assistant to easily configure automatic snapshots of /home

HikariKnight:
This is more of an RFC to see what others think
This has been requested to be included in the past and I agree that it would be a useful inclusion.

We are setting the root config so it will work with the automatic snapper systemd units that btrfs-assistant configures, these expect a root config to exist.
These systemd units are disabled by default so the user will have to enable them themselves.

snapshotting / is pretty pointless on ostree so the root config provided points to /var/home instead.

btrfs-assistant provides a nice gui for handling btrfs tasks, snapshotting, managing subvolumes and also configuring snapper which is opensuses btrfs snapshotting utility (and it is good!).

## Tested
I have tested making snapshots, automatic snapshots (requires a config named `root` to exist as a minimum), restore, browsing and restoring single files, delete subvolumes, balance and scrub.

## Overview (Yes it has dark mode, i was just lazy when taking screenshots)
image

Look at subvolumes and manage them
image

Create/restore snapshots
image
image

Configure snapper (the systemd services are disabled by default)
With the provided config, if the user wants automatic snapshotting of their home folder they just need to open `btrfs-assistant` and enable the services, everything else is pre-configured.
Settings shown are the defaults i set for `/var/home`
image

Can also enable automatic btrfs maintenance (default disabled)
image


Zeglius:
What are the opinions of pointing  the root config to snapshot `/var` instead of `/var/home`?
As far as I'm aware, `/var/home` is not the only persistent mount point, there is also `/var/usrlocal`

HikariKnight:
> What are the opinions of pointing the root config to snapshot `/var` instead of `/var/home`? As far as I'm aware, `/var/home` is not the only persistent mount point, there is also `/var/usrlocal`

you can make configs for that (and it even suggests `/var` as a location since it is a subvolume), in general the user would by default be more interested in their home directory
rpm-ostree at least when you boot between ostrees tracks what you had in /etc and i believe /var when you boot into older updates.
we can not point it to `/var/usrlocal` as that is not a subvolume

HikariKnight:
ok making a default config is a bit more involved, changing this back to a draft until i am done playing around with this on my personal image.

HikariKnight:
I have found a workaround for https://gitlab.com/btrfs-assistant/btrfs-assistant/-/issues/105
If btrfs-assistant runs in xwayland it will not create issues with xdg-document-portal (which then breaks flatpaks)
a desktop file override has been commited in the PR until this is fixed upstream as the linked issue is only an issue on fedora 40.

HikariKnight:
We should agree on the default configs retention rules for snapshots here @KyleGospo 
```
# limits for timeline cleanup
TIMELINE_MIN_AGE="1800"
TIMELINE_LIMIT_HOURLY="8"
TIMELINE_LIMIT_DAILY="4"
TIMELINE_LIMIT_WEEKLY="16"
TIMELINE_LIMIT_MONTHLY="0"
TIMELINE_LIMIT_YEARLY="0"
```

unless we decide to use and patch and test timeshift from Linux Mint instead to make it work with the home subvolume as `/home` instead of `@home`

I do not mind which we use, but personally i prefer snapper from opensuse for snapshots as it is way more flexible in my opinion, and it has a nice cli to manage the snapshots too.

Timeshift would however give us the option  of rsync backing up `/var/home` while excluding Steam.
Snapshots however for both will still include `~/.local/share/Steam` as they are based on subvolumes as a whole.

PR #1416, opened 2024-07-27
fix: Add script to monitor and disconnect DS4|DS5 Bluetooth gamepads 

thekk1:
This is a workaround for #1289 and should be removed if there comes a native solution from https://github.com/ValveSoftware/steam-for-linux/issues/8678

- Added `bazzite-bluetooth-ds4-ds5-workaround` script to monitor PS4 and PS5 controllers
  - Script checks for connected gamepads, extracts their MAC addresses, and starts a monitoring process
  - The monitoring process listens for specific button combinations (Home + Triangle) to disconnect the gamepad
  - Includes cleanup functionality to stop monitoring processes for disconnected gamepads

- Created Systemd service `bazzite-bluetooth-ds4-ds5-workaround.service`
  - Service description and dependencies defined in `[Unit]` section
  - `[Service]` section specifies script execution, restart policy, user, and group
  - `[Install]` section enables the service to start on boot

- Added instructions to load, start, and enable the service
  - Included commands for reloading Systemd, starting the service, and enabling it on boot
  - Provided guidance on checking the service status and logs for troubleshooting

Author: thekk1
Version: 1.0

PR #1476, opened 2024-08-08
feat: Add script to disconnect dualshock/sense with HOME + TRIANGLE

Zeglius:
- Reimplementation of pr #1416 to Python
- Use udev rules to activate when connected
- Disabled by default, can be enabled with `ujust toggle-ds45-workaround`


Zeglius:
NOTE: This PR enters in conflict with #1484. Apply one or other.

HikariKnight (approved):
Having thought long and hard about this i think this one will be the easiest one to maintain

KyleGospo (approved):

Zeglius:
Forgot this PR existed 😅. Will have to do a rewrite sadly, given:

1. Workaround spans 3 files, too many:
    - script that tracks button press.
    - udev rule that detects DualShock/sense connected, then starts scripts passing it's MAC address as argument.
    - empty udev rule that nullifies the previous one.

2. Lack of reactivation mechanism
This is due the udev rule that starts the script only triggers once per session, requiring the script to keep running all the time, even when no DualShock/sense is not connected anymore.

3. Script is poorly written

PR #1479, opened 2024-08-09
feat(deck): Add Per-Core CPU Undervolt service

Tobe95:
Systemd service that does the same thing as `ryzenadj.service`, but allows the user to change curve optimization for each core individually via ryzenadj `--set-coper` option.


HikariKnight:
Not quite sure about including this, it is hardcoded to 4 cores and you cannot edit `/usr`, yes they can copy to `/etc` but the majority wont.
Plus i feel like this is something the user can add it themseves to `/etc/systemd/system` if they wanted it instead of having it in the image.

Would like a second opinion though @KyleGospo @m2Giles

Tobe95:
I didn't think of devices other than the Steam Deck, that was my bad...

What about just adding it to the deck image though? It would still be as easy to enable as the "normal" UV, for people that don't want to mess with it, but still gives the opportunity for further tweaking if desired.

m2Giles:
hard no.

image

This should be in `/etc/systemd/system` and needs actual checks beyond just sleep.

>  What about just adding it to the deck image though?

The deck image is not just for the Steam Deck.

system_files/desktop/shared/usr/lib/systemd/system/ryzenadj_coper.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Per-Core CPU Undervolt for Steam Deck
+
+[Service]
+Type=oneshot
+EnvironmentFile=-/etc/default/%p
+ExecStartPre=/usr/bin/sleep 10
```
Zeglius, line 7: Use `system_files/desktop/shared/usr/libexec/hwsupport/valve-hardware` to check the device is a steam-deck
```suggestion
ExecCondition=/usr/libexec/hwsupport/valve-hardware
```
```suggestion

**Comment by Zeglius** [comment]
Check is a steam-deck what we are running
```
[Service]
Type=oneshot
EnvironmentFile=-/etc/default/%p
ExecCondition=/usr/libexec/hwsupport/valve-hardware
```

PR #1483, opened 2024-08-10
feat(fish): enhanced fish performance

achengli:
**fish shell** take long time to be ready since **homebrew** vendor script must be executed in every shell instance.

I substitute the brew instance by a function that evaluates those commands which are slowing down the fish initial runtime only when you call brew _"when you need it"_


KyleGospo:
Looks good to me, thanks!

KyleGospo (approved):

HikariKnight (approved):

PR #1484, opened 2024-08-10
fix: Issue #1289 ds4|5 disconnect

thekk1:
Add script to monitor and disconnect Bluetooth gamepads and corresponding Systemd service.
This is a workaround for #1289 and should be removed if there comes a native solution from
https://github.com/ValveSoftware/steam-for-linux/issues/8678

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -701,6 +701,7 @@ RUN rm -f /etc/profile.d/toolbox.sh && \
     systemctl enable bazzite-hardware-setup.service && \
     systemctl enable tailscaled.service && \
     systemctl enable dev-hugepages1G.mount && \
+    systemctl --global enable bazzite-bluetooth-ds4-ds5-workaround.service && \
     systemctl --global enable bazzite-user-setup.service && \
     systemctl --global enable podman.socket && \
     systemctl --global enable systemd-tmpfiles-setup.service && \
--- a/system_files/desktop/shared/usr/lib/systemd/system/bazzite-bluetooth-ds4-ds5-workaround.service
+++ b/system_files/desktop/shared/usr/lib/systemd/system/bazzite-bluetooth-ds4-ds5-workaround.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Disconnect DS4 and DS5 controller on shortcut HOME + triangle
+After=network.target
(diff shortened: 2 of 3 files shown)


thekk1:
Because I messed up the original PR, I created a new clean one.

Zeglius:
Would suggest checking #1476 to see if is better suited. Not saying that this wouldn't work, but in my opinion is far easier to maintain a python script than an AWK one.

thekk1:
> Would suggest checking #1476 to see if is better suited. Not saying that this wouldn't work, but in my opinion is far easier to maintain a python script than an AWK one.

I choosed this way because of the performance not because of the maintainability. The task to check for the key combo will run all the time in backgroup while a DS4|5 Controller is connected so this a very important point.
I am not familiar enough with python but I see some points in #1476 which could cause a higher cpu load. I will check other day on this.

udev could might be a good point but I remember that starting an infinite prozess with udev should be a nono!?

Zeglius:
> I choosed this way because of the performance not because of the maintainability. The task to check for the key combo will run all the time in backgroup while a DS4|5 Controller is connected so this is a very important point

Same here with the python version, the udev rule will will start the service using `@` systemd unit parameters, so the script will be running all the time for that mac address, but only start when the device is connected.

> udev could might be a good point but I remember that starting an infinite prozess with udev should be a nono!?

It starts a systemd service with `--no-block`, so it will run at background and the lifespan of the script wont rely entirely in the udev rule, only to start the service.


Also must ask for the reason of your version running as a user service (`systemctl enable --global`) instead of a system one. 
Under my perspective, it would lead to have duplicate services running simultaneously, potentially including root user (which I saw being an headache with Handheld Daemon case).

If user service is a must by your rationale, might suggest moving the service to `/usr/lib/systemd/user/` and add a check for system users like this:

```ini
# /usr/lib/systemd/user/bazzite-bluetooth-ds4-ds5-workaround.service
[Unit]
Description=Disconnect DS4 and DS5 controller on shortcut HOME + triangle
After=bluetooth.target
ConditionFileIsExecutable=/usr/libexec/bazzite-bluetooth-ds4-ds5-workaround
ConditionUser=!@system
```

And change the `WantedBy` to `default.target`:
```ini
[Install]
WantedBy=default.target
```

PR #1494, opened 2024-08-12
feat(cli): Alias `xdg-open` to `open`

castrojo:
This just makes `open blah.pdf` and so on work via the command line. It opens whatever app is set for that mime type


p5 (approved):
One of the most-used things I PR'd to Bluefin :smile:

noelmiller:
oh man, this is awesome!

PR #1531, opened 2024-08-22
fix(decky): change Decky's default 45 sec timeout to 5 seconds

aarron-lee:
Decky, by default, has a 45 second timeout in it's systemd service. Decky doesn't seem to capture any signals such as SIGINT from the OS, and thus, is not shutting down gracefully.

This leads to perceived slowness when shutting down a device with Decky installed.

As a workaround, this PR changes the 45 second timeout with the `setup-decky` ujust

Decky issue for 45s timeout: https://github.com/SteamDeckHomebrew/decky-loader/issues/683

### DIFF
--- a/system_files/desktop/shared/usr/share/ublue-os/just/80-bazzite.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/80-bazzite.just
@@ -138,6 +138,8 @@ setup-decky ACTION="install":
       fi
       curl -L https://github.com/SteamDeckHomebrew/decky-installer/releases/latest/download/install_release.sh | sh
       sudo chcon -R -t bin_t $HOME/homebrew/services/PluginLoader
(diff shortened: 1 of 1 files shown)


KyleGospo:
Instead of a sed in the installer I think this can be done as a pre included override file so it applies to existing installs and exists in a file decky doesn't control.

KyleGospo:
Should be something like: 
`/etc/systemd/system/plugin_loader.service.d/override.conf`

```[Service]
TimeoutStopSec=5```

aarron-lee:
> Should be something like: `/etc/systemd/system/plugin_loader.service.d/override.conf`
> 
> ```
> [Service]
> TimeoutStopSec=5
> ```

makes sense, I'll close this PR for now and investigate opening another one later with the recommended override.

KyleGospo:
@aarron-lee Merged the fix I suggested w/ you tagged, thanks!

aarron-lee:
@KyleGospo saw that, thanks for adding the fix :+1:

Issue #1412, opened 2024-07-27
[Feature Request] Environment variable to prevent Steam from interfering with TDP


Tobe95:
On my Steam Deck LCD with a configured TDP of 20W, the power cap gets constantly reset back down to 15W by Steam itself, resulting in reduced performance and stutter. 

This happens mid-game, after setting a TDP above 15W with Powertools and confirming the change with `cat /sys/class/hwmon/hwmon4/power1_cap`.

My temporary solution has been to use `ostree admin unlock --hotfix` and edit `/usr/bin/steamos-polkit-helpers/steamos-priv-write`, preventing Steam from writing to `power*_cap` under `/sys/class/hwmon/hwmon4/` by commenting out this whole block: 

```
if MatchFilenamePattern "$WRITE_PATH" "/sys/class/hwmon/hwmon*/power*_cap"; then
      if /usr/libexec/hwsupport/valve-hardware; then
          CommitWrite
      else
          Log "commit: Skipped $WRITE_VALUE -> $WRITE_PATH - Valve Hardware not detected"
      fi
fi
```
Since this concerns any Steam Deck user running a higher TDP than the default 15W, it would be great to have some feature implemented to make this process easier, maybe in the form of an environment variable?

My current attempt looks like this:

```
if ! grep -q "^DISABLE_STEAM_TDP_CONTROL=1$" "/etc/environment"; then
  if MatchFilenamePattern "$WRITE_PATH" "/sys/class/hwmon/hwmon*/power*_cap"; then
      if /usr/libexec/hwsupport/valve-hardware; then
          CommitWrite
      else
          Log "commit: Skipped $WRITE_VALUE -> $WRITE_PATH - Valve Hardware not detected"
      fi
  fi
fi
```
While this only acts similar to an ENV, I could not get `if [ ${DISABLE_STEAM_TDP_CONTROL:=0} = "0" ]; then` to actually work...

Also I'm unsure if this would even be the best way to handle this, or if there might me a much simpler solution around?

KyleGospo:
Great idea, I will look into this soon and update this issue

Tobe95:
@KyleGospo Sorry to bother... But is there any update on this? 

I could make a pull request, if there is no better way to implement this.

KyleGospo:
> @KyleGospo Sorry to bother... But is there any update on this? 
> 
> I could make a pull request, if there is no better way to implement this.

If you're open to it, that would be great.

Tobe95:
[...]

PR #1535, opened 2024-08-23
ENV to disable Steam TDP-control on deck-hardware

Tobe95:
Optional environment variable to prevent Steam from controlling/interfering with TDP on Steam Deck. 

This is needed when having configured a higher TDP than the default 15W in the BIOS. Otherwise Steam "randomly" resets the power cap back to 15W, resulting in degraded performance. A bit more info about the issue [here](https://github.com/ublue-os/bazzite/issues/1412). 

I hope I've selected the right path for the config file location.


KyleGospo:
Looks good, only change I'd make is == 0 to != 1 so we're certain that's what the user meant.

KyleGospo:
Thanks!

Tobe95:
Done!

antheas:
We could also re-use the TDP machinery from hhd to make the TDP slider in steam show 20W? What's the maximum for that these days

KyleGospo (approved):

PR #1441, opened 2024-07-31
feat: Prepare mdBook workflow for documentation

Zeglius:
Add tooling (scripts, Github actions) in order to transcribe Discourse documentation to one using [mdBook](https://rust-lang.github.io/mdBook/) in efforts of improving accessibility when contributing to documentation.

Note: now with commit d42a29cb902035b76778ec183f4ef8904240c251, transcription is automated, staying in sync unidirectionally with discourse.

PR #1548, opened 2024-08-26
feat(docs): Replace mdbook with mkdocs workflow

Zeglius:
PR #1441 was meant to address: 
1. Lack of up-to-date offline documentation shipped with a Bazzite image.
2. Adding tooling to progressively moving documentation from Discourse to GitHub (if wished so).

Which was accomplished:
1. PDF generation out of documentation, which is then downloaded and added in the image build action (#1530).
2. Adding `fetch_discourse_md.py`, a script that facilitates scrapping of Discourse pages, both automatically by mdbook and contributors who wish to transcribe to a markdown file in GitHub.

It works, but would come to face several limitations imposed by mdbook:

1. MdBook ecosystem is written in Rust, which **add complexity to develop preprocessors** in simpler languages (ex.: python).
2. Related to limitation 1, **plugin ecosystem is far more scarce**, increasing the amount of self-developed solutions.
3. The current translation plugin ([mdbook-i18n-helpers](https://github.com/google/mdbook-i18n-helpers)) is far **too complex for translators**, as uses Gettext. Not a bad tool, but far too overkill for our purpose.
4. Part of limitation 3, increased workarounds to deal with the default build output file layout, which can be potentially prone to errors.

Thus, MkDocs enters in scene:
1. Written in Python, far more populated plugin ecosystem.
2. Plugin API is way friendlier with [hooks](https://www.mkdocs.org/user-guide/configuration/#hooks).
3. Less complex translation with [mkdocs-static-i18n](https://github.com/ultrabug/mkdocs-static-i18n), using markdown files like `article.es.md` instead of using Gettext.

Not only that, but we replace the PDF generation with a locally hosted version of the documentation, which can be served with `python -m http.server` (documentation shortcut handles this).
Far more friendly for Steam Game Mode users, as they can then access from Steam builtin web browser.


KyleGospo (approved):

nicknamenamenick (approved):
My approval doesn't affect whether or not this merges, but you got my seal of approval!!!!!

PR #1556, opened 2024-08-27
chore(docs): Delete Physical Keyboard Layout

nicknamenamenick:
It's now part of the Steam Gaming Mode documentation.  Yes I know this also affects Gamescope on the desktop too, but that information should be documented upstream anyways.


KyleGospo:
Need to remove references to this file as well

KyleGospo (approved):

PR #1585, opened 2024-08-31
fix(cec-control): compare variable rather than variable name

jsnyder:
Recently acquired a Pulse Eight USB CEC device and found bazzite wasn't switching on boot or other events it looks like it was configured to do switching on.  Found a fix for the script that should enable it to work again and not only turn the TV on, but also switch inputs.


KyleGospo:
Thanks!

KyleGospo (approved):

castrojo (approved):

HikariKnight (approved):

PR #1587, opened 2024-08-31
chore(docs): Rename user password doc

nicknamenamenick:
Since we now urge users to change their password if they install Bazzite without a physical keyboard, the name of this documentation is misleading.  This is strictly if the user **forgets** their password.


KyleGospo (approved):

HikariKnight (approved):

nicknamenamenick:
Redoing this GRRRRRRRRRR

PR #1593, opened 2024-09-01
chore(just): Improve install-resolve ux

Zeglius:
- Add a more informative message to install-resolve
- Bump up install-resolve SCRIPT_URL version


castrojo (approved):

castrojo:
We should think about publishing releases for the script so that dependabot can just update the version when zeglius does a release.

PR #1636, opened 2024-09-08
feat: add support for onexfly 8840U

HikariKnight:
However this sysid is shared between the 16gb model and the 32gb model ~~will this cause issues with the GTTSize being applied if this happens on a 16GB model? or will i have to separate out the 2 models in hardware setup by adding additional checks.~~


KyleGospo:
We can keep the 8GB VRAM for now and revisit with a mem check

KyleGospo (approved):

KyleGospo (approved):

noelmiller (approved):

PR #1637, opened 2024-09-08
chore(just): Improve wording in install-resolve prompts

Zeglius:


HikariKnight:
oh if italic is needed it might be worth PRing it into https://github.com/ublue-os/config/blob/main/build/ublue-os-just/lib-ujust/libformatting.sh for the future.

HikariKnight (approved):

KyleGospo (approved):

PR #1654, opened 2024-09-10
chore(docs): Remove theme override

nicknamenamenick:
I'll keep this as a draft for a while then I'll open it, and by the time it gets ack'd it won't be new anymore.


nicknamenamenick:
Zeglius is doing this in an upcoming [PR](https://github.com/ublue-os/bazzite/pull/1659)

Issue #1675, opened 2024-09-16
Laggy steam Big Picture Mode when switching in from desktop mode


Zeglius:
### Describe the bug

Quote from [discord](https://discord.com/channels/1072614816579063828/1284262108552368220/1285011847170232331) 

> I had something similar but in Nvidia. Running desktop UI and then switch to big picture mode would make the client really laggy.
> 
> But launching directly into big picture mode from the terminal (bazzite-steam -tenfoot) would make the lag disappear.
>
> _~Zeglius_

Only seems to happen when switching from desktop to Big Picture Mode.

Seems is caused by preloading `extest` in `bazzite-steam`, as by modifying it to not preload it, switching between desktop / BPM doesn't suffer from said lag. 

### What did you expect to happen?

To not be laggy

### Output of `rpm-ostree status`

```shell
State: idle
Deployments:
● ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-nvidia:stable
                   Digest: sha256:019b85f3ba44fcd0f7304eafe864ca9f4995da26fabadd70a0a5441de00fb695
                  Version: 40.20240908.0 (2024-09-09T05:03:59Z)
          LayeredPackages: code docker-ce-cli hyprland qt6ct swww

  ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-nvidia:stable
                   Digest: sha256:019b85f3ba44fcd0f7304eafe864ca9f4995da26fabadd70a0a5441de00fb695
                  Version: 40.20240908.0 (2024-09-09T05:03:59Z)
          LayeredPackages: code containerd.io docker-buildx-plugin docker-ce docker-ce-cli
                           docker-compose-plugin hyprland qt6ct swww
```

### Hardware

   zeglius@bazzite

 󱋩  bazzite-nvidia:stable 
 󰣛  Bazzite 40 (FROM Fedora Kinoite)
   Linux 6.9.12-205.fsync.fc40.x86_64
 󰅐  38 mins
 󰔠  Spawned on Feb 28 2024

 󰾰  HP Pavilion Notebook
 󰻠  Intel(R) Core(TM) i5-8250U (4) @ 3.40 GHz
 󰍛  NVIDIA GeForce GTX 1050 Mobile [Discrete]
 󰍛  Intel UHD Graphics 620 @ 1.10 GHz [Integrated]
   3.93 GiB / 7.63 GiB (52%)
   602.17 GiB / 888.76 GiB (68%) - btrfs [Read-only]
[...]

PR #1676, opened 2024-09-16
fix(steam): Do not preload extest in KDE plasma

Zeglius:
This would fix issue #1675 in KDE images

### DIFF
--- a/system_files/desktop/shared/usr/bin/bazzite-steam
+++ b/system_files/desktop/shared/usr/bin/bazzite-steam
@@ -51,7 +51,8 @@ if [[ $(echo "${switcheroo_state}" | grep -o 'Device:' | wc -l) -gt 1 ]]; then
 fi
 unset -v switcheroo_state
 
-if [[ "$XDG_SESSION_TYPE" == "wayland" ]]; then
+# Do not preload extest if we are using KDE with libei
+if [ "$XDG_SESSION_TYPE" = "wayland" ] && [ "$XDG_CURRENT_DESKTOP" != "KDE" ]; then
   # https://github.com/Supreeeme/extest
   # Extest is a drop in replacement for the X11 XTEST extension.
   # It creates a virtual device with the uinput kernel module.


KyleGospo (dismissed):

HikariKnight:
Wont this break steam input on KDE? like the Desktop Layout and such.

Zeglius:
> Wont this break steam input on KDE? like the Desktop Layout and such.

Didn't think of that. Thanks for pointing  that out. 
Probably this should do better in the testing branch.

HikariKnight:
im more comfortable with this in testing to test it properly first, will do some testing after work tomorrow to check steam input as i rely on it a lot on desktop

HikariKnight (approved):

HikariKnight:
@Zeglius i have tested this today, seems to not impact steam input just gives a dialog asking for remote input access and once granted seem to not ask again (will check again after reboot) but so far seems good.

Zeglius:
> ok so after every reboot (or relogin) the user will have to re-accept the remote control prompt when trying to use steam input with steam chords (like opening the keyboard), this i can imagine is fine on the desktop image not on the `-deck` image
> image
> 
> Also when i tested it did not fix the big picture mode lag on my end :thinking: 
> 

How did you test this? This PR seems didn't get merged in `:testing`.

Zeglius:
Ok so I have been testing a little bit and seems extest is not the cause of the lag.

I could reproduce the lag by launching bazzite-steam in the terminal, **but not when launching from the app menu**

I checked the difference between both, from terminal and from app menu, and it seems it was switcheroo all along, which doesnt inject the required env vars when in terminal.

As soon I launched bazzite-steam with `switcheroo launch bazzite-steam`, the lag was gone

HikariKnight:
> > ok so after every reboot (or relogin) the user will have to re-accept the remote control prompt when trying to use steam input with steam chords (like opening the keyboard), this i can imagine is fine on the desktop image not on the `-deck` image
> > image
> > Also when i tested it did not fix the big picture mode lag on my end 🤔
> 
> How did you test this? This PR seems didn't get merged in `:testing`.

just downloaded the script and tested with that :D

PR #1695, opened 2024-09-22
fix: add potentially missing selinux labels back to nvidia-modeset

HikariKnight:
Apparently this stops sddm from crashing a lot when logging in with a VRR monitor conntected (gdm untested as everyone affected were on kde)

PRing into testing since i have not gotten proper verification that this fully fixes the issue.
can be tracked here: https://www.answeroverflow.com/m/1285186896137425016


KyleGospo (approved):

KyleGospo:
Open to it, I look forward to the day this can be removed

KyleGospo:
Believe this was confirmed to not work? Let me know if I'm mistaken

Issue #1720, opened 2024-10-01
[Game Mode] Steam Overlay hotkey collision with games


AkazaRenn:
### Describe the bug

Hitting `ctrl + 1` and `ctrl + 2` in game opens the the Steam Overlay, while there are some other functions to be done by the game. 

Use the classic `Counter-Strike` for example, `ctrl` is for crouching, `1` and `2` switch to the major and alternative weapons respectively. They are supposed to be working in combinations, but now they will bring up Steam Overlay and QAM instead.

**Steam OS doesn't have this issue.**
**Bazzite OS Deck image on Desktop Mode doesn't have this issue.**

### What did you expect to happen?

Same as Steam OS, disable the aforementioned hotkeys while in game

parkerlreed:
This is enabled for the gamescope session as a convenience for keyboard users to get back to Steam/exit game.

The reason it only happens here as it's explicitly patched in to either the Steam client or gamescope I don't recall.

There should be a way to reverse the patch.

parkerlreed:
Looks like you'd need to rebuild gamescope with this patch reverted. https://github.com/ublue-os/bazzite/blob/main/spec_files/gamescope/v2-0001-always-send-ctrl-1-2-to-steam-s-wayland-session.patch

Likely either to just use desktop mode for playing those specific games.

AkazaRenn:
> This is enabled for the gamescope session as a convenience for keyboard users to get back to Steam/exit game.
> 
> The reason it only happens here is it's explicitly patched in to either the Steam client or gamescope I don't recall.
> 
> There should be a way to reverse the patch.

Is there a way to get `shift + tab` to work instead? `ctrl + 1` is too disruptive... I wish I don't have to keep switching between modes.

AkazaRenn:
### Use `shift + tab` instead
```diff
diff --git a/src/wlserver.cpp b/src/wlserver.cpp
index 9c82a76..1179d57 100644
--- a/src/wlserver.cpp
+++ b/src/wlserver.cpp
@@ -312,8 +312,7 @@ static void wlserver_handle_key(struct wl_listener *listener, void *data)
 		keysym == XKB_KEY_XF86PowerOff;
 
 	// Check for steam keys (ctrl + 1/2)
-	bool is_steamshortcut = (keyboard->wlr->modifiers.depressed & WLR_MODIFIER_CTRL) && (keysym == XKB_KEY_1 ||
-																						 keysym == XKB_KEY_2);
[...]

PR #1724, opened 2024-10-04
fix(deck): Use super + 1/2 for Overlay/QAM

AkazaRenn:
For https://github.com/ublue-os/bazzite/issues/1720. The reason I'm not using shift + tab is that it just doesn't work... I think Steam in deck mode ignores that key combination. This should still be a safe combination since that's a Windows reserved hotkey and no game would consider using it.

I've tested this change to be working as expected by building the binary and replacing it at /usr/bin/gamescope


antheas:
We cannot merge this yet as certain functionality relies on it. However, this patch looks really interesting.

AkazaRenn:
> We cannot merge this yet as certain functionality relies on it. However, this patch looks really interesting.

Is there anything I can help? It would be nice to get this merged so we no longer block ctrl-hotkeys.

antheas:
Modify this patch to include two env vars: one for disabling ctrl + 1/2, and one for enabling win + 1/2.

Then this patch can merge without breaking other stuff, and we can slowly work on transitioning to win + 1/2.

Specifically, hhd uses ctrl+1/2 on tablets so it would need to coordinate with the win env var. Also, a lot of users have started depending on ctrl+1/2.

antheas:
Probably both ctrl/win+1/2 will co-exist for a few months. Specifically, ctrl + 1/2 causes issues in certain games. You are right.

AkazaRenn:
> Probably both ctrl/win+1/2 will co-exist for a few months. Specifically, ctrl + 1/2 causes issues in certain games. You are right. This is even if the user does not want to use it in-game.

What you mean is that we add an environment variable check, if that's true, we still passthrough ctrl + 1/2?

antheas:
The ctrl + 1/2 needs to be opt out (i.e., var disables it). win + 1/2 needs to be opt in (i.e. var enables it).

Otherwise this patch will take a lot longer to merge.

AkazaRenn:
> The ctrl + 1/2 needs to be opt out (i.e., var disables it). win + 1/2 needs to be opt in (i.e. var enables it).
> 
> Otherwise this patch will take a lot longer to merge.

I can help make this change, but I don't know about personal configurations on Bazzite. Where do we do this opt-in/out?

antheas:
For these things, you can add a variable to `/etc/environment.d` and it will apply to gamescope. This way you could use win + 1/2 in your personal machine.

antheas:
Use the variables `GS_CTRL12_DISABLE` and `GS_WIN12_ENABLE` and reference other gamescope patches to see how to fit them in.

AkazaRenn:
> Use the variables `GS_CTRL12_DISABLE` and `GS_WIN12_ENABLE` and reference other gamescope patches to see how to fit them in. 

Thanks, I'll start working on it.

AkazaRenn:
f1fa969f16181b712c375de98d82cf875ed26464
`GS_ENABLE_CTRL_12` in `/etc/environment` commented out, not existing == true for update compatibility. I didn't add another one for `super` because that's not causing any troubles and feel like it's unnecessary.

KyleGospo (approved):

HikariKnight (approved):

AkazaRenn:
Hi @antheas, this change has been added to the latest `testing` branch of Bazzite and I've tested it to be fully functional. You may consider adjusting the HHD codes to rely on `super + 1/2` instead so we can drop the option in the future. Thanks a lot for your help!

PR #1726, opened 2024-10-05
feat: Add HPET configuration from Atomic Studio

EPOCHvoyager:
_I have to preface this by stating that I do not understand the underlying mechanisms adequately myself_ - but after speaking to the Atomic Studio dev last month, I'm told this tweak might help those facing audio issues, particularly on constrained handhelds - and that it should have no ill effect otherwise.

Atomic Studio currently implements it [here](https://github.com/atomic-studio-org/Atomic-Studio/blob/5a215a9a8442c7753d45c6f00d4529d2bc80d55d/files/files/shared/etc/sysctl.d/50-atomicstudio.conf).

### DIFF
--- a/system_files/deck/shared/usr/lib/sysctl.d/65-hpet.conf
+++ b/system_files/deck/shared/usr/lib/sysctl.d/65-hpet.conf
@@ -1,0 +1,1 @@
+dev.hpet.max-user-freq=3072
--- a/system_files/deck/shared/usr/lib/sysctl.d/65-hpet.conf
+++ b/system_files/deck/shared/usr/lib/sysctl.d/65-hpet.conf
@@ -1,1 +1,0 @@
-dev.hpet.max-user-freq=3072
--- a/system_files/desktop/shared/usr/lib/sysctl.d/65-hpet.conf
+++ b/system_files/desktop/shared/usr/lib/sysctl.d/65-hpet.conf
@@ -1,0 +1,1 @@
+dev.hpet.max-user-freq=3072


EPOCHvoyager:
Do note that this is my first technical PR, and that as such this might not be the proper or preferable way of implementing this.

HikariKnight:
I do not get it myself, however you do not need to add the file to both desktop and deck.
the deck image is built from the desktop image so adding it there is enough, if the deck image needs something different it is then overridden with a file from the deck folder instead :)

HikariKnight:
@tulilirockz you might be able to give more insight in this for us as to how this works and help.

system_files/deck/shared/usr/lib/sysctl.d/65-hpet.conf
```diff
@@ -0,0 +1 @@
+dev.hpet.max-user-freq=3072
```
HikariKnight, line 1: This is not needed when the file exists in the desktop image, as the deck image is built from the desktop image.
If we needed the number to be different for the deck image then this file would be needed.

RayBa82 (approved):

RayBa82:
Cool I could approve that, shouldn't I be a maintainer or something? :-D
But fixing audio issues is always nice :-)

HikariKnight:
After having talked to @tulilirockz about the details of this, i don't think i will be approving this as it would increase power consumption (and the desktop image is also used on laptops) and battery drain for little to no gain for gaming audio.

EPOCHvoyager:
Alright; with that in mind, I think it might be even better to implement as part of a  `ujust setup-realtime` command - which could automatically layer the `setup-realtime` package and start the appropriate `systemd` services, with this little tweak as a bonus. Though with RT being merged for kernel 6.12, it could be better to wait out a bit.

HikariKnight:
> Alright; with that in mind, I think it might be even better to implement as part of a `ujust setup-realtime` command - which could automatically layer the `setup-realtime` package and start the appropriate `systemd` services, with this little tweak as a bonus. Though with RT being merged for kernel 6.12, it could be better to wait out a bit.

Does setup-realtime add any kernel modules? if so it cant be done on the users end.

EPOCHvoyager:
> Does setup-realtime add any kernel modules?

No - but the `realtime-setup` package adds a service(of two, the other is a workaround for network interrupts) that specifically supports a RHEL-RT kernel, which might become regular upstream after 6.12. It's mainly just an `/etc/limits.d/` configuration file and some user group management under the hood, as far as I can tell. 

> also i think setup-realtime-audio would be more descriptive even though it would have a description.

It would be the main use most likely, but there are others. I think mostly sharing the name with the package could leave less guesswork to the user?

HikariKnight:
> > Does setup-realtime add any kernel modules?
> 
> No - but the `realtime-setup` package adds a service(of two, the other is a workaround for network interrupts) that specifically supports a RHEL-RT kernel, which might become regular upstream after 6.12. It's mainly just an `/etc/limits.d/` configuration file and some user group management under the hood, as far as I can tell.
> 
> > also i think setup-realtime-audio would be more descriptive even though it would have a description.
> 
> It would be the main use most likely, but there are others. I think mostly sharing the name with the package could leave less guesswork to the user?

if it does more than audio then `setup-realtime` works better
i would say maybe wait with this until bazzite is on the 6.12 kernel?
can still prep the PR but add the note to not merge until bazzite uses the 6.12 kernel

EPOCHvoyager:
I'll convert this PR to a draft for now.

PR #1736, opened 2024-10-11
feat(gamescope): prepare to start tracking upstream + cleanup

antheas:
Spring cleaning (in the fall) for gamescope. Remove all extraneous patches and prepare for dropping the remaining ones as the lua config system solidifies.

Previous launch arguments were preserved. The external display arg was modified to only lie to Steam instead of messing with gamescope and --force-external-orientation was modified to act as --force-orientation to avoid early session changes.


antheas:
Unfortunately, external patch did not cooperate with steam breaking the fps slider. So it had to go as well. Oh well.

antheas:
Added the ability to enable the refresh rate slider while VRR is on for the Ally with the flag `--enable-vrr-modesetting`.

KyleGospo (approved):

KyleGospo (approved):

PR #1742, opened 2024-10-14
chore: gate stable and add promotion action

antheas:
The autoupdater pushed a partial update today, and the previous update broke the framework kmod (because it is now "upstream")

The update regression rate is too high for the current state of bazzite, for the vast majority of its users.

That being said, there are a lot of users that like to be a bit ahead.

Therefore, gate the stable builds behind an `edge` tag for power users, and add a small action to promote the current `edge` build to `stable` once it has been validated to work.

Just in time for F41. 

The regression in this update was small: the Ally frame limiter stopped working and we were caught by surprise. It would have been worse, as with F41. The edge channel also allows us to update that channel to F41 earlier than Fedora.

### DIFF
--- a/.github/workflows/promote.yml
+++ b/.github/workflows/promote.yml
@@ -1,0 +1,62 @@
+name: Promote Edge builds to Stable
+
+on:
+  workflow_dispatch:
+  workflow_call:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}-promote
+  cancel-in-progress: true
+
+jobs:
+  promote:
+    runs-on: ubuntu-24.04
+    permissions:
+      packages: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        image_name:
+          - bazzite
+          - bazzite-nvidia
+          - bazzite-nvidia-open
+          - bazzite-gnome
+          - bazzite-gnome-nvidia
+          - bazzite-gnome-nvidia-open
+          - bazzite-deck
+          - bazzite-deck-gnome
+          - bazzite-ally
+          - bazzite-ally-gnome
+          - bazzite-asus
+          - bazzite-gnome-asus
+          - bazzite-asus-nvidia
+          - bazzite-gnome-asus-nvidia
+          - bazzite-asus-nvidia-open
+          - bazzite-gnome-asus-nvidia-open
+        major_version: [40]
+    steps:
+      - name: Login to GHCR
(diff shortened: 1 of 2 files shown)


.github/workflows/promote.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Lowercase Registry
+        id: registry_case
+        uses: ASzc/change-string-case-action@v6
+        with:
+          string: ${{ env.IMAGE_REGISTRY }}
+
+      - name: Pull Image (Edge tag)
+        uses: Wandalen/wretry.action@v3.5.0
+        id: pull
+        with:
```
p5, line 48: Do we need to pull here?  Skopeo can add tags to images in a remote registry without pulling.

antheas: Give me the command and I will replace it. Am not aware of it.

p5: I'm on mobile, so can't make something that's drop-in, but this is how I've done it in the past

```
            skopeo copy docker://${IMAGE_REGISTRY}@${IMAGE_DIGEST} docker://${IMAGE_REGISTRY}:latest
```

The digest can be replaced with an existing tag

HikariKnight (approved):
i think this is a good idea and will give some users piece of mind.
alternative (in my head) is to have testing be the one that builds daily and unstable be where we play with new stuff.

KyleGospo (approved):

PVermeer:
With this setup the latest edge build is released (on Github) as stable https://github.com/ublue-os/bazzite/releases/tag/41.20241112 but the image has no stable tag yet.

I would also not recommend this setup as it will be confusing as to what the main branch is. There is no separation between `edge` and `main` this way. The setup that Bazzite had was in my opinion enough. I would recommend to use it as a merge tree though:

`testing` -> `unstable` -> `stable`

This goes only one way with the exception of hotfixes on `stable` which should also be merge in `unstable` and `testing`. So no pushes on main except hotfixes and merge commits. Edge is now just unstable with another name that adds complexity and confusion.

PVermeer:
Thanks for the reply.

> This got merged a tad bit too early. It should be called rolling.

I agree.

> The problem it's designed to solve is that once an image is built it can't be rebuilt again

I understand.

> Just because a ref got built in testing it won't be the same when it builds in stable due to dependency drift. So stable builds need to be tested before they land in the stable tag

I understand. Merging between branches doesn't necessarily mean building a new image. It could also be a promotion. I simply just prefer different git branches for `rolling` -> `stable` but git tags also works.

Thanks anyways!

PR #1766, opened 2024-10-23
chore: remove sudo from bazzite setup script

antheas:
bazzite setup seems to run with root priviledges as it writes to /etc directly. However some commands use sudo.

For some reason, using sudo delays boot by over 2 minutes in certain instances. So remove it.

Also remove pmf script from Ally since it is part of 6.11.


tulilirockz:
The delayed boot probably happens due to networking/hostname not yet being configured and sudo getting confused then timing out. Had toooooooooons of these problems when I was testing out systemd-nspawn

PR #1767, opened 2024-10-23
feat: add automatically generated changelogs

antheas:
Script is where it should, but action still needs work


antheas:
Added pretty version generation if we dont have it for the title. Should look the same as build id

PR #1771, opened 2024-10-24
fix: temporarily disable "turn off screen when idle" in kde for handhelds

HikariKnight:
Temp workaround until we can figure out what causes it to not come back upon input immediately on handhelds. I was unable to reproduce this on a laptop(desktop image as laptop is too old for deck image) and a desktop (deck image) but i can reproduce it on the Legion Go.


HikariKnight:
I will be trying to look into this issue more when i have the energy for it as i am still on the mend and my energy level is out of whack.
So i wrote this mainly as a temp workaround so people do not get stuck for now.

PR #1789, opened 2024-10-27
feat(ujust): Update pwfeedback ujust to include optional arg

aarron-lee:
add optional arg for to explicitly turn pwfeedback on/off

```
ujust toggle-password-feedback on
ujust toggle-password-feedback off
```

useful for programmatically turning on/off pwfeedback

e.g. in the Bazzite portal


HikariKnight:
would suggest a "help" action too so it matches the other just recipes with actions, but its not a blocker

HikariKnight (approved):

KyleGospo (approved):

PR #1804, opened 2024-10-29
fix(ujust): Replace missing f41 lact package for 40 one

Zeglius:
Drop this as soon as possible


KyleGospo:
-1 on this, let the LACT team fix it. It affects Fedora as of today as well.

PR #1838, opened 2024-10-31
feat(just): Update surround to use spatializer

EPOCHvoyager:
Updates the `ujust` to setup virtual surround to use the new PipeWire Spatializer module in place of the older HeSuVi-style convolver.

Not only is this method now recommended in the PipeWire Wiki, but free `.sofa` HRTF files are much more plentiful online; which is particularly relevant given head-related transfer functions must pertain to the individual's anatomy for proper effect. 

The sample HRTF is downloaded from the Sofacoustics database in the script, with the original measurements from MIT being freely available [here](https://sound.media.mit.edu/resources/KEMAR.html), where they're allowed free redistribution.

Additionally, I've taken the liberty to add small step-by-step instructions at the end of the script's execution, including links, as to how to identify and obtain a `.sofa` which will better match the user, and provide fuller effect.

### DIFF
--- a/system_files/desktop/shared/usr/share/ublue-os/just/83-bazzite-audio.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/83-bazzite-audio.just
@@ -94,12 +94,12 @@ setup-virtual-channels ACTION="":
       echo "Virtual audio channels config removed, the channels will be removed next time you login."
     fi
 
-# Setup a simple Virtual Surround 7.1 sink using the ASH Control Room 1 convolver file (you can change this yourself after setup)
+# Setup a virtual 7.1 surround sink for headphones using the PipeWire Spatializer module
 setup-virtual-surround ACTION="":
     #!/usr/bin/bash
     source /usr/lib/ujust/ujust.sh
     mkdir -p ~/.config/pipewire/pipewire.conf.d
-    mkdir -p ~/.config/pipewire/hrir_hesuvi
+    mkdir -p ~/.config/pipewire/hrtf-sofa
     OPTION={{ ACTION }}
     if [ "$OPTION" == "help" ]; then
       echo "Usage: ujust setup-virtual-surround <option>"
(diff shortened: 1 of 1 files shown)


HikariKnight:
i would suggest deleting the old config (`~/.config/pipewire/pipewire.conf.d/virtual-surround-71.conf`) if found to avoid potential issues or at least moving it out of the pipewire.conf.d and with .bak extension
otherwise this looks good.
will test and approve after work tomorrow once that change is done.

EPOCHvoyager:
> i would suggest deleting the old config (~/.config/pipewire/pipewire.conf.d/virtual-surround-71.conf) if found to avoid potential issues or at least moving it out of the pipewire.conf.d and with .bak extension

Done - backs the file up on enabling, and removes it instead on disable.

EPOCHvoyager:
Something of note otherwise is that while the configuration is based on [this one](https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/daemon/filter-chain/spatializer-7.1.conf) from the PipeWire repo, I've changed the output channels to Side Left and Right, instead of Front Left and Front Right - as that's consistent with how EasyEffects handles things and doesn't make it so the sound is always coming from in front.

HikariKnight:
hmm i ran the just recipe and restarted pipewire, but i did not get any new sink available to me to select :thinking:

EPOCHvoyager:
> hmm i ran the just recipe and restarted pipewire, but i did not get any new sink available to me to select 🤔

Weird, works perfectly here; though you have to be on a recent image for the added sofa module. Anything on logs?

HikariKnight:
> > hmm i ran the just recipe and restarted pipewire, but i did not get any new sink available to me to select 🤔
> 
> Weird, works perfectly here; though you have to be on a recent image for the added sofa module. Anything on logs?

might be why, my system is still on F40 due to some bugs in a few packages i rely on for my main system. so i dont have any other ideal hardware to test with.

HikariKnight (approved):
Tested by Vet Gaming and confirmed working

AlexNPavel:
I'm having issues with the new spatializer. I removed the old convolver and tried this one out and there is no bass. It's as if there is no subwoofer registered. The subwoofer does appear in the sound settings though and it tests correctly as "rear center" audio. I reset the pipewire config and went back to the old convolver to make sure it wasn't something else that broke and it works correctly there, so I think something is wrong with this new spatializer based config.

EPOCHvoyager:
@AlexNPavel, this version should sound different from the previous HRIR-based one given it's using an HRTF based on completely different measurements(literally measurements of a dummy head at MIT); it's only meant as a placeholder file that we can distribute. Did you follow the instructions given at the end of the new installation process?

AlexNPavel:
@EPOCHvoyager I understand that it should be different, but it is a pretty bad initial experience. After doing some more research, it seems most HRTF models (like MIT KEMAR and the ARI models) don't have any testing for frequencies below 300Hz since they're designed for research, not virtual surround. This means that there's essentially no subwoofer with those models. Even the ARI B models only go down to 50Hz and lack depth as a result. IMO, one of the "Widespread" models as a default would make more sense, since those are generated using computer simulations, allowing them to get info for lower frequencies that isn't possible with physical in-ear microphones.

There definitely should still be the message to experiment with other models for users to find what they like best, but the default should also provide a good experience IMO.

EPOCHvoyager:
> one of the "Widespread" models as a default would make more sense

The reason I picked KEMAR was because the measurements explicitly allowed for free redistribution, outside of any other practical reasons. Wouldn't be wise to knowingly distribute something in license limbo on an open source project like Bazzite. If you can find another with such a clause that I missed, I'd be happy to hear it.

> There definitely should still be the message to experiment with other models for users to find what they like best, but the default should also provide a good experience

This is orthogonal to the actual issue, but with HRTFs, it's far less of an emotional preference over how your anatomy interacts with your present soundscape. You can do some more generalized models like the commercial options do, but it's always going to be a trade-off between basic utility and general naturality. Even still, there will be a sizeable portion of the populace to which they'll provide neither, and largely couldn't provide without alienating others. Although I have to outline the caveat that I'm not anything resembling a researcher on the field.

With that in mind, and the limitations outlined to your previous point, I still believe it's best to try and guide people on how to get something that better matches their anatomy, than to make an effort to provide one that's "good enough" right away - as whatever that means would be particularly unclear.

PR #1840, opened 2024-11-01
feat: Add atom for controlling refresh rate halving in Steam

antheas:
Adds atom for controlling whether we halve the hz in SteamUI. See https://github.com/hhd-dev/hhd/commit/ea00e99a446f0face75f82b95bd6d801ac7f7efc


antheas:
also cleans up `git format-patch` to make diffs cleaner in the future

PR #1848, opened 2024-11-03
feat(ujust): Add setup-boot-windows-steam ujust script

wolfyreload:
I'm seeing a number of people using my https://youtu.be/WR31wkrxock video for making a script to boot into Windows from Steam in GameMode. However, quite a few people miss a step or type a character incorrectly. Thought that it might be possible to automate this with a ujust script.

Also want to give thanks to u/0nlin3 on Reddit for the idea.

### DIFF
--- a/system_files/desktop/shared/usr/share/ublue-os/just/86-bazzite-windows.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/86-bazzite-windows.just
@@ -1,0 +1,8 @@
+# vim: set ft=make :
+
+# Adds a script in Steam to boot Windows which is useful for dual-boot setups
+setup-boot-windows-steam:
+    echo "Making efibootmgr -n usable without sudo password"
+    echo "%wheel ALL=(root) NOPASSWD: /usr/sbin/efibootmgr" | sudo tee /etc/sudoers.d/efibootmgr-config
+    echo "Adding /usr/bin/boot-windows as a Non-steam game"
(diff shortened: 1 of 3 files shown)


Zeglius:
So, for what I'm seeing,  the script creates another script on the fly as it gets executed.

Would prefer if we added an already written one, rather than creating it on the fly, otherwise is going to be more difficult to debug/maintain

wolfyreload:
@Zeglius Didn't think of that, that would be a great improvement. Will make a separate script and just call that.

system_files/desktop/shared/usr/share/ublue-os/just/86-bazzite-windows.just
```diff
@@ -0,0 +1,9 @@
+# vim: set ft=make :
+
+# Adds a script in Steam to boot Windows which is useful for dual-boot setups
+setup-boot-windows-steam:
+    echo "Making efibootmgr -n accessible without sudo"
+    echo "%wheel ALL=(root) NOPASSWD: /usr/sbin/efibootmgr" | sudo tee /etc/sudoers.d/efibootmgr-config
+
```
HikariKnight, line 7: just syntax checking will be angry about the empty line IIRC

wolfyreload: I was not aware of that. I have removed the empty line from the script

KyleGospo (approved):

HikariKnight (approved):

PR #1905, opened 2024-11-21
feat(ci): Replace rpm-ostree with dnf5 during build

KyleGospo:


HikariKnight (approved):
so much cleaner!

Containerfile
```diff
@@ -37,125 +37,79 @@ COPY system_files/desktop/shared system_files/desktop/${BASE_IMAGE_NAME} /
 
 # Update packages that commonly cause build issues
 RUN --mount=type=cache,dst=/var/cache/rpm-ostree \
```
LukeShortCloud, line 39: We've been using DNF 5 in PlaytronOS for a few months now and, yes, it's amazing!

One of the great things is that it takes care of dependencies. It's okay if a package is slightly out-of-date, DNF will automatically deal with that by updating the package(s) as necessary. rpm-ostree had a hard time because it didn't know how to update base packages. You should be able to remove this entire "Update packages that commonly cause build issues" block.

LukeShortCloud (commented):
I haven't personally encountered any of these problems in PlaytronOS, but Bluefin has run into some odd issues with DNF 5 over in their (new) related PR: https://github.com/ublue-os/bluefin/pull/1954

PR #1913, opened 2024-11-22
fix(gamescope): remove dpms early path which may be causing a crash

antheas:
GPD Win Mini experiences a display crash after sleep when using the DPMS function. Gamescope stops showing stuff on the screen but you can still hear steam in the background. Remove the early present which might be causing this.

https://github.com/hhd-dev/hhd/issues/115


HikariKnight (approved):

antheas:
Made resuming from DPMS a bit faster as well.

PR #1937, opened 2024-12-01
fix(btrfs-assistant) Enable snapper-cleanup.timer

Zeglius:
This will avoid snapper creating snapshots without removing old one, eventually filling up the space

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -769,6 +769,7 @@ RUN rm -f /etc/profile.d/toolbox.sh && \
     systemctl disable tailscaled.service && \
     systemctl enable dev-hugepages1G.mount && \
     systemctl enable bazzite-snapper-setup && \
+    systemctl enable snapper-cleanup.timer && \
     systemctl --global enable bazzite-user-setup.service && \
     systemctl --global enable podman.socket && \
     systemctl --global enable systemd-tmpfiles-setup.service && \


Containerfile
```diff
@@ -769,6 +769,7 @@ RUN rm -f /etc/profile.d/toolbox.sh && \
     systemctl disable tailscaled.service && \
     systemctl enable dev-hugepages1G.mount && \
     systemctl enable bazzite-snapper-setup && \
```
antheas, line 771: what why is bazzite-snapper-setup enabled? That should not enable snapper right? @HikariKnight

antheas:
Exacerbates the problem. Snapper should not be enabled by default, trying to fix the problem of it being enabled is not a solution

Containerfile
HikariKnight: nope it should not enable any of the automatic snapshotting, it didnt on my testing

antheas: Lets remove it anyway, it is causing issues

You can convert it to a ujust instead that also enables the service

Zeglius: No need for ujust, the cleanup thing can be enabled from btrfs-assistant gui

antheas:
Ok, reopen, this service is needed now

Thank you Hikari for the oopsie

According to the arch wiki, snapper runs on cron by default
https://wiki.archlinux.org/title/Snapper

Here is my ally
image

So, lets suggest mitigation options now

antheas:
On the next update, the config created by this service needs to be purged by the update script and this service needs to be removed from Bazzite

antheas:
/home snapshots are completely useless. Youre more likely to delete your files while managing them than they are to help. They should have never been enabled by default.

Zeglius:
Ok nvm now I realized @antheas the guy said he didn't have cleanup enabled, not  the automatic snapshotting itself

antheas:
I never enabled snapper, yet i have snapshots

antheas:
Mitigation needs to remove config, all services, and all snapshots of the /home directory, only if they were created by the script, and result in no data loss.

This is a wishlist so other than no data loss, the others are nice to have

PR #1938, opened 2024-12-01
fix(btrfs-assistant): Prevent snapper from saturating storage with snapshots

Zeglius:
This PR disables snapper-timeline.timer whenever snapper config is setup.
It also disables snapshots on any currently affected devices. Users might need to reenable them afterwards.


Zeglius:
Closed in favor of https://github.com/ublue-os/bazzite/pull/1939

PR #1939, opened 2024-12-01
fix: remove default snapper config and incorrect snapshots

antheas:
Something lead to snapper turning on on its own and making snapshots forever. Revert the niceties that can still be added via ujust (and were broken) and remove the incorrect snapshots that might have been created.


HikariKnight:
I still think fixing the config using `snapper set-config key=value` and then running the snapper-cleanup service to cleanup the invalid snapshots would be enough.
we already confirmed the service is somehow enabled by itself but is off by default.
we can just update the bazzite-snapper-setup to force set the config values after making the config

antheas:
It would create too much drift with previous images

This PR fixes the problem cleanly and removes previous drift

More than 95% of users do not need snapper. And for the other 5% messing with their config is not acceptable

HikariKnight:
we are talking about a config where the timers somehow got enabled and 10 years worth of snapshots are being kept, thats not really a sane snapper setup in general unless we are talking about some serious data hoarding in /var/home

HikariKnight:
> It only deletes timeline snapshots

`snapper cleanup timeline && snapper cleanup number`
this will cleanup all automatic snapshots that fall outside of the numbers range and the timeline ranges in the config (which we can set to 0 on deck images)

manual ones are kept

Zeglius:
Alright so only timeline snapshots. Fair. 

Though this would be better off by not doing it in the hardware setup script, that doesn't make any favors.
Maybe in an oneshot Systemd service that disables itself, not before executing the cleanup script, and maybe only on deck images(?)

antheas:
> Alright so only timeline snapshots. Fair.
> 
> Though this would be better off by not doing it in the hardware setup script, that doesn't make any favors. Maybe in an oneshot Systemd service that disables itself, not before executing the cleanup script, and maybe only on deck images(?)

Thats what the hardware setup script is for. Random oupsies. We cannot add new services every time there is a mistake.

Zeglius:
Fair. There is also the part of nuking the configuration. Maybe instead of completely removing it, setting more sane defaults would alive the issue instead of leaving btrfs-assistant unusable

antheas:
> > It only deletes timeline snapshots
> 
> `snapper cleanup timeline && snapper cleanup number` this will cleanup all automatic snapshots that fall outside of the numbers range and the timeline ranges in the config (which we can set to 0 on deck images)
> 
> manual ones are kept

Would break users that want to actually use snapper. The only clean condition for cleaning snapshots is if the invalid config exists, because no sane user would set it up that way. After that, it should never run again or mess with the config

antheas:
> Fair. There is also the part of nuking the configuration. Maybe instead of completely removing it, setting more sane defaults would alive the issue instead of leaving btrfs-assistant unusable

Opening btrfs-assistant allows you to make a config with one click and enable the services as well.

HikariKnight:
> Opening btrfs-assistant allows you to make a config with one click and enable the services as well.

that config will be for `/` since it will not let you make one for `/var/home` without one for `/` in the gui last time i tested IIRC

Zeglius:
> > Fair. There is also the part of nuking the configuration. Maybe instead of completely removing it, setting more sane defaults would alive the issue instead of leaving btrfs-assistant unusable
> 
> Opening btrfs-assistant allows you to make a config with one click and enable the services as well.

Alright fair. Bit nuclear of an option. Not that mine is any better by reducing the timeline duration to like 10 days or so(?), so not against it.

Pls add a `FIXME:` or whatever so we can remember to remove this after these builds with the old config disappear 🙏

antheas:
> > Opening btrfs-assistant allows you to make a config with one click and enable the services as well.
> 
> that config will be for `/` since it will not let you make one for `/var/home` without one for `/` in the gui last time i tested IIRC

Seems to allow me to create /var/homw with no config

HikariKnight:
> > > Opening btrfs-assistant allows you to make a config with one click and enable the services as well.
> > 
> > 
> > that config will be for `/` since it will not let you make one for `/var/home` without one for `/` in the gui last time i tested IIRC
> 
> Seems to allow me to create /var/homw with no config

well im fine with this and then later add a proper config since we found out the config in the repo was not used.
instead later i will just make it create a config and use the snapper command to set the values to 0 for deck and use the default values of the original PR for desktop instead of changing the default template since that clearly is not respected on the end image.

HikariKnight (approved):

KyleGospo (approved):

PR #1941, opened 2024-12-02
fix(snapper): fix invalid config and remove incorrect snapshots

HikariKnight:
for some users the `snapper-timeline.timer` has been enabled out of the blue, it is not enabled by default (it is off on a new install and on my legion go)

Also somehow the config we made in the repository is not even added to the final image, so everyone has a default config where the timeline retention is 10 years of snapshots, this did not happen during the months of testing i did.

This will correct the config by setting all snapshot retention values to `0` so the user **HAS** to make a changes themselves for snapper to work (this was the idea from the beginning but somehow the snapper-timeline timer gets enabled and we do not know how)
and adds a safeguard against it and deletes all incorrectly taken snapshots
The same safeguard is applied to `snapper-boot.timer`

This PR is in conflict with #1939 and it way less nuclear and achieves the same thing.
I do not mind which PR is chosen, main point is we have 2 options to fix the issue.


Zeglius:
You still need some way to start the script to apply the new config on already affected devices, maybe an oneshot service that disables itself afterwards or calling it at bazzite-hardware script

HikariKnight:
> You still need some way to start the script to apply the new config on already affected devices, maybe an oneshot service that disables itself afterwards or calling it at bazzite-hardware script

bazzite-snapper-setup runs at startup

PR #1948, opened 2024-12-05
fix(tuned-ppd): correct config existence check in bazzite-hardware-setup

ledif:
While looking through the system journal, I noticed one of the Bazzite scripts was throwing an error on boot:

```
Dec 04 18:46:39 bazzie bazzite-hardware-setup[1936]: /usr/libexec/bazzite-hardware-setup: line 268: -f: command not found
```

This is coming from a recent change (https://github.com/ublue-os/bazzite/commit/389f4a55c905b1efef50dd0bce9ad373543add3d) that is trying to update the `/etc/tuned/ppd.conf` file if it exists, but the existence check is missing brackets.


KyleGospo:
Thanks!

PR #1991, opened 2024-12-18
feat(deck): Smart video handling, allow configuring.

SuperRiderTH:
This adds verification to the video files, allowing them to be replaced if they have been changed or become corrupt.

Additionally, this also allows for video overrides to be disabled. A new just command was added to configure this.

### DIFF
--- a/system_files/deck/shared/usr/share/ublue-os/just/85-bazzite-image.just
+++ b/system_files/deck/shared/usr/share/ublue-os/just/85-bazzite-image.just
@@ -144,6 +144,39 @@ install-speaker-firmware ACTION="":
     fi
     rm /tmp/bazzite-speaker.sh
 
+# Configure video overrides for Steam
+configure-override-videos ACTION="":
+    #!/usr/bin/bash
+    source /usr/lib/ujust/ujust.sh
+    OPTION={{ ACTION }}
+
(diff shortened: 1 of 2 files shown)


system_files/deck/shared/usr/share/ublue-os/just/85-bazzite-image.just
```diff
@@ -144,6 +144,39 @@ install-speaker-firmware ACTION="":
...
+    #!/usr/bin/bash
+    source /usr/lib/ujust/ujust.sh
+    OPTION={{ ACTION }}
+
+    if [ ! -f $HOME/.local/share/Steam/config/uioverrides/movies/bazzite_novideo ]; then
+        VIDEO_STATE="${b}Enabled${n}"
+    else
+        VIDEO_STATE="${b}Disabled${n}"
+    fi
+
```
HikariKnight, line 158: i would suggest avoiding empty lines like this when writing bash (or other script languages) in justfiles, the linter sometimes really hates it. and the just syntax highlighters also dislike the empty lines a lot :sweat_smile:

HikariKnight (approved):
Since i have seen some people be unable to switch the startup movie when the bazzite one exists in the folder, im approving this as i think its a good ujust to have.

KyleGospo (approved):

PR #2000, opened 2024-12-20
fix: Adjust yafti and sunshine ujust for layered version of sunshine

wolfyreload:
If we decide to switch to sunshine being included in the Bazzite image, its service would be off by default like sshd. So I've adjusted the logic in the yafti.yml files to only enable the service. I've adjusted the `ujust setup-sunshine` script too accordingly.

I'm not certain if the sunshine-workaround.service is needed or not but I've included it in the enable and disable logic

### DIFF
--- a/system_files/deck/shared/usr/share/ublue-os/firstboot/yafti.yml
+++ b/system_files/deck/shared/usr/share/ublue-os/firstboot/yafti.yml
@@ -31,8 +31,7 @@ screens:
           description: A self-hosted game stream host for Moonlight
           default: false
           packages:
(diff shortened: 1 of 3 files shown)


HikariKnight:
Pretty sure the sunshine workaround is still needed for setcaps IIRC
i will approve and let @KyleGospo merge when ready either to main or testing wherever he wants it

HikariKnight (approved):

KyleGospo:
The caps setting is required if it's layered only, when we do it in the image we're able to fix it ahead of time.

KyleGospo (approved):

PR #2008, opened 2024-12-22
chore(yafti): Remove Sunshine from the Bazzite Portal

nicknamenamenick:
If Sunshine being part of the image now works fine, then removing it from the Bazzite Portal should also happen.  Please note, this commit does **not** remove the `ujust` commands.


KyleGospo:
This should instead be renamed to enable sunshine, the ujust enables the service so it runs in the background.

wolfyreload:
Already changed the wording in yafti to "Enable Sunshine" in #2000

nicknamenamenick:
Closed in favor of #2000

PR #2009, opened 2024-12-22
feat: create interactive rebase helper

wolfyreload:
I've created an interactive rebase helper with an associated ujust command based of the one in Bluefin/Aurora. It does have an additional feature where you can rebase between images but I have excluded the ability to rebase between desktop environments to prevent additional support tickets.


wolfyreload:
Asked for feedback in the Ublue discord. Seems to be a bit controversial if the ability to rebase between images should be in the tool at all. Might make sense to convert this to a draft pull request

Issue #2017, opened 2024-12-24
ujust setup-boot-windows-steam failing on desktop versions of Bazzite


wolfyreload:
### Describe the bug

If you run `ujust setup-boot-windows-steam` you get the error `sh: line 1: /usr/bin/steamos-add-to-steam: No such file or directory` looks like we don't include the `/usr/bin/steamos-add-to-steam` script on the desktop images but only in the deck images. It would still be useful having this script work on the desktop version, especially for the nvidia users. I've only tested in bazzite-gnome-nvidia but I suspect that we'll have this issue on all desktop images.

I'll put in a PR when I get a chance after Christmas

### What did you expect to happen?

I expected the boot-windows shortcut to get added as an executable in steam

### Output of `rpm-ostree status`

```shell
State: idle
Deployments:
● ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-gnome-nvidia:testing
                   Digest: sha256:138588f3ecec1a2ada575489db3b2188e676273b1e6676907f5824860d391ab4
                  Version: testing-41.20241222.1 (2024-12-22T18:29:52Z)
          LayeredPackages: cosmic-store
```

### Hardware

HP ProBook 450 G5

wolfyreload:
This has been merged in and appears to be working. Closing issue.

PR #2018, opened 2024-12-25
fix(ujust): removed removed script dependencies for ujust setup-boot-windows-steam

wolfyreload:
This is a fix for #2017, looks like the `/usr/bin/steamos-add-to-steam` file isn't available on Bazzite Gnome desktop, so I've removed the reliance on this script file in the `ujust setup-boot-windows-steam` script


MurderFromMars:
You're an absolute legend for this quick fix tysm

wolfyreload:
@MurderFromMars you are welcome! I missed that it didn't work in bazzite-gnome when I made the ujust script in the first place 😅

PR #2022, opened 2024-12-27
Update set_default_user.tmpl

skynetssl:
bazzite to skynetssl login credentials - plain text still, skynetssl and skynetssl!


KyleGospo:
what?

Issue #2035, opened 2024-12-31
[FR] Just script for docker-ce + vscode dev containers setup


mowglixx:
## Context
![Discord convo with kyle that said -dx is coming](https://github.com/user-attachments/assets/1f55a3e6-c09b-4626-bf48-b6c224403c08)

I'm pleased to hear that bazzite*-dx is coming, I'd happily accept bluefin-deck as well but the gaming has been so straight forward with bazzite that I can't bail.

## What I'm proposing
I'd like to add a justfile for a full docker layered install and a vscode layered install for A "devmode" pre -dx release and possibly have this transition into -dx and main have a rebase when this is ready.

for now this issue is open for branching only for said justfiles.

### Image

Desktop

noelmiller:
Bazzite-DX is the right place for this request to go to (if it doesn't already have this feature): https://github.com/ublue-os/bazzite-dx

PR #2036, opened 2024-12-31
feat: Pre dx devmode

mowglixx:
See #2035


mowglixx:
could do with patching with the whole remote dev extension pack but cba

system_files/desktop/shared/usr/share/ublue-os/just/87-bazzite-devcontainers.just
```diff
@@ -0,0 +1,32 @@
...
+    #!/usr/bin/bash
+    # check for updates
+    ublue-update --wait
+    # add yum repo
+    echo -e "[docker-ce-stable]\nname=Docker CE Stable - $basearch\nbaseurl=https://download.docker.com/linux/fedora/$releasever/$basearch/stable\nenabled=1\ngpgcheck=1\ngpgkey=https://download.docker.com/linux/fedora/gpg" | sudo tee /etc/yum.repos.d/docker-ce.repo > /dev/null
+    # install the layer
+    rpm-ostree install docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin 
+    # Enable the service
+    systemctl enable docker
+    echo 'Complete, please reboot to apply changes.'
```
wolfyreload, line 24: Don't you need to also run `systemctl enable docker.socket`?

mowglixx: good spot, I'll add it now

KyleGospo:
It might be worth visiting a vscode distrobox instead of layering it for the future.

KyleGospo:
It might be worth visiting a vscode distrobox instead of layering it for the future.

KyleGospo (approved):

mowglixx:
> It might be worth visiting a vscode distrobox instead of layering it for the future.

Agreed, the problem is viewing the host docker socket so it'd mean a few other steps, obviously in the -dx release it can be a regular package or maybe a bluefin dev could do the flatpak magic?

HikariKnight (approved):

wolfyreload:
@mowglixx I think I found one more issue. The .just files are imported manually in the [Containerfile](https://github.com/ublue-os/bazzite/blob/main/Containerfile#L738). So your custom scripts ujust commands won't work until you add an entry there too

mowglixx:
thanks for pointing that out, would have missed that, I've been testing with `just --justfile ./thisfile.just --choose` to make sure it's all working in just, the commands run with no errors (appart from this is working already), if someone has a vm env to test this with a fresh install I'd really appreciate it.

Pushing the change now.

wolfyreload:
I have a VM all set up, I'll test as soon as I get a chance. Probably tomorrow sometime.

wolfyreload (approved):
I'll need to test in a VM but everything looks good to me

mowglixx:
Thanks, I've been meaning to do the just script for virt-manager today so I might end up doing it myself, I'll drop a message if i do and I'm going to take a look at the image building process for my own image, I've been admiring @HikariKnight 's  [personal image](https://github.com/HikariKnight/Deckstation) for a starting point, I've done passthrough before so was good to see some familiar commands in the place it should be. =D

KyleGospo:
No need for anything virt-manager, it's a flatpak and already done the optimal way here.

KyleGospo:
I'm going to give this a containerization pass later today, we want to avoid layers at all costs. After that I'll be happy to merge into testing.

mowglixx:
no no no, I meant I'm going to run the just script (I fresh installed the other day from cachyOS) to get virt manager all set up so might run the VM to test ^this further myself

mowglixx:
echo was printing null vars from shell

KyleGospo:
https://github.com/ublue-os/bazzite/commit/1ac8f4902599277ede222137abc05d7a1291e33b

^ Here's docker as a container. I'm going to leave this PR open as a workshop for a potential best-solution for VSCode as well, may also end up being distrobox until DX.

system_files/desktop/shared/usr/share/ublue-os/just/87-bazzite-devcontainers.just
```diff
@@ -0,0 +1,34 @@
...
+    # Enable the service
+    systemctl enable docker
+    # Enable the socket
+    systemctl enable docker.socket
+    echo 'Complete, please reboot to apply changes.'
+
+add-user-to-docker-group:
+    #!/usr/bin/bash
+    if ! grep -q "^docker:" /etc/group; then
+      grep '^docker:' /usr/lib/group | sudo tee -a /etc/group > /dev/null
```
duhow, line 31: are `regexp` patterns enabled by default?
```suggestion
    if ! grep -qE "^docker:" /etc/group; then
      grep -E '^docker:' /usr/lib/group | sudo tee -a /etc/group > /dev/null
```

mowglixx:
Had some time away and started using the flatpak with a remote Ubuntu VM and docker installed not ideal but it's working, I'd like to continue working on this but I'm starting to realise that the DX mode needs to be a rebase like in bluefin due to docker being installed in the image and not requiring extra layers.

I'll try to put in some time this weekend learning how make the images work, tbh tho if DX mode is gunna be a rebase it may as well include the vscode package with the extensions installed for Dev containers.

mowglixx:
Think flatpak would be overkill in the instance of having purpose installed packages in the base image, maybe there is a licensing issue?

system_files/desktop/shared/usr/share/ublue-os/just/87-bazzite-devcontainers.just
mowglixx: I'm afraid that's a little outside my scope, I'm not super familiar with grep, copied this from working code in the input group justfile and changed the word from input to docker, I then added the : to stop dockerroot from being added to the users grouplist

mowglixx: Also, it worked 😅

castrojo:
> I'll try to put in some time this weekend learning how make the images work, tbh tho if DX mode is gunna be a rebase it may as well include the vscode package with the extensions installed for Dev containers.

Making a bazzite-dx will be easier once bazzite finishes transitioning to bootc/dnf, @tulilirockz and I have been experimenting with making DX easier to implement when using bootc (see the work here: https://github.com/centos-workstation/achillobator)

It'll be much easier that how we did it in bluefin, tulilirockz worked with Timothee Ravier to add QEMU/KVM support to virt-manager via a flatpak extension so that won't even need to be on the image - then it's just a few dnf lines to add docker, vscode, and the fonts. Everything else is just adding packages from distro into packages.json.

(That's how I would do it)

mowglixx:
So an extra build target in the containerfile, Then you could rebase?

PR #2055, opened 2025-01-03
feat(grub): hide grub in existing installs

antheas:
Grub is ugly so lets hide it again with a hidden timeout set to 2. We have not had an oopsie in 5 months, lets keep it that way, combine it with a health check, and remove this eyesore.

Only applies if grub timeout is 5s and is a one-shot.

Needs to be paired with an ISO fix probably so this will be for current installs only.


KyleGospo:
If we're gonna do this, let's make sure it's -deck only and have green boot in a working state

antheas:
Maybe we will do a variant which just lowers the timeout if its 5 to 3 because those 2 extra seconds somewhat bug me

PR #2057, opened 2025-01-03
feat(greenboot): Add simpler greenboot variant

antheas:
The greenboot project is a bit too complicated for our needs and could footgun us.

Make a variant that sets the default option to ostree 1 if boot fails twice in a row.

Lets squash this one.

motd is temporarily removed as it does not apply anymore.


KyleGospo (approved):

antheas:
Most of greenboot is built into fedora already. We only need the ostree service

antheas:
We also need to unset boot_counter to finalize an update. Do it 5 min after boot. grub-boot-success will also set boot_success after 2 min on its own.

https://github.com/rhboot/grub2/blob/6cac608cbe05b95ec2903897ad19dbd0499ab60d/docs/grub-boot-success.service

PR #2084, opened 2025-01-07
chore(docs): Remove TuneD, update fsync kernel and add a missing point on README-BR

koitorin:
Should be up to date with the last two commits.


EPOCHvoyager:
Period was missing on the original as well

Issue #2115, opened 2025-01-13
`/var/log/libvirt` would not be correctly labeled


Zeglius:
> Similar with Win11 vm has well. I got same issue with libvirt even in manual or iso is the same case. 
> 
> 
> ### Output of `bootc status`
> 
> ```shell
> ● Booted ostree
>            Commit: fa2dee84eefe2e04779baef56b1e40e50dde4ef61756a517955f716382ae4f84
> 
>   Rollback ostree
>            Commit: e58257b15a96dc9b090fb457f078ed2e295a2aa7557a4a00cc79d0e3f9e9c821
> ```
> 
> ### Output of `groups`
> 
> ```shell
> wolfshards wheel libvirt plugdev
> ```
> 
> ### Extra information or context
> 
> ### Screenshot of config before begining the creation:
> 
> ![Image](https://github.com/user-attachments/assets/7bb3c699-d030-493a-b514-b26d8dc72de0)
> 
> ### Neofetch Desktop Configuration: 
> ![Image](https://github.com/user-attachments/assets/58e340c2-90d7-4da9-82b0-f9acaa4652f2)
> 
> ### Fallback error logs: 
> 
> Impossible de terminer l’installation : « Impossible de se connecter à virtlogd: Impossible d'ouvrir le fichier : /var/log/libvirt/qemu/Windows11.log: Permission non accordée »
> 
> Traceback (most recent call last):
>   File "/app/share/virt-manager/virtManager/asyncjob.py", line 71, in cb_wrapper
>     callback(asyncjob, *args, **kwargs)
>   File "/app/share/virt-manager/virtManager/createvm.py", line 2008, in _do_async_install
>     installer.start_install(guest, meter=meter)
>   File "/app/share/virt-manager/virtinst/install/installer.py", line 726, in start_install
>     domain = self._create_guest(
>              ^^^^^^^^^^^^^^^^^^^
>   File "/app/share/virt-manager/virtinst/install/installer.py", line 667, in _create_guest
>     domain = self.conn.createXML(initial_xml or final_xml, 0)
>              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/app/lib/python3.12/site-packages/libvirt.py", line 4545, in createXML
>     raise libvirtError('virDomainCreateXML() failed')
[...]

PR #2116, opened 2025-01-13
fix(virt): Enhance service setup and restrict root execution for virtualization script

Zeglius:
Improve the bazzite-libvirtd-setup.service to restore SELinux context and prevent running the setup-virtualization script as root.

Fixes issue #2115.


system_files/desktop/shared/usr/lib/systemd/system/bazzite-libvirtd-setup.service
p5, line 1: QQ - how will this be invoked?  
Do you need to `systemctl enable` it?

p5 (commented):
The only thing this PR doesn't appear to do is create the /var/log/libvirt directory.  So I think you'd be running restorecon on a non-existant location (unless simply starting the libvirtd service creates it).

HikariKnight (approved):

KyleGospo (approved):

PR #2126, opened 2025-01-13
Initial support for BitTorrent distribution of ISOs

mskiptr:
Hello everyone! Before even becoming a user, I present you with this patchset. The aim here is to make BitTorrent an alternative supported way of obtaining Bazzite ISO images for installation.

The PR is marked as a draft because not everything here is ready to be merged just yet. While the first patch should work (and result in `.torrent` files being published for each built ISO), I wasn't able to test it on my own. When I try to run the GH action from a cloned repo, the build fails at the "Determine Flatpak Dependencies" step (`docker: Error response from daemon: Head "https://ghcr.io/v2/mskiptr/bazzite/manifests/stable": denied.`).

The second patch is a work-in-progress. I would like to mark each generated `.torrent` with the information which build number | version | revision it corresponds to. But how do I obtain that? I couldn't find it anywhere among all variables in that YAML file.

Finally, the third patch adds the official Fedora tracker. It isn't necessary, since there's always the DHT and peer exchange, so the torrent should work without a tracker but adding one might still be beneficial. The patch is marked as a draft because using that tracker will have to be coordinated with the Fedora infrastructure admins most likely. (And btw, if this _really_ needs a tracker, we can always look for some general-purpose one.)

Now, how will the swarm start up? At first I thought that for each variant and version someone will have to obtain the ISO manually and inject it into a torrent client. That would certainly be _not ideal_ to say the least, but at least it should be simpler than [getting someone else to send you parts of the file](https://www.answeroverflow.com/m/1328306949741350963) when the transfer keeps failing for you.

But it turns out we have a much better option. There's this thing called Web Seeds. It basically means that a `.torrent` file can contain a list of URIs that should be tried automatically, side-by-side with any actual BitTorrent peers.

The apparent lack of support for HTTP range requests from the current CDN might throw a wrench in that. But even then having an official `.torrent` available would still be an improvement here. Another problem might be that currently the web download link is not unique between consecutive Bazzite releases. So if someone tries to download the ISO using a `.torrent` file from an older release, the software might get confused.

### DIFF
--- a/.github/workflows/build_iso.yml
+++ b/.github/workflows/build_iso.yml
@@ -147,7 +147,26 @@ jobs:
           additional_templates: '/github/workspace/installer/lorax_templates/remove_root_password_prompt.tmpl /github/workspace/installer/lorax_templates/set_default_user.tmpl'
           repos: '/github/workspace/bazzite.repo /etc/yum.repos.d/fedora.repo /etc/yum.repos.d/fedora-updates.repo'
 
-      - name: Move ISOs to Upload Directory
+      - name: Generate a .torrent file
+        id: create-torrent
+        shell: bash
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y mktorrent
+          TORRENT_DIR=${{ github.workspace }}/torrent
+          mkdir ${TORRENT_DIR}
+          container_uri=docker://${{ steps.registry_case.outputs.lowercase }}/${{ matrix.image_name }}:${{ steps.generate-tag.outputs.tag }}
+          installer_version=$(skopeo inspect $container_uri | jq -r '.Labels["org.opencontainers.image.version"]')
+          cd ${TORRENT_DIR} && mktorrent \
+            --announce="udp://fosstorrents.com:6969/announce" \
+            --announce="http://fosstorrents.com:6969/announce" \
+            --comment="Installer for "${{ steps.build.outputs.iso_name }}" version "$installer_version \
+            --piece-length=20 \
+            --web-seed=https://download.bazzite.gg/${{ steps.build.outputs.iso_name }} \
+            ${{ steps.build.outputs.iso_path }}/${{ steps.build.outputs.iso_name }}
+          echo "torrent-dir=${TORRENT_DIR}" >> $GITHUB_OUTPUT
+
+      - name: Move ISO, checksum and torrent to Upload Directory
         id: upload-directory
         shell: bash
         run: |
@@ -155,9 +174,10 @@ jobs:
           mkdir ${ISO_UPLOAD_DIR}
           mv ${{ steps.build.outputs.iso_path }}/${{ steps.build.outputs.iso_name }} ${ISO_UPLOAD_DIR}
           mv ${{ steps.build.outputs.iso_path }}/${{ steps.build.outputs.iso_name }}-CHECKSUM ${ISO_UPLOAD_DIR}
+          mv ${{ steps.create-torrent.outputs.torrent-dir }}/${{ steps.build.outputs.iso_name }}.torrent ${ISO_UPLOAD_DIR}
           echo "iso-upload-dir=${ISO_UPLOAD_DIR}" >> $GITHUB_OUTPUT
 
-      - name: Upload ISOs and Checksum to Job Artifacts
+      - name: Upload ISO, checksum and torrent to Job Artifacts
         if: github.ref_name == 'testing'
         #if: github.event_name == 'pull_request'
         uses: actions/upload-artifact@v4
@@ -169,7 +189,7 @@ jobs:
           compression-level: 0
           overwrite: true
 
-      - name: Upload ISOs and Checksum to R2
+      - name: Upload ISO, checksum and torrent to R2
         if: github.event_name == 'workflow_dispatch' && github.ref_name == 'main'
         shell: bash
         env:


AzemaViator:
For reference, you can't just use the Fedora tracker blindly. The Fedora Infrastructure Team must submit every torrent.
https://docs.stg.fedoraproject.org/en-US/infra/release_guide/torrentrelease/

mskiptr:
> For reference, you can't just use the Fedora tracker blindly. The Fedora Infrastructure Team must submit every torrent. https://docs.stg.fedoraproject.org/en-US/infra/release_guide/torrentrelease/

Indeed. It would need to be coordinated with them somehow. But also automated and overall handled very differently than their regular procedure, which means it would require even more assistance from them. So that patch definitely needs to wait. (It's a draft and it shows.)

But if that turns out impossible (e.g. because of licensing issues), we can either forgo a tracker entirely or use some general-purpose one instead. I guess we will want to avoid ones even remotely associated with piracy, so something "Linux-related" would be the best. Maybe <https://linuxtracker.org/>? The documentation is pretty sparse, so from a quick glance I can't really tell if it's suitable. It seems to rely on using the web interface and accounts to submit new torrents (which could probably be automated with a simple `curl` incantation from GH actions). But can we upload just the `.torrent`s and not entire ISOs? And how will it handle name collisions every single week? These need further research I guess.

> For the web seed this is blocking range support on the downloads. https://community.cloudflare.com/t/public-r2-bucket-doesnt-handle-range-requests-well/434221

Yeah. I've now found a few reports like that. Seems to be a Cloudflare bug (or at least an edge-case in the HTTP specification which doesn't mesh well with many clients). Nothing conclusive tho

I was hoping to test how well Transmission handles it. That's why I would love to see that first patch tested in action. It would let me try downloading an ISO with just the web seed so I could check how reliable it is for me. But it has just occurred to me I can simply create this `.torrent` file on my end. (That's assuming I can make it before the next ISOs are generated. What an awful timing lol)

AzemaViator:
From my understanding, the web seed will not function at all because it depends on the Range header being respected and working correctly. I have not tried making a torrent to test webseeding yet, but I have confirmed the range header does not result in expected returns from the download cdn.

There does seem to be [fosstorrents](https://fosstorrents.com/), which has an open [tracker](https://fosstorrents.com/tracker/); they also have a discord for having official partnerships. One of which is [ultramarine](https://fosstorrents.com/distributions/ultramarine-linux/)

On the topic of whether or not to have a tracker, DHT is not something everyone can use. Depending on NATs, some clients require a tracker to function at all.

castrojo:
> There does seem to be [fosstorrents](https://fosstorrents.com/)

I joined their discord in december with the intent on investigating - seems like a good fit for us, might be worth investigating and then dropping over some cash to help for infra?

.github/workflows/build_iso.yml
```diff
@@ -147,17 +147,34 @@ jobs:
...
+      - name: Generate a .torrent file
+        id: create-torrent
+        shell: bash
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y mktorrent
+          TORRENT_DIR=${{ github.workspace }}/torrent
+          mkdir ${TORRENT_DIR}
+          cd ${TORRENT_DIR} && mktorrent \
+            --announce="http://torrent.fedoraproject.org:6969" \
```
AzemaViator, line 159:
```suggestion
            --announce="udp://fosstorrents.com:6969/announce" \
            --announce="http://fosstorrents.com:6969/announce" \
```
As the fedora project tracker is not open I suggest using fosstorrents's tracker

.github/workflows/build_iso.yml
```diff
@@ -147,17 +147,35 @@ jobs:
...
+        shell: bash
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y mktorrent
+          TORRENT_DIR=${{ github.workspace }}/torrent
+          mkdir ${TORRENT_DIR}
+          cd ${TORRENT_DIR} && mktorrent \
+            --announce="udp://fosstorrents.com:6969/announce" \
+            --announce="http://fosstorrents.com:6969/announce" \
+            --comment="TODO: put at least a version number here" \
```
AzemaViator, line 161:
```
skopeo inspect docker://${{ steps.registry_case.outputs.lowercase }}/${{ matrix.image_name }}:${{ steps.generate-tag.outputs.tag }} | jq -r '.Labels["org.opencontainers.image.version"]'
```
This can be used to get the release version. ie
```
$ skopeo inspect docker://ghcr.io/ublue-os/bazzite:stable | jq -r '.Labels["org.opencontainers.image.version"]'
41.20250106.3
```
I am not sure if skopeo or jq are available in this job, but you should be able to use it to get something in the comments like `Installer for bazzite-deck-nivida version 41.20250106.3`

.github/workflows/build_iso.yml
mskiptr: [I will clean up the commits in a bit]

mskiptr:
> From my understanding, the web seed will not function at all because it depends on the Range header being respected and working correctly. I have not tried making a torrent to test webseeding yet, but I have confirmed the range header does not result in expected returns from the download cdn.

I've now tested this with Transmission 4.0.6. I couldn't get it to work with <https://download.bazzite.gg/bazzite-deck-stable.iso> as the web seed, but whether R2 reliably supports range requests seems to vary between people, so it might work for someone else. Then I tried the same, but with the image served by Python's `http.server`. It does not support range requests either, so it didn't work. Finally, I used (RangeHTTPServer)[https://github.com/danvk/RangeHTTPServer] – a drop-in replacement, specifically created to support this feature and Transmission picked it up right away. I didn't even need to restart anything.

> There does seem to be [fosstorrents](https://fosstorrents.com/), which has an open [tracker](https://fosstorrents.com/tracker/); they also have a discord for having official partnerships. One of which is [ultramarine](https://fosstorrents.com/distributions/ultramarine-linux/)
> 
> On the topic of whether or not to have a tracker, DHT is not something everyone can use. Depending on NATs, some clients require a tracker to function at all.

Yeah. I meant that in case we couldn't find a suitable tracker, getting this done would still be useful to some people.

But while testing these things out I found one more reason not to use trackerless torrents: `file` and `xdg-open` do not understand them! (I should probably report that…)

castrojo:
I've reached out to fosstorrents and applied for support for our images.

AzemaViator:
> I've now tested this with Transmission 4.0.6. I couldn't get it to work with https://download.bazzite.gg/bazzite-deck-stable.iso as the web seed, but whether R2 reliably supports range requests seems to vary between people, so it might work for someone else. Then I tried the same, but with the image served by Python's `http.server`. It does not support range requests either, so it didn't work. Finally, I used (RangeHTTPServer)[https://github.com/danvk/RangeHTTPServer] – a drop-in replacement, specifically created to support this feature and Transmission picked it up right away. I didn't even need to restart anything.

From what I linked before in the Cloudflare community support thread, R2, where the ISOs are stored, does support range requests. However, when using a custom domain, it applies Cloudflare's cache to the files, which does not support range headers on files over 500 MB. Setting a page rule for the (sub)domain to bypass Cloudflare's cache fixes the issue, as all requests directly hit R2, which will handle the Range header right.

mskiptr:
Here are two examples for anyone who wants to test them out.
* <https://0x0.st/8oqG.torrent/bazzite-deck-stable.iso.torrent>
* <https://0x0.st/8oqk.torrent/bazzite-deck-stable.iso_trackerless.torrent>

They were created with these two commands:
```sh
mktorrent --announce={udp,http}"://fosstorrents.com:6969/announce" --piece-length=20 --web-seed=http{,s}"://download.bazzite.gg/bazzite-deck-stable.iso" .../bazzite-deck-stable.iso
mktorrent --piece-length=20 --web-seed=http{,s}"://download.bazzite.gg/bazzite-deck-stable.iso" .../bazzite-deck-stable.iso
```

What to check for:
1. Does the current CDN works as a web seed for you?
2. Does the trackerless version connect to any other peers? (I'm probably the only one for now. And don't worry if it's quite slow – I have the upload speed limited to 80 KiB/s as to not overwhelm and clog up my uplink.)
3. Can you connect to the tracker? (Doesn't seem to work for me, but I haven't looked into it yet.)

AzemaViator:
For the trackers alone I can connect to the UDP but not the http, I am not sure why that is the case.
image
The CDN setup as it is currently for webseeding just ends up in it trying to connect making a connection and then failing then retrying over and over.
image
I do not have a working DHT setup on my network so I can not test if that functions.

mskiptr:
I've pushed a v2 and I will address your earlier replies tomorrow.

.github/workflows/build_iso.yml
mskiptr: Thanks! Yes, both [`skopeo`](https://github.com/actions/runner-images/blob/d0ef4c965fc6a368f06e2e314fec80a5bae15a7e/images/ubuntu/scripts/build/install-container-tools.sh#L4) and [`jq`](https://github.com/actions/runner-images/blob/d0ef4c965fc6a368f06e2e314fec80a5bae15a7e/images/ubuntu/scripts/build/configure-apt.sh#L4) are available by default in `ubuntu-latest` images.

mskiptr:
I went over all the loose ends here and prepared a list of answers, questions and ideas. But it has been a long day for me and I'm simply too tired right now to turn all of that into a clear and coherent text. Unfortunately, that will have to wait for tomorrow. Sorry for the delay and thanks for all the assistance so far!

mskiptr:
Alright, I'm back working on this. Sorry for making you all wait for far longer than I had stated.

There's a lot to unpack here, so I divided the whole thing into sections. Some of the later parts are general ideas on things I noticed when researching and working on this. These don't need to be acted on and some are frankly quite far-fetched. But I still would love to hear your thoughts on them (as long as you have the time and energy for it).

I put todo checkboxes next to the points directly linked to and necessary for this PR.

## Cloudflare R2 vs Cloudflare proxy cache

> From what I linked before in the Cloudflare community support thread, R2, where the ISOs are stored, does support range requests. However, when using a custom domain, it applies Cloudflare's cache to the files, which does not support range headers on files over 500 MB.

You're right! When I had first looked at it I saw a bunch of seemingly conflicting reports but no official statement from Cloudflare support. For the record, [here](https://community.cloudflare.com/t/public-r2-bucket-doesnt-handle-range-requests-well/434221/4)'s the OP of that thread explaining the whole situation.

> Setting a page rule for the (sub)domain to bypass Cloudflare's cache fixes the issue, as all requests directly hit R2, which will handle the Range header right.

- [ ] Do you know how we should go about doing that? Is there an admin that manages this or maybe the configuration is automatically deployed from files in some repo?

## Other web seeding ideas

Besides uploading the build artifacts to R2, the GitHub action also stores them within GitHub itself for other workflows to access. Could we use GitHub as a second web seed here? We [can get stable URLs for them now](https://github.com/actions/upload-artifact/issues/27), but [these only work with authentication](https://github.com/actions/upload-artifact/issues/51). Thankfully there's this [GitHub app](https://github.com/oprypin/nightly.link) that proxies the traffic and enables anonymous downloads. Alternatively, we could just attach the artifacts to their respective releases, as these are already being created anyway. ([rel](https://github.com/marketplace/actions/torrent-action))

Tangentially, it would be awesome if some torrent clients out there could still handle web seeds that don't support range requests. Transmission handles such cases [by just not using them](https://github.com/transmission/transmission/issues/4595) (and then trying over and over I guess). I couldn't easily find what libtorrent does. I mean, when there's no other source available, trying to download the entire file in one go is a much better option than failing indefinitely. Probably more trouble than it's worth but I'll ask how hard it would be to implement that.

## DHT and trackers

DHT most likely works, as my node had some activity even though neither of the tracker URIs worked on my end. It received about a megabyte and sent nothing (I have ~120MB total, mostly from the earlier LAN web seed tests I did earlier). Oh, and for a split second I also saw some peers connect (one from Hungary and one from France).

- [ ] We should investigate why the fosstorrents.com tracker doesn't appear to work.
- [ ] Also, should we look for a second tracker? Even if the current problem is from a misconfiguration on our end, having a backup option would be very useful for mitigating other outages in the future.

While we're at it, could someone else go over my logic here and tell me whether I understand this correctly? When using multiple trackers, you organize them in tiers. Several tracker URLs can be put into the same tier for the purpose of load balancing, as just one tracker from a tier will be chosen at any particular time. On the other hand, tiers are tried in order until a satisfactory number of peers is obtained. So if we find a second tracker, it should definitely be put in a different tier. But what about our case, when the tracker offers two URLs, one per protocol? Should they go in the same tier (for load balancing) or in consecutive tiers (to give priority to the protocol we consider better)? Or does it not matter because it will be handled reliably either way and there's no point in chasing any efficiency here (since this traffic is easily dwarfed by the transfer of the actual data). Though it matters much more for the server, so me should think it through at least a bit.

What about HTTPS trackers? It would be one step forward in terms of maintaining privacy, but it's a public torrent anyway so it really doesn't matter, right? But what is the threat model here tho? And what about adoption / compatibility? Same goes for HTTP web seeds I guess.

## Testing

- [ ] So, how do we properly test this?

Before opening this PR I tried to run the workflow from my clone of the repo. It failed, which is not surprising since I didn't really have anything set up. Would I need to get this working or can someone else test it much more easily?

## Future steps

Once the `.torrent` files are working and available at `https://download.bazzite.gg/`, they will have to be added to the website. [This](https://github.com/KyleGospo/bazzite.gg) is the repo, right? At that point, we could also offer magnet links. ([This](https://nutbread.github.io/t2m/) might then be useful.)

Is there anything we would like to coordinate with fosstorrents.com on? On the main page they have a "recent releases" section. Maybe we should help in getting Bazzite onto that list? (or maybe not, since it has a very rapid release cadence)

## Random thoughts and ideas:

- (somewhat) stable URLs? unique image names?
- string quoting? most of the code will break as soon as a space appears in any variable
- how much can the code be refactored? how many things outside it depend on its behavior?
- can _its dependencies_ be tweaked? e.g. can the variables | parameters be renamed?
- [ShellCheck](https://www.shellcheck.net/)?
- could we swap Ubuntu for Alpine?
- could all that build logic be moved to an actual build tool instead?
- is slower release cadence planned for some time in the future? I mean, sooner or later the device support should be consistent enough that creating installers for just the major releases will make more sense. And at least the `-deck` images seem to already use an internet connection during initial setup, so maybe it could just rely on updating the OS then? Or even be mostly replaced by a network installer, with a fallback of just the major releases having a networkless one?

castrojo:
>  We should investigate why the fosstorrents.com tracker doesn't appear to work.

> Is there anything we would like to coordinate with fosstorrents.com on? On the main page they have a "recent releases" section. Maybe we should help in getting Bazzite onto that list? (or maybe not, since it has a very rapid release cadence)

Thanks for your work on this! 

I went ahead and made a donation to fosstorrents.com on universal blue's behalf (so we can get aurora and bluefin going too), I'm currently unable to spend time coordinating with them but ideally we can connect this work to whatever automation they have to make it all nice and magical? 

I've sent them the link to this issue but they have a discord if someone wants to follow up! (I'm too time limited right now to help!)

ledif:
I started working on a torrent mirror for Aurora/Bluefin/Bazzite on my own before it was pointed out that this effort was already being undertaken here.

Here is what I have so far: https://github.com/ledif/ublue-torrent-mirror. The full list of torrents available are at  https://gaia.feralhosting.com/ublue.

Maybe my effort can be seen as an unofficial torrent mirror while we sort out the fosstorrents situation.

mskiptr:
> I started working on a torrent mirror for Aurora/Bluefin/Bazzite on my own before it was pointed out that this effort was already being undertaken here.
> 
> Here is what I have so far: https://github.com/ledif/ublue-torrent-mirror. The full list of torrents available are at https://gaia.feralhosting.com/ublue.

Awesome! I've taken a quick glance at the code and it will definitely be useful here! It could help us side-step a bunch of problems until we get them fixed properly. Like the fact that without a working web seed, the swarm doesn't have a way to start up.

Could [`ublue-torrent-manager.py`](https://github.com/ledif/ublue-torrent-mirror/blob/main/pyinfra/files/ublue-torrent-manager.py) check whether any relevant`.torrent` files are present next to the ISOs, and if that's the case just use them directly (instead of generating its own ones)? I could prepare a patch for that. Though since we're at it: I couldn't find any copyright or licensing statements. While simple config files and short scripts might be not copyrightable, could you still add some license file or at least [SPDX headers](https://spdx.dev/learn/handling-license-info/) to make the legality of contributing (and distribution) fully unambiguous?

Also, is it a full-blown VPS, or rather a managed ( | multiuser | time-sharing) server instead? Tho either way it's not really important – as long as it's not just web hosting. If we could run a backup tracker there, it could lift this PR from 'just a draft' to 'a working MVP' basically overnight. Would trying that be alright with you?

Are old ISO files being removed? I couldn't really find any code doing that. There were some "TODO"s for stopping torrents and whatnot, so I guess all that is yet to be implemented. I think it would be good to leave at least one older version of each file seeded, so that people who start their downloads right before a new release aren't just left hanging. Also, Could the `-CHECKSUM` and `.torrent` files be hosted indefinitely? Being able to verify the authenticity of an older file is sometimes quite useful. But that's more of a question for the ublue team, since without any official signatures you can't really trust a checksum anyway (be it PGP or simply HTTPS).

And finally, we should think about eventually retiring this. It's costing you money after all. (It has just occurred to me I could have done something similar from the very beginning, as I do have a ~$3 / month Debian VPS for email and stuff… Tho knowing myself, I would be putting it off until after I finally revamp how it's deployed.) Anyway, if we get any working web seeds (either fix the Cloudflare cache problem or switch to something else), fix the fosstorrents.com setup, possibly get some backup trackers + web seeds and get the PR merged, then there should be no reason to run this anymore. So it is hopefully just a temporary solution, but a very welcome one. Thank you for helping!

ledif:
@mskiptr Thanks for the comments. A few responses.

I added an Apache 2.0 License. Feel free to send a PR for any improvements that you have.

Feral Hosting offers what are called "seedboxes," which I agree doesn't really fit the definition of a VPS since it is a multi-user shared host without the ability to select an OS or install system-wide software. I am okay with supporting the seedbox for a few of months or so, but obiously long-term this should be sunset.

There is no code currently to replace the old ISOs, but my thoughts are to simply delete the old ISO and regenerate a new torrent for the new ISO. The reason is that long-term, we should ideally rely on webseeds and the ISO URLs themselves are not versioned and thus only point to the latest ISO. My goal with this is to provide an alternative method to download the latest ISOs through torrents, and not necessarily to archive older versions of the ISOs.

That being said, it wouldn't hurt to keep an archive of the old torrents themselves. But I do not wish to start archiving and seeding old ISOs when the original ISO source (Cloudflare R2) doesn't even provide them anymore.

mskiptr:
> I added an Apache 2.0 License. Feel free to send a PR for any improvements that you have.

Thanks @ledif!

> without the ability to select an OS or install system-wide software. I am okay with supporting the seedbox for a few of months or so, but obiously long-term this should be sunset.

Yeah. The proper long term solution cannot rely on resources of a single volunteer. But do you have any objections (or see any technical obstacles) to running a closed tracker on that server alongside the current seeding node?

> My goal with this is to provide an alternative method to download the latest ISOs through torrents, and not necessarily to archive older versions of the ISOs.
> 
> That being said, it wouldn't hurt to keep an archive of the old torrents themselves. But I do not wish to start archiving and seeding old ISOs when the original ISO source (Cloudflare R2) doesn't even provide them anymore.

Indeed. I would like to eventually make the file names versioned (since non-unique names will cause issues with web seeds in certain scenarios). Archiving (or better yet, explicit support for matching reproducible builds) would be nice too, but that is outside the scope for now and a task for official ublue infra anyway.

ledif:
I have a few thoughts regarding running a closed tracker:
- What would be the advantage of running a tracker versus using something like `udp://tracker.opentrackr.org:1337/announce`? I am using opentrackr currently and  it appears to be working just fine for the various bazzite/aurora/bluefin torrents I've tried to download from various machines.
- Setting up and maintaining a tracker introduces another point of failure and maintenance burden, which is especially exacerbated on a multi-user server without access to containers or systemd. We would have to build it from source and ensure that it is continuously running using a mix of cronjobs or something similar, which seems brittle in my opinion.

mskiptr:
Agh, sorry for the long wait @ledif

> What would be the advantage of running a tracker versus using something like `udp://tracker.opentrackr.org:1337/announce`?

Well, we do not want the project to be even remotely associated with piracy. So until I get an explicit ack, I would rather avoid public open trackers here (unless they are very clearly just about FOSS).

> We would have to build it from source and ensure that it is continuously running using a mix of cronjobs or something similar, which seems brittle in my opinion.

Yeah. Setting that up sounds like a pain. Especially when it's supposed to be just a stopgap solution.

p5:
Thank you for the PR!

I'm going to close this one since we recently announced Bazzite (along with the other UBlue projects) are now available on FossTorrents - https://fosstorrents.com/distributions/bazzite/

If I've misunderstood something, and the PR is not the same, please ping myself or another maintainer and we can reopen this PR.

PR #2131, opened 2025-01-14
feat(waydroid) - Add waydroid force restart shortcut

aarron-lee:
Occasionally the Waydroid container becomes unresponsive, but a simple force restart fixes it. This just adds a convenience shortcut for it.

### DIFF
--- a/system_files/desktop/shared/usr/share/ublue-os/just/82-bazzite-waydroid.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/82-bazzite-waydroid.just
@@ -29,6 +29,7 @@ setup-waydroid ACTION="":
       fi
       sudo waydroid init -c 'https://ota.waydro.id/system' -v 'https://ota.waydro.id/vendor'
       sudo restorecon -R /var/lib/waydroid
+      cp /usr/share/applications/waydroid-container-restart.desktop ~/.local/share/applications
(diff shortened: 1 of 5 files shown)


system_files/desktop/shared/usr/share/ublue-os/just/82-bazzite-waydroid.just
```diff
@@ -29,6 +29,7 @@ setup-waydroid ACTION="":
       fi
       sudo waydroid init -c 'https://ota.waydro.id/system' -v 'https://ota.waydro.id/vendor'
       sudo restorecon -R /var/lib/waydroid
+      cp /usr/share/applications/waydroid-container-restart.desktop ~/.local/share/applications
```
aarron-lee, line 32: not sure if this will work as-intended, or if there's a better approach. the idea is to add the waydroid-container-restart desktop entry after the user inits waydroid

HikariKnight (approved):

system_files/desktop/shared/usr/share/ublue-os/just/82-bazzite-waydroid.just
HikariKnight: we can merge to testing instead and find out

KyleGospo (approved):

PR #2160, opened 2025-01-21
fix: update sunshine package ver (#2159)

### LINKED ISSUE #2159: Sunshine - update ujust version (NVFBC support)
### Describe the package

Requesting that the build of Sunshine be updated to at least v2025.118.151840 as this version supports NVFBC support WITHOUT needing patches to the Nvidia driver on consumer cards.

When running the Bazzite Nvidia beta distro, the sunshine build included in ujust and Bazzite's package manager is many versions behind making this feature impossible.

It also seems to be very difficult to force upgrade sunshine, or at the very least apply the Nvidia patch that allows NVFBC on consumer cards.

What this leads to currently for most Nvidia users is a stream that has stuttering audio or huge performance issues when using a game scope enabled Bazzite distro on Nvidia RTX series cards.

Simple fix would be to push sunshine build to Bazzite using at least version v2025.118.151840 which has been tested and works for fedora based systems.

Thank you!

### Image

Nvidia

xXJSONDeruloXx:
bump sunshine package version to address NVFBC support per #2159 

Also investigating added ujust action to check for newer version to upgrade

Closes #2159

### DIFF
--- a/spec_files/sunshine/sunshine.spec
+++ b/spec_files/sunshine/sunshine.spec
@@ -1,5 +1,5 @@
 Name:           sunshine
-Version:        2025.101.215655
+Version:        2025.118.151840
 Release:        100%{?dist}
 Summary:        Corrects a name change issue between original unofficial sunshine builds and the new official ones
 License:    	GPLv2


KyleGospo:
We don't build sunshine ourselves anymore, it comes from the official copr

PR #2189, opened 2025-01-26
fix: Whole github 404 page was fetched into `/etc/distrobox/incus.ini`

pk-kampanart:
Because filename changes from origin


KyleGospo:
Thanks!

PR #2192, opened 2025-01-27
fix(hardware-setup): Explicitly disable composefs

EyeCantCU:
Composefs hasn't been enabled upstream yet due to a few issues. See: https://github.com/ublue-os/main/issues/712


travier:
https://github.com/ublue-os/main/issues/712#issuecomment-2616647439

Issue #1570, opened 2024-08-29
Zram can be configured more optimally by using lz4 instead of zstd1


ahydronous:
### Describe the bug

Zram is currently configured to use zstd1, which is suboptimal

### What did you expect to happen?

I've spent an inordinate amount of time optimizing zram on my system.

### Benchmarks on `zstd-1` vs `lz4`
https://www.reddit.com/r/Fedora/comments/mzun99/new_zram_tuning_benchmarks/

### An explanation of vm.swappiness
https://docs.kernel.org/admin-guide/sysctl/vm.html#swappiness
https://stackoverflow.com/questions/72544562/what-is-vm-swappiness-a-percentage-of

### Overcommitting memory (zram being bigger than RAM size) is *good*
https://issuetracker.google.com/issues/227605780

### Gathered wisdoms
- Swap is usually written sequentially but read randomly
- Zram writes don't matter nearly as much as because Linux page cache data is already in RAM, and RAM>RAM transfers are absurdly quick
- Zram hardcoded blocksize is 4K
- For system memory performance applications care jack about bandwidth, latency (CAS etc in hardware, IOPS in software) is what matters
- You still want a swapfile, even with zram, due to the fact that you're gonna have incompressible or very idle pages that you want to evict from RAM. This goes double for the Deck, which will be RAM starved on some games due to the GPU claiming a lot of RAM. More on this later.

### IOPS benchmark on Samsung 970 EVO Plus 1TB
```
- lz4:      2 030 000 (!)
- zstd1:    820 000
- 970 EVO:  15 300
```

### Compression ratios on mixed data
```
- lz4:    2.1
- zstd1:  2.9
```
This is very relevant for the Deck, because <12GB is right where in a lot of scenarios the benefits of extra memory from `zstd1` start to outstrip the latency benefits of `lz4`.
Valve probably has a lot of profiled data, but as far as I've been able to tell, even the heaviest games don't go much over 4GB of VRAM.

### Swappiness
Swappiness can be derived via formula. On kernel.org, they state 
> For example, if the random IO against the swap device is on average 2x faster than IO from the filesystem, swappiness should be 133 (x + 2x = 200, 2x = 133.33).

You can reduce that to `(yx = 200 -x)`, where `y` is filesystem-to-swap IO ratio.
With the 970 Evo Plus as example again, we have aforementioned read IOPS values. `970EVO vs. lz4 =  15 300 / 2 030 000 = 0.008`, so `0.008` is our ratio.
[...]

PR #2202, opened 2025-01-28
fix: General ZRAM Optimizations

PeronGH:
### Summary

This PR implements several ZRAM improvements, including:
- Adjusting the default ZRAM size to match system RAM
- Optimizing kernel parameters for better ZRAM utilization
- Updating the outdated README

### Background

I've been using Bazzite on my Legion Go for several weeks and noticed performance issues related to the 16GB RAM limitation. Specifically, Palworld multiplayer would frequently crash and occasionally freeze the system due to out-of-memory (OOM) errors.

After experimenting with various solutions, I found that increasing ZRAM size and tweaking kernel parameters significantly improved stability. While I personally use ZSTD compression on my device, I noticed this project switched from ZSTD to LZ4, so I've maintained LZ4 in this PR to respect that decision.

These changes should help users with similar RAM constraints get better performance and stability from their systems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ Variant designed for usage as an alternative to SteamOS on the Steam Deck, and f
 - Update break something? Easily roll back to the previous version of Bazzite thanks to `rpm-ostree`'s rollback functionality. You can even select previous images at boot.
 - Steam and Lutris preinstalled on the image as layered packages.
 - [Discover Overlay](https://github.com/trigg/Discover) for Discord pre-installed and automatically launches in both Game mode and on the Desktop if Discord is installed. [View the official documentation here](https://trigg.github.io/Discover/bazzite).
-- Uses ZRAM<sub><sup>(4GB)</sup></sub> with the ZSTD compression algorithm by default with the option to switch back to a 1GB swap file and set a custom size for it if desired.
+- Uses ZRAM<sub><sup>(same size as RAM)</sup></sub> with the LZ4 compression algorithm by default with optimized kernel parameters.
 - [LAVD](https://crates.io/crates/scx_lavd) and [BORE](https://github.com/firelzrd/bore-scheduler) CPU Schedulers for smooth and responsive gameplay.
 - Kyber I/O scheduler to prevent I/O starvation when installing games or during background `duperemove` process.
 - Applies SteamOS's kernel parameters.
--- a/system_files/deck/shared/etc/systemd/zram-generator.conf
+++ b/system_files/deck/shared/etc/systemd/zram-generator.conf
@@ -1,3 +1,0 @@
-[zram0]
-zram-size=max(ram/4, 4096)
-compression-algorithm=lz4
--- a/system_files/desktop/shared/etc/systemd/zram-generator.conf
+++ b/system_files/desktop/shared/etc/systemd/zram-generator.conf
@@ -1,2 +1,3 @@
 [zram0]
+zram-size=ram
 compression-algorithm=lz4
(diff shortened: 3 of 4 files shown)


PeronGH (commented):
Note: I've removed the mention of swap file as it was no longer supported, as noted in this [forum post](https://universal-blue.discourse.group/t/how-do-i-switch-from-zram-to-the-1gb-swap-file/3771).

KyleGospo:
Hi! Thanks for the PR, it's much appreciated.
We actually set all of your options here via tuned, this was broken due to an oversight but was recently fixed. The only thing we differ on is the zram size, can you explain a bit more about why you're using zram matching total ram size here?

PeronGH:
> Hi! Thanks for the PR, it's much appreciated. We actually set all of your options here via tuned, this was broken due to an oversight but was recently fixed. The only thing we differ on is the zram size, can you explain a bit more about why you're using zram matching total ram size here?

Thanks for letting me know about tuned - I wasn't aware these settings were supposed to be managed that way.

Regarding the ZRAM size, I set it to match total RAM because I've run into memory issues with some games, particularly on APU systems where 16GB needs to serve as both system RAM and VRAM. For example, when hosting Palworld multiplayer, ZRAM can compress up to 10GB using zstd (potentially more with lz4). With only 4GB of ZRAM, the game crashes within an hour. Sons of the Forest has similar issues, crashing every 1-2 hours due to insufficient memory.

That said, I realize matching total RAM might not make sense for systems with more than 16GB, since they don't face the same memory pressure. Even on 16GB systems, when ZRAM compresses over 10GB, the system starts thrashing and FPS drops significantly. A better approach might be to set a minimum of 10GB ZRAM for gaming systems, and cap it at 50% of total RAM for higher-end systems.

EPOCHvoyager:
I think setting it to half RAM size, maxing out to 8GB should be more than enough - I believe this is the current Fedora default? 

> when ZRAM compresses over 10GB, the system starts thrashing and FPS drops significantly.

Likewise, it might be good to look into setting up a writeback device as mentioned in #1570, to flush out incompressible pages and ease these memory-constrained situations. Though this might come into conflict with the current setup for hibernation.

EPOCHvoyager:
I have run into `systemd-oomd` on occasion here, so I don't think this is without merit.

xXJSONDeruloXx:
This has been open for 6 months, can we revisit and make a decision on this one? I'm gonna give it an ack since I recently ran into this issue on a 5700u laptop and OOM killer was a bit agressive, and manually bumping zram allocation and swappiness did significantly improve things. Most handhelds we target should have more than enough storage to spare IMO

CaptainN:
Oh, I didn't see this in search - I have had some substantial instability do to memory pressure, even on 16GB machine (mostly productivity - not sure Bazzite cares that much in this case). Here's my experience, but it sounds similar to what you've done here: zstd compression and RAM * 1 for zram - these changes solved my stability issues. https://github.com/ublue-os/bazzite/issues/3409

Mitsu13Ion:
I've had similar issues #3409  and increased it to half of my current RAM (32GB), so I think it's time to merge the PR?

KyleGospo:
> I've had similar issues #3409 and increased it to half of my current RAM (32GB), so I think it's time to merge the PR?

There is no universe where half of 32GB of ram being swap makes sense

CaptainN:
8GB makes sense - 4GB is not enough.

Mitsu13Ion:
> > I've had similar issues #3409 and increased it to half of my current RAM (32GB), so I think it's time to merge the PR?
> 
> There is no universe where half of 32GB of ram being swap makes sense

Yes, I play extremely modded versions of games, and without that, it just doesn't work. No need to talk about universes nobody is omniscient.

PR #2205, opened 2025-01-29
feat(fcitx5): add `kcm-fcitx5` for better integration with KDE

coxde:
Hiya! I'd like to suggest adding `kcm-fcitx5` for better integration with KDE:

1. Users can tweak their Fcitx5 settings in `System Settings / Input Method` instead of launching Fcitx5 individually.
2. Users can use `KDE Plasma (Experimental)` theme for Fcitx5, which is nice because it can dynamically generate theme to fit their Plasma theme. https://planet.kde.org/weng-xuetian-2022-07-04-fcitx-5-plasma-theme-support/

Thank you!

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -639,6 +639,7 @@ RUN --mount=type=cache,dst=/var/cache/rpm-ostree \
             fcitx5-mozc \
             fcitx5-chinese-addons \
             fcitx5-hangul \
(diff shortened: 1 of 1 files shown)


KyleGospo (approved):

KyleGospo:
Thanks! Good suggestion

PR #2207, opened 2025-01-29
feat: add USB hot plugging configuration for virtualization script

itinkering:
When using virt-manager as a flatpak is is impossible to use spice-client-glib-usb-acl-helper due to Flatpak sandboxing.
Without this tool USB redirection in running VM fails due to insufficient permissions.

To solve this issue I've added new options to virtualization ujust script which enable/disable udev rule allowing USB redirection (adding `uaccess` tag to all USB devices).

Rule uses `72` prefix to comply with https://github.com/systemd/systemd/issues/4288#issuecomment-348166161

### DIFF
--- a/system_files/desktop/shared/usr/share/ublue-os/just/84-bazzite-virt.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/84-bazzite-virt.just
@@ -30,6 +30,8 @@ setup-virtualization ACTION="":
       echo "  Use 'vfio-on' to select Enable VFIO drivers"
       echo "  Use 'vfio-off' to select Disable VFIO drivers"
       echo "  Use 'kvmfr' to select Autocreate Looking-Glass shm"
+      echo "  Use 'usbhp-on' to select Enable SPICE USB hot plugging"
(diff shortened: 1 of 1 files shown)


HikariKnight:
I think this should be made part of enabling/disabling virtualization since you would usually want the ability to use usb redirection

itinkering:
Without the udev rule, it is still possible to redirect USB to turned off VM via "Add device" -> "USB host device" because that is handled by qemu itself.
Hotplugging is done is userspace, and a udev rule for all USB devices can be a security hole for some folks, therefore I opted for a separate toogle.
If its ok, then sure, I can put it into the general enable/disable section.

HikariKnight:
> Without the udev rule, it is still possible to redirect USB to turned off VM via "Add device" -> "USB host device" because that is handled by qemu itself. Hotplugging is done is userspace, and a udev rule for all USB devices can be a security hole for some folks, therefore I opted for a separate toogle. If its ok, then sure, I can put it into the general enable/disable section.

fair enough

HikariKnight (dismissed):

PR #2217, opened 2025-01-31
fix: restore-gnome-de-settings - indent syntax, handle empty dirs, add error checks

xXJSONDeruloXx:
seems just syntax caused issues with recipe, also took the opportunity to improve error handling

image


bpjoestar:
Thank you, that works!

PR #2226, opened 2025-02-01
chore(ci): Use dnf5 instead of rpm-ostree

m2Giles:
Possible implementation for switching to dnf5


Zeglius:
Dnf5 is already used in unstable branch. 
Only thing missing is dealing with this error
https://github.com/ublue-os/bazzite/actions/runs/13098963407/job/36544960017#step:24:138

m2Giles:
You should probably version lock the kernel packages. Probably kernel-devel
from the repos is being pulled in as a dependency.

PR #2227, opened 2025-02-02
chore(TuneD): Move tweaks to inherit upstream

EPOCHvoyager:
- This should lower the maintenance burden in automatically keeping up
with upstream changes to the default profiles.


EPOCHvoyager:
In case I messed up anything in the file structure for building, please correct it or let me know.

KyleGospo (approved):

PR #2238, opened 2025-02-05
chore(ci): support combined akmods-kernel-cache in build

bsherman:
Relates: ublue-os/akmods#295

This is expected to fail until the akmods PR is merged.


KyleGospo:
main is le kill

PR #2260, opened 2025-02-10
feat: Add DX variant

Zeglius:


nicknamenamenick:
For those interested:
- [Bazzite Developer Experience](https://github.com/ublue-os/bazzite-dx) which matches Bluefin/Aurora's Developer Experience
- [Bazzite Gamer Developer Experience](https://github.com/ublue-os/bazzite-gdx) which is centered around game development.

PR #2336, opened 2025-03-05
feat: Introduce scopebuddy

HikariKnight:
This is a modified version of @coolavery's script from https://github.com/ValveSoftware/gamescope/issues/835#issuecomment-2496383830
It introduces the `scb` command which is a symlink to `scopebuddy` that will do the whole unset `LD_PRELOAD` for gamescope and re-apply it for `%command%` song and dance.

`scb` can replace `gamescope` in the launch options for games in steam
ex: `gamescope -w 1920 -h 1080 -W 2560 -H 1440 -r 180 --hdr-enabled -- %command% --nointro --etc` --> `scb -w 1920 -h 1080 -W 2560 -H 1440 -r 180 --hdr-enabled -- %command% --nointro --etc`

Furthermore if the user has a config file `~/.config/scopebuddy/scb.conf` it will be sourced before starting gamescope.
If the above file exists and a config with the games AppID exists in `~/.config/scopebuddy/AppID/` then that file will be loaded instead of the `scb.conf` unless a custom one is set through the SCB_CONF env var.

The file can set a `SCB_DEFAULT_ARGS` variable to provide `scb` with default gamescope arguments as long as the user gives `scb` no arguments before the `--` in the launch options, providing __any__ args to `scb` will make it ignore `SCB_DEFAULT_ARGS` completely and use whatever the user provides in the launch options.

This lets the user apply a default set of env vars and gamescope args that they use regularly for games to avoid messes like
`env -u LD_PRELOAD ENABLE_GAMESCOPE_WSI=0 SteamDeck=0 XKB_DEFAULT_LAYOUT=no MANGOHUD_CONFIG=read_cfg,preset=2 gamescope --mangoapp -f -w 2560 -h 1440 -W 2560 -H 1440 -e --force-grab-cursor --hdr-enabled -- env LD_PRELOAD="$LD_PRELOAD" %command%`

Instead they can define the defaults in `~/.config/scopebuddy/scb.conf` like this
```ini
export XKB_DEFAULT_LAYOUT=no
export MANGOHUD_CONFIG=read_cfg,preset=2
SCB_DEFAULT_ARGS="--mangoapp -f -w 2560 -h 1440 -W 2560 -H 1440 -r 180 --force-grab-cursor --hdr-enabled -e --adaptive-sync"
```

now this is simplified in the launch options as
`SteamDeck=0 ENABLE_GAMESCOPE_WSI=0 scb -- %command%`

If the user wants a specific config they do not need to remember for a specific game (in this example warframe) then they can make an AppID config in `~/.config/scopebuddy/AppID/230410.conf`, `230410` is the Steam App ID for warframe. In this we can add
```ini
export MANGOHUD_CONFIG=read_cfg,preset=3
export SteamDeck=0
export ENABLE_GAMESCOPE_WSI=0
SCB_DEFAULT_ARGS="--mangoapp -f -w 1920 -h 1080 -W 2560 -H 1440 -r 60 --force-grab-cursor --hdr-enabled -e --adaptive-sync"
```
So now i can set the launch options for warframe to `scb -- %command%` and it will source `~/.config/scopebuddy/scb.conf` then source `~/.config/scopebuddy/AppID/230410.conf` which will override anything set by `scb.conf` if they touch the same variables, in this example it will upscale the game from 1080p to 1440p and limit refresh rate to 60 while setting the `SteamDeck` and `ENABLE_GAMESCOPE_WSI` environment variables, change MangoHud to preset 3 and keep the XDG keyboard layout set by `scb.conf`.
This also means it will run any bash code written in the config file before launching the game too!

Then to not use the gamescope args defined in `SCB_DEFAULT_ARGS` the user just has to supply any gamescope arg like so, as they clearly do not want the defaults then. (current implementation still will apply default env vars though)
`scb -r 180 -- %command%`

The user can also define a different file that contains different defaults too by overriding `SCB_CONF` with the new filename.
`SCB_CONF=1080_180hz_upscaled_hdr.conf scb -- %command%`
this will then load the config from `~/.config/scopebuddy/1080_180hz_upscaled_hdr.conf` instead

RFC:
* Unsure if we should make the script not load the config file at all if a gamescope arg is provided or if we should make it just load the env vars
* ~~Should we preemptively have this named something else in prep for upstream fixing this issue?~~

PROS:
* Fixes overlay when using nested gamescope in most titles
* Gives users a command to use for using nested gamescope with a default set of env vars and args (giving this a purpose once it is fixed upstream)

Limitations:
* ~~`%command%` has to be in double quotes, args after `"%command%"` does not need to be quoted unless you would normally~~ fixed with https://github.com/ublue-os/bazzite/pull/2336/commits/58f5bf648da87049fd4f1edd22a691fadfa6e2ab
* does not fix the overlay in all games (usually games where it didn't work to begin with)
* `gamemode.conf` will be forced as the only possible default config when running inside gamemode (appid configs will still be loaded as usual if they exist)

### DIFF
--- a/system_files/desktop/shared/usr/bin/scopebuddy
+++ b/system_files/desktop/shared/usr/bin/scopebuddy
@@ -41,14 +41,15 @@ else
     # Make the default config file
     mkdir "$SCB_CONFIGDIR"
     cat << 'EOF' > "$SCB_CONFFILE"
-# This is the config file that let's you assign defaults for gamescope
+# This is the config file that let's you assign defaults for gamescope when using the scopebuddy script
 # lines starting with # are ignored
 # 
-# Example for always exporting a specific environment variable for gamescope
+# Example for always exporting specific environment variables for gamescope
 #export XKB_DEFAULT_LAYOUT=no
+#export MANGOHUD_CONFIG=read_cfg,preset=2
 #
-# Example for providing default gamescope arguments through scopebuddy if no arguments are given to the scopebuddy script
-# To not use this default set of arguments, just launch goi with any gamescope option like "-r 60" and this variable will be ignored
+# Example for providing default gamescope arguments through scopebuddy if no arguments are given to the scopebuddy script, this does not need to be exported.
+# To not use this default set of arguments, just launch scb with any gamescope option like "-r 60" before the '-- "%command%"' and this variable will be ignored
 #$SCB_DEFAULT_ARGS="--mangoapp -f -w 2560 -h 1440 -W 2560 -H 1440 -r 180 --force-grab-cursor --hdr-enabled -e"
 EOF
 fi
--- a/system_files/desktop/shared/usr/bin/scopebuddy
+++ b/system_files/desktop/shared/usr/bin/scopebuddy
@@ -55,7 +55,7 @@ else
 #
 # Example for providing default gamescope arguments through scopebuddy if no arguments are given to the scopebuddy script, this does not need to be exported.
 # To not use this default set of arguments, just launch scb with any gamescope option like "-r 60" before the '-- "%command%"' and this variable will be ignored
-#$SCB_DEFAULT_ARGS="--mangoapp -f -w 2560 -h 1440 -W 2560 -H 1440 -r 180 --force-grab-cursor --hdr-enabled -e"
+#SCB_DEFAULT_ARGS="--mangoapp -f -w 2560 -h 1440 -W 2560 -H 1440 -r 180 --force-grab-cursor --hdr-enabled -e"
 EOF
 fi
 
--- a/system_files/desktop/shared/usr/bin/scopebuddy
+++ b/system_files/desktop/shared/usr/bin/scopebuddy
@@ -128,28 +128,7 @@ else
     # Transfer LD_PRELOAD before we unset it
     LD_PRELOAD_REAL=$LD_PRELOAD
 
-    # Create named pipe (fifo) to read from gamescope stderr
-    rm -f $gsout && mkfifo $gsout
-    trap 'rm -vf $gsout' EXIT
-
-    # Unset LD_PRELOAD for gamescope (as it breaks the overlay) then start gamescope
-    # shellcheck disable=SC2086
-    echo "Launching: env -u LD_PRELOAD $GAMESCOPE_BIN $gamescope_opts"
-    # shellcheck disable=SC2086
-    env -u LD_PRELOAD $GAMESCOPE_BIN $gamescope_opts > >(tee -ia "$gsout" >&2) 2>&1 &
-    gs_pid=$!
-
-    # Check the output file to locate the Xwayland display that gamescope launches
-    while read -r line; do
-        display=$(sed -nE 's/^.*Starting Xwayland on :([0-9]+).*$/\1/p' <<<"$line")
-        if [ -n "$display" ]; then
-            break
-        fi
-    done <"$gsout"
-
-    # Add back LD_PRELOAD and launch %command% on the gamescope xwayland display until it finishes
-    (DISPLAY=":$display" eval "env LD_PRELOAD=\"$LD_PRELOAD_REAL\" $command")
-
-    # Kill gamescope when done
-    kill "$gs_pid"
+    # Unset LD_PRELOAD for gamescope (as it breaks the overlay) then start gamescope with LD_PRELOAD set for %command% instead
+    echo "Launching: env -u LD_PRELOAD $GAMESCOPE_BIN $gamescope_opts -- env LD_PRELOAD=$LD_PRELOAD_REAL $command"
+    eval "env -u LD_PRELOAD $GAMESCOPE_BIN $gamescope_opts -- env LD_PRELOAD=$LD_PRELOAD_REAL $command"
 fi
--- a/system_files/desktop/shared/usr/bin/scopebuddy
+++ b/system_files/desktop/shared/usr/bin/scopebuddy
@@ -24,13 +24,17 @@ GAMESCOPE_BIN=${GAMESCOPE_BIN:-gamescope}
 while [[ $# -gt 0 ]]; do
     if [ "${1:-}" == "--" ]; then
         shift
-        # Add LD_PRELOAD from gamescope to %command% then exit loop
+        # Add LD_PRELOAD from gamescope to %command%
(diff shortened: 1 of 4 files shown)


HikariKnight:
@wolfyreload and @Zeglius i would like some input from both of you on this one

HikariKnight:
Suggested by Crono in discord:
* rename it to `scopebuddy` with the alias `scb`

HikariKnight:
next question would be, do we turn this into a package or something down the line :thinking:

wolfyreload:
@HikariKnight  I'm not using this as intended as I'm using it on bazzite-nvidia and it's pretty cool being able to just add `scb -- %command%` and not need to run the gigantic list of gamescope parameters. I wouldn't jump straight into making a package for it, maybe see what the adoption rate is like first. If people start wanting to use it outside of Bazzite then it will make some sense making a package out of it.

system_files/desktop/shared/usr/bin/scopebuddy
```diff
@@ -0,0 +1,89 @@
...
+    cat << 'EOF' > "$SCB_CONFIG"
+# This is the config file that let's you assign defaults for gamescope when using the scopebuddy script
+# lines starting with # are ignored
+# 
+# Example for always exporting specific environment variables for gamescope
+#export XKB_DEFAULT_LAYOUT=no
+#export MANGOHUD_CONFIG=read_cfg,preset=2
+#
+# Example for providing default gamescope arguments through scopebuddy if no arguments are given to the scopebuddy script, this does not need to be exported.
+# To not use this default set of arguments, just launch scb with any gamescope option like "-r 60" before the '-- "%command%"' and this variable will be ignored
```
wolfyreload, line 62: You still using your double quoted "%command%" here

HikariKnight: > You still using your double quoted "%command%" here

the cleanup never ends, thank you :rofl:

system_files/desktop/shared/usr/bin/scopebuddy
```diff
@@ -0,0 +1,101 @@
...
+if [ -f "$SCB_CONFIGFILE" ]; then
+    # If a config exists for this games STEAM_APPID and SCB_CONF is set to scb.conf (default)
+    if [ -f "$SCB_CONFIGDIR/AppID/${STEAM_APPID}.conf" ] && [ "$SCB_CONF" == "scb.conf" ]; then
+        # Source the STEAM_APPID specific config
+        source $SCB_CONFIGDIR/AppID/${STEAM_APPID}.conf
+    else
+        # Source the config from SCB_CONF
+        source $SCB_CONFIGFILE
+    fi
+    # If the user has supplied ANY args to gamescope, do not load the GOI_DEFAULT_ARGS
```
wolfyreload, line 54: Old reference in the comments here GOI_DEFAULT_ARGS

KyleGospo (approved):

system_files/desktop/shared/usr/bin/scopebuddy
```diff
@@ -0,0 +1,101 @@
...
+    shift
+done
+
+# Get the STEAM_APPID from %command%
+STEAM_APPID=$(echo $command | perl -pe 's/.+"AppId=(\d+)"\s.+/\1/')
+
+# Load the GOI config file if it exists then apply the default args
+if [ -f "$SCB_CONFIGFILE" ]; then
+    # If a config exists for this games STEAM_APPID and SCB_CONF is set to scb.conf (default)
+    if [ -f "$SCB_CONFIGDIR/AppID/${STEAM_APPID}.conf" ] && [ "$SCB_CONF" == "scb.conf" ]; then
```
wolfyreload, line 47: Not sure if you really need the AppId's need to be a different folder, they should be easy enough to differentiate from the normal config files with your perl script above.

HikariKnight: Personally i think it would just look cleaner since all the configs will be a bunch of numbers and wont clash with a users custom "1080.conf" or something

HikariKnight:
Pretty sure this is feature complete now :sweat_smile:

EPOCHvoyager:
Commenting here to see in case someone has more information, but this seems to break menu use in things like Ryujinx (built with Avalonia). Even with the same exact arguments for gamescope — checked with `ps ax | grep gamescope` — and seemingly identical environment, menus will fullscreen flicker for a brief moment before closing when it is run with `scb`, but not `gamescope` directly — where they will work perfectly fine.

I suspect something might be upsetting XWayland, but otherwise am, along with @HikariKnight (who's much smarter than me here), at a loss. Does anyone have a lead as to what would be different in the environment between the two — in a way that might be affecting gamescope's compositing?

HikariKnight:
> Commenting here to see in case someone has more information, but this seems to break menu use in things like Ryujinx (built with Avalonia). Even with the same exact arguments for gamescope — checked with `ps ax | grep gamescope` — and seemingly identical environment, menus will fullscreen flicker for a brief moment before closing when it is run with `scb`, but not `gamescope` directly — where they will work perfectly fine.
> 
> I suspect something might be upsetting XWayland, but otherwise am, along with @HikariKnight (who's much smarter than me here), at a loss. Does anyone have a lead as to what would be different in the environment between the two — in a way that might be affecting gamescope's compositing?
> 
> P.S.: Do note that I'm testing this script by placing it in `/var/usrshare/bin`, not by rebasing.

at this point i am willing to chalk this up to some incompatibility with how the automatic LD_PRELOAD fix has to be applied, so far it seems either only flatpak applications are affected or only ryujinx, so either its a flatpak incompatibility or ryujinx incompatibility.

EPOCHvoyager:
> Commenting here to see in case someone has more information, but this seems to break menu use in things like Ryujinx (built with Avalonia). Even with the same exact arguments for gamescope — checked with `ps ax | grep gamescope` — and seemingly identical environment, menus will fullscreen flicker for a brief moment before closing when it is run with `scb`, but not `gamescope` directly — where they will work perfectly fine.
> 
> I suspect something might be upsetting XWayland, but otherwise am, along with @HikariKnight (who's much smarter than me here), at a loss. Does anyone have a lead as to what would be different in the environment between the two — in a way that might be affecting gamescope's compositing?
> 
> P.S.: Do note that I'm testing this script by placing it in `/var/usrlocal/bin`, not by rebasing.

Solved by [097e35a](https://github.com/ublue-os/bazzite/pull/2336/commits/097e35aa66b2d620cbf998a847163e9a61ba7431).

PR #2366, opened 2025-03-12
feat: add hplip printer driver

dreamyukii:
add hplip printer driver


HikariKnight:
This is already included as part of the main images :)

PR #2375, opened 2025-03-13
feat: ujust cmd to enable wake from controllers with 2.4g dongle (non xbox)

xXJSONDeruloXx:


xXJSONDeruloXx:
hit a wall with this one, closing for now

PR #2384, opened 2025-03-14
fix: port changes from dnf to yum

ledif:
The changes to detect whether dnf is being run from a container (https://github.com/ublue-os/bazzite/pull/2357) should be ported to `/usr/bin/yum`.

Instead of copying it to the yum override, it makes sense to me to make `/usr/bin/yum` a link to `/usr/bin/dnf` instead.

I am currently making this change for Aurora in https://github.com/ublue-os/aurora/pull/270.


system_files/overrides/usr/bin/yum
Zeglius, line 1: File has `644` perms, you need to chmod it and push it again.

Zeglius (changes requested):

Zeglius (approved):

PR #2389, opened 2025-03-15
fix: downgrade dnf5 to avoid regression exposed by Terra

ledif:
Builds are failing right now because the latest dnf5 package version 5.2.11.0 has a regression that the Terra repos are exposing.

Related:
- https://github.com/ublue-os/aurora/pull/274
- https://github.com/rpm-software-management/dnf5/issues/2134
- https://github.com/ublue-os/aurora/issues/275


ledif:
This change seems to have fixed the main images but the NVIDIA images are still failing. I'm unsure if it's related.

PR #2390, opened 2025-03-15
ci: move to best-practices config for renovate

tulilirockz:
This will be a bit annoying due to it pinning everything but this is super important, we need to have consistent, pinned actions so we get rid of a few attack vectors.

https://docs.renovatebot.com/presets-config/#configbest-practices

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "config:recommended"
+    "config:best-practices"
   ]
 }
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,21 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "config:best-practices"
+    "config:best-practices",
+  ],
+
+  "rebaseWhen": "never",
+
+  "packageRules": [
+    {
+      "automerge": true,
+      "matchUpdateTypes": ["pin", "pinDigest"]
+    },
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -1,6 +1,21 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "config:recommended"
+    "config:best-practices",
```
antheas, line 4: not valid json

antheas:
So what does this do?

renovate.json
p5: There's other trailing commas in the same file.  
Trailing commas are valid json5, but not json.

p5:
> So what does this do?

The `config:best-practices` preset enables pinning GitHub Actions to their sha digests.

The first rule tells Renovate to automatically merge the PRs to pin these GHA sha digests once the required approvals are met
```
    {
      "automerge": true,
      "matchUpdateTypes": ["pin", "pinDigest"]
    },
```

The second rule excludes any container digest pinning inside YAML files in the `.github/workflows/` directory.  This would be useful when enabling multi-arch builds, since labelling requires a pre-release Podman version and therefore cannot be ran on Ubuntu 24.04 runners.

renovate.json
tulilirockz: To be completely honest I just copied this over from Bluefin LTS, didnt know this would fail

p5: Turns out it didn't fail. Renovate is smart enough to treat it as JSON5, but posted a warning in the dependency dashboard

PR #2424, opened 2025-03-26
chore(just): Remove CoolerControl Copr import

EPOCHvoyager:
CoolerControl is packaged in Terra, which is now included by default in all images; as that was already taking priority in release revisions anyway due to a higher number, it's probably better to remove this.


antheas:
Probably needs an accompanying change on the ujust that adds coolercontrol

EPOCHvoyager:
> Probably needs an accompanying change on the ujust that adds coolercontrol

That's what this is; what do you mean?

antheas:
Indeed you're right

KyleGospo (dismissed):

PR #2433, opened 2025-03-29
feat: Add ujust enable/disable-automounting

Zeglius:
It was causing issues, see https://www.reddit.com/r/Bazzite/s/0yV3nmTuDH 
Can be enabled with 'ujust enable-automounting' ~~or in yafti~~


antheas:
Shouldn't the change happen in the package? Do we need a package for this?

antheas:
My suggestion is if you want to keep it as a package topve the etc file somewhere else and create a symlink once someone runs the ujust to avoid drift

You cannot remove files from /etc

Zeglius:
> My suggestion is if you want to keep it as a package topve the etc file somewhere else and create a symlink once someone runs the ujust to avoid drift
> 
> You cannot remove files from /etc

Added `ujust disable/enable-automounting` for that case

PR #2452, opened 2025-04-05
feat: Add ckb-next

DeVelox:
In order to address the concerns that this might affect users of non-Corsair peripherals I've disabled the service by default and added a ujust command to enable it, similar to how Tailscale, Sunshine etc. are handled.


KyleGospo:
Unfortunately we tested that and this still broke peripherals.

PR #2456, opened 2025-04-05
feat(deck): Remove ENV GS_ENABLE_CTRL_12

AkazaRenn:
It's a follow up of https://github.com/ublue-os/bazzite/pull/1724, it's been half a year now I think we can remove the option added for backward compatibility and save new users from dealing with environment variables.

However, you don't need to merge it right now, feel free to keep it open until when you feel appropriate.


antheas:
do not touch the patch. Just enable the env var. (It is not ready to merge in any case; we need to get to that)

AkazaRenn:
@antheas done, so I may come back in another half a year for the patch update?

AkazaRenn:
https://github.com/ValveSoftware/steam-for-linux/issues/10247#issuecomment-2932677092

We might be able to get rid of this patch altogether in the future

PR #2461, opened 2025-04-07
Fix: Set -steamdeck alias in gamemode as well as desktop via ujust toggle-steamdeck-alias

xXJSONDeruloXx:
I discovered if using ujust toggle-steamdeck-alias on Bazzite to disable the -steamdeck launch flag, this works in desktop mode, but not in game mode (Gamescope session), where Steam still launches with -steamdeck.

In game mode, the CLIENTCMD variable is set in:

```/usr/share/gamescope-session-plus/sessions.d/steam```

as 

```export CLIENTCMD="steam -gamepadui -steamos3 -steampal -steamdeck"```

This setting ignores the flag file that desktop mode checks, so Steam is always launched with -steamdeck in game mode.

This update enhances the toggle-steamdeck-alias ujust recipe to fully support Bazzite’s immutable filesystem by using a writable override in ~/.config/gamescope-session-plus/sessions.d/steam.

Changes:
	•	Automatically creates the sessions.d override directory if it doesn’t exist.
	•	Writes a small shell script to ~/.config/gamescope-session-plus/sessions.d/steam that respects the ~/.config/bazzite/disable_steamdeck_flag file.
	•	Ensures CLIENTCMD is correctly set to include or exclude -steamdeck based on user choice.
	•	Informs user when a reboot is required to apply changes.

### DIFF
--- a/system_files/desktop/shared/usr/share/ublue-os/just/80-bazzite.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/80-bazzite.just
@@ -600,18 +600,34 @@ toggle-steamdeck-alias:
     #!/usr/bin/bash
     # Toggles whether Steam presents the system as a Steam Deck via -steamdeck
     override_flag="$HOME/.config/bazzite/disable_steamdeck_flag"
+    session_dir="$HOME/.config/gamescope-session-plus/sessions.d"
+    session_file="$session_dir/steam"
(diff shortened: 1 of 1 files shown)


HikariKnight:
Approving it for testing, however if this removes more than "just mangohud" from the side menu, users should be warned about it

Issue #2466, opened 2025-04-09
Can't seem to get Decky Loader to install no matter what I try.


ZenDeuo:
### Describe the bug

So fresh install of Bazzite for the SteamDeck, put it on my SteamDeck and installation went splendid which I'm very happy about, awesome stuff.

Get into Steam go through all that setup all is good.

Get to the desktop when I switch out of Game Mode to Desktop Mode and here's what I've tried so far to get Decky Loader to work but all attempts have failed thus far.

1. The Out Of Box Experience:
- Presented Me With An Easy Option For Installation Of Decky Loader.  |  Tried That And It Failed.

2. Using The `ujust setup-decky` Command.
- Tried This As People Recommended This Command If The OOBE Installer Fails To Install Decky Loader For You.  |  Again, Tried That And That Command Failed.

3. Finally, I Tried Using Decky Loaders Installer Directly From Their Site.
- Ran The Installer, Typed In My Sudo Password Etc "Installed" The Stable Version.  |  This Also Didn't Work

That's all I've tried thus far and none of these have fixed my issue and Decky Loader still does not get installed.

Here's a log from when I ran the `ujust setup-decky` command, output looks pretty similar if not 1:1 with the output I got from the OOBE Tool too.

```
deuo@deuodeck:~$ ujust setup-decky
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
100  3900  100  3900    0     0   2648      0  0:00:01  0:00:01 --:--:--  2648
[sudo] password for deuo:           
Installing Steam Deck Plugin Loader release...
mkdir: cannot create directory ‘/home/deuo/homebrew/services’: Permission denied
Installing version v3.1.5...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
Warning: Failed to open the file /home/deuo/homebrew/services/PluginLoader: No 
Warning: such file or directory
  0 24.2M    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (23) client returned ERROR on write of 16375 bytes
[...]

PR #2470, opened 2025-04-10
fix: Always chown homebrew directory when running setup-decky

Zeglius:
Fixes already existing installations like at #2466


antheas (approved):
lgtm

PR #2502, opened 2025-04-19
feat: move to `ublue-brew` for linuxbrew

tulilirockz:
This should entirely replace all these files, [ublue-brew](https://github.com/ublue-os/packages/tree/main/packages/ublue-brew) provides the update units, everything really that we need for bazzite and others. Needs some testing, though! Havent tried this at all on a VM yet


KyleGospo:
@tulilirockz need merge conflicts fixed

PR #2513, opened 2025-04-20
fix: Use bootc kargs to mount rootfs with zstd compression

Zeglius:
See:
https://discussion.fedoraproject.org/t/root-mount-options-are-ignored-in-fedora-atomic-desktops-42/148562/1

### DIFF
--- a/system_files/desktop/shared/usr/lib/bootc/kargs/20-rootfs.toml
+++ b/system_files/desktop/shared/usr/lib/bootc/kargs/20-rootfs.toml
@@ -1,0 +1,6 @@
+# Enable zstd compression in rootfs.
+# See:
+#   - https://bootc-dev.github.io/bootc/building/kernel-arguments.html#usrlibbootckargsd
+kargs = [
+    "rootflags=subvol=root,noatime,lazytime,discard=sync,compress-force=zstd:3,space_cache=v2",
+]


system_files/desktop/shared/usr/lib/bootc/kargs/20-rootfs.toml
```diff
@@ -0,0 +1,6 @@
+# Enable zstd compression in rootfs.
+# See:
+#   - https://bootc-dev.github.io/bootc/building/kernel-arguments.html#usrlibbootckargsd
+kargs = [
+    "rootflags=subvol=root,noatime,lazytime,discard=sync,compress-force=zstd:3,space_cache=v2",
```
p5, line 5: Are these kargs also applied when rpm-ostree handles the update?  I.E. if a user has any packages layered?

fiftydinar: > Are these kargs also applied when rpm-ostree handles the update? I.E. if a user has any packages layered?

Not even without layered packages in `rpm-ostree`, `bootc` must be used for the update for kargs to apply.

Zeglius: > > Are these kargs also applied when rpm-ostree handles the update? I.E. if a user has any packages layered?
> 
> Not even without layered packages in `rpm-ostree`, `bootc` must be used for the update for kargs to apply.

Oh, nevermind then, that wouldn't fly in bazzite

PR #2547, opened 2025-04-28
fix(deck): Update videos for latest Steam client.

SuperRiderTH:
Valve has decided to suddenly want to use the Steam OS variant of the startup movies.

It appears that Steam's Deck and OLED override is currently broken, but we are going to set up all of the different video variants just in case they decide to fix it.


antheas:
Already fixed in testing

antheas:
Rebase there and keep the ujust though, I missed that

Issue #2536, opened 2025-04-26
Dangling symlink to /usr/share/ublue-os/firstboot/launcher/login-profile.sh after yafti removal


bvitnik:
Hi. Yafti was removed in recent updates but a related symlink is still created in `/etc/profile.d`

https://github.com/ublue-os/bazzite/blob/8e0a65f741d80d735145a06c6d9680cba14485d9/Containerfile#L531

wolfyreload:
Thanks for picking this up, the change has been merged in. Going to close this.

PR #2564, opened 2025-04-30
chore: Remove symlink to missing script /usr/share/ublue-os/firstboot/launcher/login-profile.sh

wolfyreload:
yafti was removed, and the first boot script was also removed but we still have a symlink to it in the container file as reported here #2536

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -528,8 +528,6 @@ RUN --mount=type=cache,dst=/var/cache \
     rm -f /usr/share/vulkan/icd.d/lvp_icd.*.json && \
     rm -f /usr/lib/systemd/system/service.d/50-keep-warm.conf && \
     mkdir -p "/etc/profile.d/" && \
-    ln -s "/usr/share/ublue-os/firstboot/launcher/login-profile.sh" \
-    "/etc/profile.d/ublue-firstboot.sh" && \
     mkdir -p "/etc/xdg/autostart" && \
     cp "/usr/share/applications/discover_overlay.desktop" "/etc/xdg/autostart/discover_overlay.desktop" && \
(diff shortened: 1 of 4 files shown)


Containerfile
```diff
@@ -525,7 +525,6 @@ RUN --mount=type=cache,dst=/var/cache \
     rm -f /usr/share/vulkan/icd.d/lvp_icd.*.json && \
     rm -f /usr/lib/systemd/system/service.d/50-keep-warm.conf && \
     mkdir -p "/etc/profile.d/" && \
-    ln -s "/usr/share/ublue-os/firstboot/launcher/login-profile.sh" \
     "/etc/profile.d/ublue-firstboot.sh" && \
```
duhow, line 528: this line must be removed as well.
```suggestion
```

KyleGospo:
Thanks, fixed in testing since there was a merge conflict

Containerfile
```diff
@@ -527,8 +528,6 @@ RUN --mount=type=cache,dst=/var/cache \
     rm -f /usr/share/vulkan/icd.d/lvp_icd.*.json && \
     rm -f /usr/lib/systemd/system/service.d/50-keep-warm.conf && \
     mkdir -p "/etc/profile.d/" && \
```
wolfyreload, line 530: I only changed these two lines

spec_files/gnome-shell-extension-restart-to/gnome-shell-extension-restart-to.spec
```diff
@@ -0,0 +1,40 @@
+%global uuid restartto@tiagoporsch.github.io
```
wolfyreload, line 1: I'm a bit concerned here, I think with the force pushes you took something off testing then it showed those changes as if it's part of my pull request since those changes are still in my fork of Bazzite.

I have no idea what this is, not something I changed.

Issue #2432, opened 2025-03-29
Too many ways of installing emudeck, and some don´t work


ZeferinoLee:
### Describe the bug

There are too many ways of installing Emudeck, and some don´t work
- Bazzite Portal: Does not work. Does something but Emudeck is nowhere to be found
- ujust get-emudeck: Did not try
- ujust install-emudeck: Works

### What did you expect to happen?

I expected the option put in front of me to work (Bazzite Portal)

### Output of `rpm-ostree status`

```shell

```

### Hardware

Steam Deck LCD using bazzite-deck-gnome 2025-03-25

CheckYourFax:
@KyleGospo The bazzite portal one is obviously gone, but on -deck images you have `get-emudeck` and `install-emudeck` both of which do different things. One seems to install emudeck-electron, and the other ones just pulls a .desktop file from emudeck.com

wolfyreload:
Bazzite-portal is gone. We've removed the `ujust get-emudeck` (the script that doesn't work) and made an alias that "get-emudeck" just points to "install-emudeck". Going to close this

wolfyreload:
Also had a look at why the old `ujust get-emudeck` script was failing. It downloads the emudeck .desktop file. That file runs the <https://raw.githubusercontent.com/dragoonDorise/EmuDeck/main/install.sh> installer. In the installer it tries to use `dnf install jq zenity flatpak unzip bash fuse git rsync newt python`. In Bazzite, we've blocked running on dnf commands as it will just show "ERROR: Fedora Atomic images utilize rpm-ostree instead (and is discouraged to use)."

CheckYourFax:
> Also had a look at why the old `ujust get-emudeck` script was failing. It downloads the emudeck .desktop file. That file runs the https://raw.githubusercontent.com/dragoonDorise/EmuDeck/main/install.sh installer. In the installer it tries to use `dnf install jq zenity flatpak unzip bash fuse git rsync newt python`. In Bazzite, we've blocked running on dnf commands as it will just show "ERROR: Fedora Atomic images utilize rpm-ostree instead (and is discouraged to use)."

I've used it in the past and worked fine, unless it was updated. I think on deck images at least those packages are probably already available?

alshdavid:
`ujust install-emudeck` doesn't work for me. I get 

```
AppImageProvider.create_list_element_from_file() 1 missing required positional argument: 'return_new_el'`

[...]

PR #2587, opened 2025-05-04
chore(ujust): remove duplicate get-emudeck script

wolfyreload:
On the deck images there are two scripts for installing emudeck. "get-emudeck" and "install-emudeck". On desktop images we only have "install-emudeck" available. also "install-emudeck" is the script that used to be used when we had yafti.  So I have removed "get-emudeck"


wolfyreload:
We can close this related issue #2432

PR #2588, opened 2025-05-04
chore: switch copr repo for wallpaper engine kde plugin

dreamyukii:
switch to this copr repo for [wallpaper-engine-plugin-kde](https://copr.fedorainfracloud.org/coprs/jackgreiner/wallpaper-engine-kde-plugin/) , it also package the Wallpaper plugin and less -devel dependencies


KyleGospo:
We're trying to limit our use of copr that we don't control.

I brought the relevant changes over.

PR #2593, opened 2025-05-05
ci: Add new retagging action

Zeglius:
This allows us to pick an specific image by digest, and retag it to `:latest` or `:stable` (or any customized list of them). Useful for emergency regresions.


.github/workflows/retag.yml
```diff
@@ -0,0 +1,52 @@
...
+  packages: write
+
+on:
+  workflow_dispatch:
+    inputs:
+      image_ref:
+        description: |-
+          Image reference to retag.
+          Example:
+            ghcr.io/<owner>/<repo>/<image>:<tag>
```
antheas, line 21: It's better to hardcode everything.

Have it take an optional image tag and a source tag and make it add the tags stable, stable-FV, and latest.

If the image tag is missing it should do all of them.

Image should be bazzite and exclude the rest of the url

Description is placed as a label on a text box, so it needs to be a single line

antheas (changes requested):

KyleGospo (approved):

Zeglius:
Okay so forgot we are dealing with more than one image, so manually triggering an action per each image is silly. 

I'm thinking of closing this PR and instead writing a reusable action where we can specify a date tag (ex.: `41.20250409`) and push to the registries (ex.: `ghcr.io/ublue-os/bazzite`) listed in a yaml file

PR #2606, opened 2025-05-08
feat: add ujust 'fix-nuke-steam'

Zeglius:
Fixes issues related with 'public/steambootstraper_english.txt'

### DIFF
--- a/system_files/desktop/shared/usr/share/ublue-os/just/81-bazzite-fixes.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/81-bazzite-fixes.just
@@ -49,6 +49,13 @@ fix-reset-steam:
     bazzite-steam &
     exit 0
 
+# Fix issues related with 'public/steambootstraper_english.txt'. WARNING: It will probably delete your games installed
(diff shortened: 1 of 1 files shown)


CheckYourFax:
Maybe its a good idea to rename `fix-reset-steam` to something like `soft-reset-steam` to avoid confusion.

Zeglius:
> Maybe its a good idea to rename `fix-reset-steam` to something like `soft-reset-steam` or `redownload-steam` to avoid confusion (looks too similar), since the latter only forces a full redownload for when the client is stuck updating/validating (and is useful to make sure no changes where made to steam files). Have seen no other case where this justfile actually helped.

The `fix-reset-steam` can be used to remove stuff like customized css, if I'm not mistaken

antheas:
We don't do renames, previous one will have to be aliased

CheckYourFax:
That works, as it could only be visually ambiguous  when doing `--choose` which I think doesn't include aliases(?)

KyleGospo:
Not a good fix for this issue, fixed in an alt way

PR #2618, opened 2025-05-09
ci: Read fedora major version value from a single file

Zeglius:
This is in order to avoid duplication in several workflows


antheas:
Sounds like very complicated?

Zeglius:
> Sounds like very complicated?

Probably. Ah whatever, we can always change it by hand, retagging isnt that resource expensive, neither takes a lot of time.

PR #2648, opened 2025-05-17
fix: hide composefs volume in fastfetch

ledif:
This PR hides the 100% full read-only composefs mount that keeps confusing users.

### Before
Screenshot_20250517_093613

### After
Screenshot_20250517_093557


ledif:
I just realized that the main btrfs partition is labeled as read-only for some reason so it's hiding that as well. I will need to reevaluate this solution.

ledif:
This should be good to merge once https://github.com/fastfetch-cli/fastfetch/pull/1762 is accepted.

PR #2654, opened 2025-05-18
fix(desktop): correctly set DefaultTimeoutStopSec

Luap99:
The main systemd config requires the options to be set under the [Manager] section. without this the option is ignored and systemd logs a warning in the journal:
systemd[1]: /etc/systemd/system.conf.d/timeout.conf:1: Assignment outside of section. Ignoring.

Fixes: 3576341 ("chore: Move default timeout config to /etc for easy customization")


KyleGospo:
Merged, thank you so much!

PR #2656, opened 2025-05-19
feat: add uninstall option to Ujust setup-decky

xXJSONDeruloXx:
mainly for those decky moments where everything is busted, this makes it easy to quickly uninstall, not just swap from release and pre-release.


antheas:
I did a force push to nix that commit, can you remove it too?

Issue #2658, opened 2025-05-20
GNOME file chooser freezes applications randomly


scovik:
### Describe the bug

Multiple applications consistently freeze when opening files through the file chooser dialog. This issue:

* Affects applications across different toolkits (Libadwaita/GTK, Qt, Electron, etc.) including Firefox, Steam, Heroic, and others
* Occurs randomly – sometimes an application works normally, other times it freezes (effectively a coin toss)
* Began occurring recently without any configuration changes or system modifications
* Reproducible across several application types
* Freezes require force-quitting Libadwaita applications

### What did you expect to happen?

File chooser shouldn't freeze applications.

### Output of `rpm-ostree status`

```shell
State: idle
Deployments:
● ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-deck-gnome:stable
                   Digest: sha256:de25f06ba3cdb4454356ab4ecdffb3253c712a8b619633dfcea41e138aa27ba5
                  Version: 42.20250519.2 (2025-05-20T00:21:54Z)

  ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-deck-gnome:stable
                   Digest: sha256:ccb44a93ba1afff9667e934962e383df18ee0fe66921e9c985611627c2673aae
                  Version: 42.20250519 (2025-05-19T05:07:46Z)
```

### Hardware

Steam Deck LCD (Jupiter)

### Extra information or context

[Screen capture of the bug](https://drive.proton.me/urls/01GFW07VGR#SzKKJ4M4arRZ)

**Layered packages:**
p7zip-plugins

**Installed extensions:**
Alphabetical App Grid
Clipboard Indicator
Statisig
Mic Indicator Visibility Manager

badnetmask:
That happened to me yesterday, after I upgraded to `42.20250519`. But for me Chrome was freezing every single time I tried to save something (I was trying to download a few files). I didn't test any other apps, simply reverted to `42.20250511`.

scovik:
As a temporary workaround, version 42.20250513 does not have this bug.

In case of Deck:
`rpm-ostree rebase ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-deck-gnome:42.20250513`

For desktop:
`rpm-ostree rebase ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-gnome:42.20250513`

What a strange bug.

SinnedNima:
[...]

PR #2667, opened 2025-05-21
fix: Workaround for gnome file picker

Zeglius:
Solves issue #2658

TODO: Remove this as soon as is fixed upstream


noelmiller (approved):
LGTM

PR #2710, opened 2025-05-31
feat: Open docs in the browser when falling into the apt trap

Zeglius:
"oh man just alias them to open the docs"
~ zachrielforever - 2025


xXJSONDeruloXx (approved):

KyleGospo:
Some applications are stupid and look for the presence of these binaries to figure out what distro they're in. I would turn this into an alias instead of symlinks

PR #2711, opened 2025-06-01
feat: implement ujust picker TUI via lightweight external bin

xXJSONDeruloXx:


xXJSONDeruloXx:
Will bundle bin in containerfile per discord discussions

PR #2727, opened 2025-06-05
fix: update gamescope to fix perf. issues

cyrneko:
update gamescope to commit `918b61` to fix performance degradation due to spammed icon-changes to the host compositor

(feel free to change the commit name, forgot to use the correct semantics on the commit)


antheas:
Create an issue with the commit that fixes it and I will merge it with a co developed by

PR #2744, opened 2025-06-10
feat: Enable IPv6 privacy extensions

shyinsei:
Add this in config to get IPv6 temporary address.

### Why do this?
See https://wiki.archlinux.org/title/IPv6#Privacy_extensions. There's no harm in adding it.


KyleGospo:
Thanks, we'll try this out on testing

PR #2775, opened 2025-06-16
feat: replace topgrade with uupd for system update shortcut

Zeglius:


KyleGospo:
Might be worth a polkit so it just runs

Zeglius:
> Might be worth a polkit so it just runs

No clue how to go on that tbf, though probably this would be better solved in the uupd package

xXJSONDeruloXx:
@Zeglius @KyleGospo  can I propose that going forward, any PR that touches the updater be merged into unstable instead of testing? in multiple cases we have gotten users update stuck in gamescope and its a PITA to tell them to go to desktop. keeping handhelds in game mode as much as possible is sacred IMO

Zeglius:
> @Zeglius @KyleGospo  can I propose that going forward, any PR that touches the updater be merged into unstable instead of testing? in multiple cases we have gotten users update stuck in gamescope and its a PITA to tell them to go to desktop. keeping handhelds in game mode as much as possible is sacred IMO

Right so there might be a problem there. Unstable is used for very edge cases, and barely receives changes. So if we were to push changes to the updater to unstable, we would need dummy commits to trigger new builds and test the updater.

And even still we would need to pad between these commits enough time so someone has time to pull the updater commit then test the updater with the dummy one

KyleGospo:
I'll bring unstable up-to-date with testing and change this pr to point there for an initial test

antheas:
> Might be worth a polkit so it just runs

This needs a polkit before it merges

Zeglius:
> > Might be worth a polkit so it just runs
> 
> This needs a polkit before it merges

If anything needs polkit is upstream bootc. I think it's the main reason why uupd needs root permissions (?)

antheas:
uupd updates a lot of stuff and it probably requires root for that too. I would not assume it is only bootc

PR #2796, opened 2025-06-22
feat(ujust): add 'setup-sunshine toggle-inhibit' recipe

Zeglius:
Prevents the system from idling when sunshine is enabled

### DIFF
--- a/system_files/desktop/shared/usr/lib/systemd/user/sunshine-inhibit.service
+++ b/system_files/desktop/shared/usr/lib/systemd/user/sunshine-inhibit.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Prevent sleep when streaming with Sunshine
+After=sunshine.service
+BindsTo=sunshine.service
+PartOf=sunshine.service
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/systemd-inhibit --what=idle:sleep --who=SunshineInhibit --why="Prevent sleep when streaming with Sunshine" --mode=block /usr/bin/sleep infinity
+Restart=on-failure
+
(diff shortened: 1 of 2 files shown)


KyleGospo (approved):

fayeinmay:
I tried this out locally, because when I'm idling in sunshine, my screens turn off, and somehow it makes my second screen the primary screen (different issue)

But what this actually does is prevent sleep entirely when sunshine is even running, which it always is, as it is started on system boot.
It would make sense to only inhibit when sunshine is actively streaming.

I mean it is not misleading as it clearly states "Prevents the system from idling when sunshine is enabled"
But I don't think this improves anything. 

I think a better way would be to look at sunshine ports (UDP 47998 - 48000) and start inhibiting based on these.
Smth like starting a bash file that does smth like this:

```
while true; do
  if ss -tunap | grep -q ':47998'; then
    systemd-inhibit --what=idle:sleep --who=SunshineInhibit --why="sunshine stream" --mode=block /usr/bin/sleep infinity &
    PID=$!
    while ss -tunap | grep -q ':47998'; do sleep 5; done
    kill "$PID"
  fi
  sleep 2
done
```

antheas (changes requested):
No toggles please! It's either enable or disable

ykshek:
would be useful to bring back the `ujust setup-sunshine` to point towards the brew sunshine once it is removed from stable, or from bazaar(not sure if the flatpak or the brew is preferable here)

Currently the brew sunshine startup service seems to have some issues that can be fixed by overriding it with the correct options in the service file(namely starting after graphical.target), though that is probably better done upstream.

PR #2786, opened 2025-06-19
feat(nvidia): add CachyOS script to switch to latest dlss preset

NyxDaylight:
Adds CachyOS scripts ([1](https://github.com/CachyOS/CachyOS-Settings/blob/c60a32513719d8edc32182b739e501a9350c00e0/usr/bin/dlss-swapper)/[2](https://github.com/CachyOS/CachyOS-Settings/blob/c60a32513719d8edc32182b739e501a9350c00e0/usr/bin/dlss-swapper-dll)) to condense the launch command needed to update DLSS in proton games.

**from this:**
`PROTON_ENABLE_NGX_UPDATER=1 DXVK_NVAPI_DRS_SETTINGS=NGX_DLSS_SR_OVERRIDE=on,NGX_DLSS_RR_OVERRIDE=on,NGX_DLSS_FG_OVERRIDE=on,NGX_DLSS_SR_OVERRIDE_RENDER_PRESET_SELECTION=render_preset_latest,NGX_DLSS_RR_OVERRIDE_RENDER_PRESET_SELECTION=render_preset_latest %command%`

**to this:**
`dlss-swapper %command%`

------------------------------------------

`dlss-swapper %command%` - This forces Nvidia DLSS to use the latest preset for SR, RR and framegen + updates the dlss dlls via ngx
`dlss-swapper-dll %command%` - This forces Nvidia DLSS to use the latest preset for SR, RR and framegen + skips ngx updater

------------------------------------------
Please let me know if this script is in the wrong place or if I did not properly give credit.

PR #2803, opened 2025-06-23
feat: steam launch option shortcuts

xXJSONDeruloXx:
WIP

Credit to https://github.com/ublue-os/bazzite/pull/2786 for the idea


SimpleHeuristics:
@xXJSONDeruloXx Putting the " %command%" in the clipboard in Gamescope session persists when switching to desktop mode. Minor annoyance to have the clipboard showing up in the taskbar every time on the desktop. Is there a way to remove it from the clipboard when changing sessions?

Also maybe include an sd1? There are some games like God of War Ragnarok which will not launch without SteamDeck=1

PR #2841, opened 2025-07-04
fix: Standardize os-release

KyleGospo:
Move Image Flavor and Fedora Version to the VERSION_CODENAME field, standardize on "Bazzite (FROM Fedora Atomic)" for the pretty name.

IE:
PRETTY_NAME="Bazzite (FROM Fedora Atomic)"
VERSION_CODENAME="Silverblue 42"


EyeCantCU (approved):

antheas (changes requested):
As per @castrojo , this will hurt us if we do it now. Let's do it with F43.

KyleGospo:
> As per @castrojo , this will hurt us if we do it now. Let's do it with F43.

In that case I will draft it and merge then

nikodunk:
May I pitch merging this earlier than F43:

- F43 target release date is Nov 11 2025
- IF that’s hit by Fedora, then a few days/weeks after that (end of nov?) bazzite would release 43
- Steam cutoff is Dec 1 for survey, making December the “cutover month” when everyone’s systems update, for Jan 1 survey being a little haywire.
- Feb 1 2026 survey would then be correct

That means we’d lose out on aug, sept, oct, nov, dec and potentially Jan 1 steam hardware surveys where they’d undercount us. 

=> May be better to simply do cutover now, and then within a month everyone would be reporting the newest string properly. I understand being conservative as @castrojo  and @antheas  proposed, but looking at the timeline I thought I’d pitch re-evaluating real quick.

It’s not user-facing and I don’t believe there’s breakage risk, so may be sensible to be more aggressive for the added exposure for the next 6 months?

KyleGospo:
I'm open to that if @antheas is, we're gambling disappearing from the survey but if it's less damaging now versus later that's not a bad thing.

antheas:
The problem is not everyone will update... So we will be undercounted either way

p5:
> The problem is not everyone will update... So we will be undercounted either way

In that case, wouldn't it be better to make the switch sooner rather than later, so devices will be showing on the survey as the "final" configuration faster?

antheas:
All of my devices have gotten the hardware survey this month. Lets rediscuss for the update that will drop closer to August 1st.

nikodunk:
> All of my devices have gotten the hardware survey this month. Lets rediscuss for the update that will drop closer to August 1st.

Smart! If you were to merge towards end of July, and most of the systems update to the new PRETTY_NAME end of July/early August, then the hit should be minimal. Most people who get the survey in August would already be on the new system.

PR #2867, opened 2025-07-09
fix(kde): set bazzite wallpaper on SDDM

renner0e:
This replaces the symlinks for the default Fedora wallpapers.
Fedora's SDDM config uses `/usr/share/backgrounds/default.jxl`,
they changed this to jxl with F42.

gnome-deck also uses SDDM so I did the same thing there as well

image

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -717,7 +717,10 @@ RUN --mount=type=cache,dst=/var/cache \
             gnome-shell-extension-caribou-blocker \
             sddm && \
         dnf5 -y remove \
-            malcontent-control \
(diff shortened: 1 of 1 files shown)


KyleGospo:
I like this, but I don't like the symlink lying about the file format. How about I create a jxl for this and upload?

KyleGospo:
See: https://github.com/ublue-os/bazzite/commit/be12c8b69507e9f828a3431581fb6e56575ac6e7

renner0e:
fine by me

PR #2870, opened 2025-07-10
feat(bazaar): add LACT to utilities

NyxDaylight:
Adds LACT to utilities section of bazaar

------------------------------------------

Wasn't sure if LACT was omitted because of the ujust command but figured I would raise this in case it was. Will close if that was indeed the case.

If LACT isn't added because of what I mentioned above maybe it makes sense to investigate if bazaar could include a way to add pre/post install actions where we can emulate the ujust command.


KyleGospo:
This was intentionally omitted because we have a ujust for it that handles the necessary kernel argument.

NyxDaylight:
> This was intentionally omitted because we have a ujust for it that handles the necessary kernel argument.

Makes sense, I may raise a feat request for bazaar about the pre/post action i mentioned above. Closing for now.

PR #2872, opened 2025-07-10
chore(sunshine): fix license version

ReenigneArcher:
This PR fixes the license for Sunshine, which uses GPLv3 not GPLv2. https://github.com/LizardByte/Sunshine?tab=GPL-3.0-1-ov-file#readme


KyleGospo:
Just to be clear this is a stub package that exists only to ensure Sunshine is still using this old name. This avoids a issue with updating from very old builds.

Regardless I see no reason why the license can't match, thanks!

PR #2874, opened 2025-07-10
feat: Add gameclass tool for arbitrary game icons

EPOCHvoyager:
### Background

This tool is a hacky workaround for this issue, particularly plaguing GNOME:

Image showing a missing window icon

Because GNOME has deprecated and removed icon hints, to my knowledge, all windows missing a matching `StartupWMClass` in a `.desktop` file will use this default icon and a generic app name. 

While this is also the case with the shortcuts Steam creates by default (during installation or by right-clicking the game > Manage > Add desktop shortcut), it can be manually fixed. For native games, you can add a `StartupWMClass=<executable-name>` line to a `.desktop`. For Proton games, you can add `StartupWMClass=steam_app_<appid>`, due to Proton automatically setting this `WM_CLASS` with the game's native Steam `appid`.

The problem arises on non-Steam Windows games, installed through Lutris or Heroic. As they're not Steam apps, they do not possess an `appid` for Proton to source, so they'll usually have a window with a `steam_app_default` or `steam_app_0` generic `WM_CLASS`. Because they're not individualized, from what I can gather, it's impossible to have a matching `.desktop` for more than one game — meaning no game window icons at all on your desktop, at least under GNOME.

### The "Solution"

To work around that, this tool queries the window list every second using `wmctrl`, reads the `cmdline` for a given window PID, and then applies an arbitrary `WM_CLASS` using `xdotool`, if the `cmdline` has a match for a given string. I have no idea if this is a good way of doing this, but it seems to work — at least for X11 applications.

Do note that I have no experience writing tools like this with Bash, so this may be naive or poorly written. If this isn't fit for Bazzite due to that, or unfit in general, I'll happily modify this PR or retract it altogether.

## Example Usage

Image showing an example .desktop file, with the StartupWMClass property highlighted
Image showing the setup for gameclass under Lutris
Image showing the setup for gameclass under Heroic

## Result

Image showing Thief Gold with a matching window icon
Image showing Thief Gold with a matching icon and app name in the GNOME Dash

### DIFF
--- a/system_files/desktop/shared/usr/bin/gameclass
+++ b/system_files/desktop/shared/usr/bin/gameclass
@@ -1,11 +1,11 @@
 #!/bin/bash
 
 if (( $# == 0 )); then
-	echo -e "No arguments provided.\n\nUsage: env GAMECLASS='<executable>=<window class>' gameclass [COMMAND]..."
+	echo -e "No arguments provided.\n\nUsage: env GAMECLASS='<executable>=<window class>' gameclass \033[1m[COMMAND]\033[0m..."
 	exit 1
 else
 	if [[ -z ${GAMECLASS} ]]; then
-		echo -e "No environment variable set.\n\nUsage: env GAMECLASS='<executable>=<window class>' gameclass [COMMAND]..."
+		echo -e "No environment variable set.\n\nUsage: env \033[1mGAMECLASS='<executable>=<window class>'\033[0m gameclass [COMMAND]..."
 		exit 1
 	fi
 fi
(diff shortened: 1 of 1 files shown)


EPOCHvoyager:
Added an image with an example `.desktop` open in Text Editor for easier comprehension.

KyleGospo (approved):

EPOCHvoyager:
Sorry for the mess!

system_files/desktop/shared/usr/bin/gameclass
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+
+if [[ -z $( echo "$@" ) ]]; then
```
Zeglius, line 3:
```suggestion
if [[ $# -eq 0 ]]; then
```

KyleGospo (approved):

PR #2883, opened 2025-07-12
feat: Expose more HDMI-CEC options

8bitDesigner:
This PR exposes some of the options that are available to `cec-client` via the config file in `/etc/default/cec-control` so that you can control more aspects of your CEC settings without needing to fork the project and maintain your own image.

Namely, I'm doing the following:

- Exposing a new option, `CEC_OSD_NAME`. This controls what your TV will display when following commands from your device. In my case, this defaults to `CECTester` for my PulseEight adaptor module. By default, this new setting will use the system's hostname, falling back to the word "bazzite".

  This is nifty as now, when navigating around your TV settings, "bazzite" will be shown alongside the input bound to your HTPC.

- Exposing a new option, `CEC_PORT_NUMBER`. This setting tells `cec-client` to override the learned HMDI port number. This is critical for the `as` command as it's telling your TV which port to activate. Sometimes your CEC adaptor will get this wrong (looking at you, PulseEight module), and the `as` command will activate the wrong source. If this value is empty, it will be omitted from the `cec-client` arguments list.

- Allow `CEC_TVID` to accept multiple space-separated values. If you have a complex AV system, you may have multiple devices you need to `on` or `standby` in concert, like a separate audio system or tuner.

Lastly, this includes a couple of quality of life improvements:

- The arguments list for `cec-client` has been swapped for a Bash array using long-form argument names so that anyone reading the source doesn't need to have a copy of `cec-client` handy to understand what the `-s` and `-d 1` arguments do.

- The commands to issue wake and sleep commands have been moved to Bash functions to keep the chained set of conditionals at the end of the file easy to understand.

### DIFF
--- a/system_files/desktop/shared/usr/bin/cec-control
+++ b/system_files/desktop/shared/usr/bin/cec-control
@@ -9,15 +9,40 @@ CEC_WAKE=${CEC_WAKE:-true}
 CEC_SETSOURCE=${CEC_SETSOURCE:-true}
 CEC_ONPOWEROFF_STANDBY=${CEC_ONPOWEROFF_STANDBY:-true}
 CEC_ONSLEEP_STANDBY=${CEC_ONSLEEP_STANDBY:-false}
+CEC_OSD_NAME=${CEC_OSD_NAME:-$(hostname)}
+
+args=()
+args+=( "--single-command" )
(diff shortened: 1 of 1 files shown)


KyleGospo:
This is great, thanks!

PR #2901, opened 2025-07-15
fix: force-wol script can handle multiple network interfaces

chli1:
The current script fails if there are multiple connected networks because $INTERFACE contains a string with all connected networks. Therefore I've changed the script to iterate all detected interfaces and enable wol for each of them.


KyleGospo:
Thank you!

PR #2938, opened 2025-07-18
feat(just): Add setup-realtime

EPOCHvoyager:
### Background

Configures the system for realtime task scheduling, sacrificing some performance for consistent latencies — useful primarily for audio engineering and production applications on the desktop. This should allow audio processing to be scheduled with very tight latencies — in the single digit millisecond range — without crackling or other forms of distortion (known in the Linux audio space as Xruns).

In the aforementioned use cases, even just a dozen milliseconds could disrupt the user in playing a MIDI instrument for sequencing, monitoring a microphone's output through headphones, or simply playing back complex audio projects.

### Implementation

Fedora maintains a [realtime-setup package](https://src.fedoraproject.org/rpms/realtime-setup), layered by this script with `--apply-live` to enable its included systemd services. The udev rules added for threadirqs and `/dev/rtc` access are given to the `realtime` group, which currently needs to be [explicitly added](https://docs.fedoraproject.org/en-US/fedora-silverblue/troubleshooting/#_unable_to_add_user_to_group) to `/etc/groups` in Atomic Fedora.

As PREEMPT_RT has been in the mainline Linux kernel since 6.12, the only change required to equate modern kernels to the former RT kernel builds is the inclusion of the `threadirqs` kernel argument; given [Bazzite enables full preemption by default](https://github.com/ublue-os/bazzite/blob/6ee1967840f2707a14c03a26412181d46eaa6d78/system_files/desktop/shared/usr/libexec/bazzite-hardware-setup#L176).

### Development Note (Work in Progress)

This is currently in a proof of concept stage. I still plan on adding an installation prompt along with a reversal script.

**Supersedes #1726.**

### DIFF
--- a/system_files/desktop/shared/usr/share/ublue-os/just/83-bazzite-audio.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/83-bazzite-audio.just
@@ -303,7 +303,7 @@ setup-realtime:
       rpm-ostree install realtime-setup --apply-live
       sudo systemctl enable realtime-entsk
       sudo systemctl enable realtime-setup
-      if ! grep -q realtime /etc/group; then
+      if ! [[ $( cat /etc/group ) =~ "realtime" ]]; then
         grep -E '^realtime:' /usr/lib/group | sudo tee -a /etc/group
       fi
       sudo usermod -aG realtime "$USER"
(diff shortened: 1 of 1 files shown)


xXJSONDeruloXx:
Can you make this a toggle so user can easily undo the changes? Check out other ujust recipes that use `ugum` and our styling.

EPOCHvoyager:
> Can you make this a toggle so user can easily undo the changes? Check out other ujust recipes that use `ugum` and our styling.

For sure, that's what I had planned.

EPOCHvoyager:
Just an update — currently waiting on some patches from [this development branch](https://git.kernel.org/pub/scm/linux/kernel/git/arighi/linux.git) to hit mainline; right now it's not necessarily recommended to run strenuous real-time tasks with sched_ext, which is unfortunate, because the scx schedulers can really help with consistent low latency. Also dog-fooding this, plus some extra kernel arguments on my end, to see if there are any serious regressions.

PR #2954, opened 2025-07-23
feat: Oxidize

KyleGospo:
This commit swaps us to sudo-rs, following Ubuntu's lead. This is something I've been testing a long while now with no regressions. This only covers sudo as the coreutils swap is not yet ready for mainstream use.

See: https://thehackernews.com/2025/07/critical-sudo-vulnerabilities-let-local.html for reasoning as to why we might want to do this ahead of our upstream.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -118,6 +118,17 @@ RUN --mount=type=cache,dst=/var/cache \
     dnf5 -y config-manager setopt "*staging*".exclude="scx-scheds kf6-* mesa* mutter* rpm-ostree* systemd* gnome-shell gnome-settings-daemon gnome-control-center gnome-software libadwaita tuned*" && \
     /ctx/cleanup
 
+# Oxidize
+RUN --mount=type=cache,dst=/var/cache \
+    --mount=type=cache,dst=/var/log \
+    --mount=type=bind,from=ctx,source=/,target=/ctx \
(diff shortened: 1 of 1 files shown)


KyleGospo:
If we are to do this, I'd recommend it lives in testing exclusively for a multi-month period, IE not reaching main until at least F43. This PR covers only trying it in testing in the meantime.

antheas:
But neither vulnerability is related to a memory exploit?

Zeglius:
Honestly? No benefit in this for the users. I would rather wait upstream to be the ones pushing for this kind of changes.

KyleGospo:
> Honestly? No benefit in this for the users. I would rather wait upstream to be the ones pushing for this kind of changes.

This has a security benefit for users and is free work being paid for by Canonical and supported in one of the biggest Linux distros. I see this as a chance to contribute to Fedora by trail blazing good ideas and utilizing resources far beyond our own.

bigpod98:
From my perpective while yes non of the recent security issues was memory related switching to rust version has variaty of security benefits and drawbacks as so does remaining with sudo. 

Also main maintainer of sudo is in high support od sudo-rs.

Venefilyn:
Just a note here. Cockpit will not work with sudo-rs due to askpass not being implemented

- https://github.com/trifectatechfoundation/sudo-rs/issues/1249

Issue #2081, opened 2025-01-06
OneXPlayer1 AMD's screen doesn't wake up after suspend


kalzEOS:
### Describe the bug

The screen stays black, I can still hear the UI sounds when I move around and click on things but the screen is off. When I press the power button on the back once or twice, it takes a second to wake up then starts flashing dimming and brightening then shuts off, sometimes it works, others it goes back to black and I'd need to hold down the power button to turn it off and on

### What did you expect to happen?

Normal screen after suspend

### Output of `rpm-ostree status`

```shell
State: idle
Deployments:
● ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-deck:stable
                   Digest: sha256:011580d0843e0c6ffbd23ec56eadcfb6bf564020b9b28f2aacf612e796ea178b
                  Version: 41.20250106.1 (2025-01-06T14:19:15Z)

  ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-deck:stable
                   Digest: sha256:1c4c029309a887eb561e346ed469b72ed75c59451bd9eb0a873b0b507ec71f98
                  Version: 41.20241230.1 (2024-12-30T06:48:54Z)
```

### Hardware

Operating System: Bazzite 41
KDE Plasma Version: 6.2.5
KDE Frameworks Version: 6.9.0
Qt Version: 6.8.1
Kernel Version: 6.12.8-201.bazzite.fc41.x86_64 (64-bit)
Graphics Platform: Wayland
Processors: 16 × AMD Ryzen 7 5700U with Radeon Graphics
Memory: 12.5 GiB of RAM
Graphics Processor: AMD Radeon Graphics
Manufacturer: ONE-NETBOOK TECHNOLOGY CO., LTD.
Product Name: ONE XPLAYER

aurelios69:
I'm having the exact same issue. With the same Onexplayer AMD

KarkanAlzwayed:
> I'm having the exact same issue. With the same Onexplayer AMD

I have actually installed windows on it for now. It was just unusable at all.

aurelios69:
Too bad. I just use bazzite on my legion go and ally z1e. I'm also installing windows. Hopefully this will be fixed. This doesn't happen with nobara, just with bazzite.

KarkanAlzwayed:
> Too bad. I just use bazzite on my legion go and ally z1e. I'm also installing windows. Hopefully this will be fixed. This doesn't happen with nobara, just with bazzite.

[...]

Issue #1984, opened 2024-12-15
Ayaneo Next Lite 4800u after waking from sleep WIFI no longer works or connects until a reboot


mivlives4ever:
### Describe the bug

Ayaneo Next Lite 4800u after waking from sleep WIFI no longer works or connects until a reboot. WIFI shows as being turned on, and even after toggling the WIFI button, it doesn't not activate

### What did you expect to happen?

After waking from sleep, WIFI should be able to function and connect to WIFI access points

### Output of `rpm-ostree status`

State: idle
Deployments:
● ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-deck-gnome:stable
                   Digest: sha256:6dc8c4c0779253561b2c57b28fd71f6efa08a09c64e06f8fac5456fbd71d81a5
                  Version: 41.20241211 (2024-12-11T07:30:41Z)

  ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-deck-gnome:stable
                   Digest: sha256:a1e956b154de68c2c15909630c8209d7255e583d4c32e36618ec8b0cdf5b9017
                  Version: 41.20241210 (2024-12-10T08:52:57Z)

### Hardware

NEXT Lite

aarron-lee:
did you already try the wifi workaround, as described in the bazzite ayaneo docs?

https://docs.bazzite.gg/Handheld_and_HTPC_edition/Handheld_Wiki/Ayaneo_Handhelds/?h=ayaneo#workarounds-known-issues

aarron-lee:
You shouldn't need to modify the bios on the 4800u, since it supports S3 sleep.

The wifi workaround should be all you need, but perhaps your device has different wifi hardware. that'd require modifying the wifi workaround accordingly.

mivlives4ever:
> You shouldn't need to modify the bios on the 4800u, since it supports S3 sleep.
> 
> The wifi workaround should be all you need, but perhaps your device has different wifi hardware. that'd require modifying the wifi workaround accordingly.

The wifi card in the Next Lite is an Intel AX200...

aarron-lee:
> The wifi card in the Next Lite is an Intel AX200...

Hrmm, not sure if this is related, but when I had an Ayaneo Air Pro 5825u the wifi stopped working on suspend-resume after swapping in an Intel AX200 card. I had to revert to the wifi card that shipped with the device.

mivlives4ever:
[...]

PR #2927, opened 2025-07-17
feat(ujust): add 'fix-ax1775-sleep' recipe 

MarcosVLl2:
Added fix for issues with Intel AX1775 / BE200 Wi-Fi 7 chips entering D3sleep mode and not booting back up after sleep.

PR #2974, opened 2025-07-26
fix(ayaneo): workaround for buggy wifi/sleep interaction on first-gen Ayaneo Air

### LINKED ISSUE #2134: Black screen on Ayaneo Air 30% of the time
### Describe the bug

Hi, i installed the latest version of Bazzite on my Ayaneo Air a few days ago and i'm loving it. There is only one issue: The screen is black sometimes. It can happen immediately after a reboot, when switching between desktop/game mode or resuming from sleep. It looks like sddm isn't initializing properly because Steam makes sounds if i press buttons, and issuing `systemctl restart sddm` over ssh usually fixes the problem, but requires multiple tries at random.

The display should always come on when booting or resuming from sleep.

### Output of `rpm-ostree status`

```shell
State: idle
Deployments:
● ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-deck:stable
                   Digest: sha256:702c8397e1215800c5a589728ddc4e30822a07328f004dff3ead71b5960e400f
                  Version: 41.20250106.3 (2025-01-07T04:25:40Z)
          LayeredPackages: glances htop syncthing
```

### Hardware

AIR

https://paste.centos.org/view/0d709dc3

DCKcode:
First-gen Ayaneo Air devices appear to have a bug where their BIOS doesn't interact well with the Wi-Fi chip around sleep and wake. Sleeping and waking the device will cause error messages in dmesg and unreliable and buggy wakeup, and after wakeup the wifi interface will appear to have dissappeared.

The workaround for this issue is to remove the Wi-Fi driver's kernel module just before going to sleep and readding it when waking.This PR adds this workaround and enables it automatically on these affected devices. A version of this workaround [exists in ChimeraOS as well](https://github.com/ChimeraOS/device-quirks/blob/ee8e66690eb460ec8b13b56eabb7388298e0972f/etc/device-quirks/systemd-suspend-mods.conf#L3).

I tested this on a Ayaneo Air Pro (5560U model) and this enables sleep to work reliably! 

Fixes #2134

### DIFF
--- a/system_files/desktop/shared/usr/lib/systemd/system/kmod-sleep-workaround@.service
+++ b/system_files/desktop/shared/usr/lib/systemd/system/kmod-sleep-workaround@.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Unloads/reloads %i kernel module before/after sleep for devices where it cause sleep/wake issues
+Before=sleep.target
+StopWhenUnneeded=yes
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=-/usr/bin/rmmod %i
+ExecStop=-/usr/bin/modprobe %i
+
+[Install]
+WantedBy=sleep.target
--- a/system_files/desktop/shared/usr/libexec/bazzite-hardware-setup
+++ b/system_files/desktop/shared/usr/libexec/bazzite-hardware-setup
@@ -218,6 +218,11 @@ if [[ $IMAGE_NAME =~ "deck" || $IMAGE_NAME =~ "ally" ]]; then
   systemctl enable --now ds-inhibit.service
 fi
 
+# Enable workaround for buggy wifi sleep interaction on first gen Ayaneo Air devices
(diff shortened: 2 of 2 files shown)


DCKcode:
The new `kmod-sleep-workaround@` service introduced here could be used for other buggy driver<>sleep interactions in the future. For example, it could be used #2927 and potentially to fix #1984, but those would require testing with those particular devices (I don't have access to those).

DCKcode:
This PR now also potentially fixes #2081 pending someone that confirm this on a first gen OneXPlayer device.

xXJSONDeruloXx (approved):

DCKcode:
Reverted the changes for enabling this fix on the first gen OneXPlayer due to issues encountered in testing described in #2081 . I'm not sure how those issues can be explained with the changes in this PR given applying the workaround manually does work, but this could be something we revisit in the future.

DCKcode:
@xXJSONDeruloXx @antheas this should be ready to merge now, let me know if you're good with this.

PR #3024, opened 2025-08-04
feat: Restore evdi as an optional feature

KyleGospo:


antheas (approved):

fiftydinar:
Maybe add toggle for disable/enable, in case users want to revert to defaults?

KyleGospo:
> Maybe add toggle for disable/enable, in case users want to revert to defaults?

Disabled by default, only enabled via ujust. Same as before.

PR #3025, opened 2025-08-04
feat: Gigabyte motherboard wake hw fix, ujust for manual overrides

xXJSONDeruloXx:
for HTPCs and desktops with gigabyte motherboards, on linux there is an issue causing them to immediately wake from suspend, which can be fixed by disabling GPP0 ACPI wakeup source

This PR adds an automatic check and fix if a gigabyte mobo is detected, and for other mobo brands that may be similarly affected, add a ujust with ugum choose menu to enable and disable the service manually.


tired-runner:
Can confirm this udev rule fixes the wake from suspend issue with gigabyte b550 motherboards. I've been running with this rule for a few months on a B550 Aorus Pro AC with no apparent issues.

HikariKnight (approved):

PR #3026, opened 2025-08-04
feat: Add framework_tool binary from Framework Computer for handling various hardware features

KyleGospo:


rugk:
Hmm, why is this only in Bazzite? It's not really gaming-related, so maybe it would be cool to include it [in the base ublue](https://github.com/ublue-os/main)? Opened https://github.com/ublue-os/main/issues/1304 about that.

PR #3031, opened 2025-08-06
chore(ujust): refactor setup-decky script

Lumaeris:
This script is now using ugum for option selection, since on older setup-decky while choosing prerelease version it will grab a release version, and when uninstalling Decky it will redownload Decky for no particular reason and then uninstall it.

Also I've changed the logic of checking for Decky installation, since removing it doesn't mean that it will remove `homebrew/plugins` directory, so now setup-decky will explicitly check for systemd service that Decky enables while installing it, so it will no longer give a false sense that Decky is still enabled.


xXJSONDeruloXx:
Already are using ugum. The existence of the default install action is the cause of the behavior you mentioned. I like the change to monitoring the service instead of the plugin folder, but there's no need for a full rewrite imo

KyleGospo (approved):

PR #3046, opened 2025-08-10
Removed the removal of Konsole

dnkmmr69420:
I added Konsole back with a simple code change. I do not see much of a reason to remove Konsole (unlike kde partiton manager)


KyleGospo:
We're sticking with Ptyxis and will not support Konsole in any capacity. Create a custom image or use the flatpak if you want it.

PR #3057, opened 2025-08-12
fix(liveiso): Add dosfstools for anaconda

Zeglius:
This should fix anaconda being unable to select custom EFI partitions


KyleGospo:
Any reason this is draft? Happy to merge and build Live ISOs

Zeglius:
> Any reason this is draft? Happy to merge and build Live ISOs

Basically I doubt (and still do) that this would fix the issue. Seems more like anaconda GTK being silly, rather than missing dependencies.

Regardless, seems anaconda webui won't have this issue, so all we have to do is wait for F43

Issue #2044, opened 2025-01-02
Sunshine service failing to autostart on Bazzite gnome desktop images


wolfyreload:
### Describe the bug

It looks like there is a problem with sunshine failing to autostart on the desktop images of Bazzite. tested on bazzite-gnome-nvidia:testing-41.20250101 and bazzite-gnome:testing-41.20250101

I get the following errors in the log

```
sunshine[6282]: Error: Client failure: Daemon connection failed
sunshine[6282]: Error reading events from display: Broken pipe
systemd[1624]: sunshine.service: Main process exited, code=exited, status=1/FAILURE
systemd[1624]: sunshine.service: Failed with result 'exit-code'.
```

If I manually start the service with `systemctl --user start sunshine` is starts without issue

Also tested on "bazzite-deck-gnome:testing-41.20250101" and the service starts without issue so it seems to be specific to the Gnome desktop images. 

As a work around, I've added Sunshine to Tweaks >> Startup Applications

### What did you expect to happen?

I would expect the sunshine service to start up on bazzite-gnome-nvidia and bazzite-gnome like it does on bazzite-gnome-deck

### Output of `rpm-ostree status`

```shell
● ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-gnome-nvidia:testing-41.20250101
                   Digest: sha256:d327c14bfb9ed80279ef456a2aeee3b0b23a87167a7f20dc0faf5a23d82fac1f
                  Version: testing-41.20250101 (2025-01-01T10:37:35Z)
                     Diff: 2 upgraded, 76 removed, 37 added
          LayeredPackages: code cosmic-store nemo qdirstat

ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-deck-gnome:testing-41.20250101
                   Digest: sha256:e2a26ab8246441a3a44b9dfbf2f59708120605059be745f1d6af665580de6be4
                  Version: testing-41.20250101 (2025-01-01T10:38:30Z)
          LayeredPackages: code cosmic-store nemo qdirstat

  ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-gnome:testing-41.20250101
                   Digest: sha256:590b5e06d4875b3b99e7b20c642441b1e393af97b63c00fca17c97c34cac2c42
                  Version: testing-41.20250101 (2025-01-01T10:43:02Z)
          LayeredPackages: code cosmic-store nemo qdirstat
```

### Hardware

[...]

PR #3061, opened 2025-08-14
Fix(gnome): fix `ujust setup-sunshine enable` for gnome desktop

johnelliott:
# Summary
I was trying bazzite-gnome and noticed the sunshine enable just script didn't seem to work. 

What I came to was that gnome doesn't trigger the systemd target that sunshine's service ships with (xdg-autostart-desktop). 

As others noted on #2044 a work-around is using Ignition, but that requires GUI access and for the user to figure out that the autostart thing is called Ignition and isn't in gnome settings etc. AND realize the just script hasn't worked.

## Solution
This PR basically creates a unit override for the sunshine user service to specify that when its installed it should also be started when gnome-session.target is reached. This is done by the WantedBy systemd directive. I'm no expert but this seemed like the least-messy solution.

## Alternatives
This is my first contribution to ublue, so I wasn't quite sure how to determine whether it was the gnome image or not (looked like maybe conditionals in containerfile?!) ~~so I just looked up that it's legal to have 1+ WantedBy directives.~~ 

## How to test ujust and the override
- use the gnome desktop image
- copy the override file to your `~/.config/systemd/user/sunshine.service.d/`
- run `ujust setup-sunshine disable` and see that it cleans up `~/.config/systemd/user/` of the systemd wants files for xdg-desktop-autostart
- run `ujust setup-sunshine enable` and see that it creates a directory structure like this:

```
.
├── gnome-session.target.wants
│   └── sunshine.service -> /usr/lib/systemd/user/sunshine.service
└── sunshine.service.d     <-- This is only here for testing
    └── override.conf      <-- This is only here for testing

```
- reboot your gnome desktop image'd machine and see that sunshine starts

Update: we moved the file to the silverblue directory for this gnome-only fix since KDE works fine.

### DIFF
--- a/system_files/desktop/silverblue/etc/systemd/user/sunshine.service.d/override.conf
+++ b/system_files/desktop/silverblue/etc/systemd/user/sunshine.service.d/override.conf
@@ -1,0 +1,2 @@
+[Install]
+WantedBy=gnome-session.target


KyleGospo:
Just wanted to say this is an absolutely excellent first pr, well done

KyleGospo:
If this fix is GNOME exclusive I'd just move it to the silverblue directory instead of shared

johnelliott:
> If this fix is GNOME exclusive I'd just move it to the silverblue directory instead of shared

OK great, I didn't know silverblue was gnome, so it all makes more sense. I may just replace the xdg-desktop-autostart WantedBy because that target isn't reached in gnome.

johnelliott:
Updated changes:
1. Moved the override path to silverblue
2. Removed the WantedBy directive that KDE plasma uses since it's not used in gnome

johnelliott:
Sorry for the CI noise, I was getting commit signing configured for this project.

johnelliott:
IDK how important the build successes are for this, but I notice the one I'd use on my main machine [succeeded](https://github.com/ublue-os/bazzite/actions/runs/17000014112/job/48204407882?pr=3061). It looked like some others failed on the GPU install shell script the workflow grabs down from another ublue repo.

PR #3076, opened 2025-08-17
fix(brh): fix rebasing from one signed image to another

Lumaeris:
Some users can't rebase from one signed image to another lately, this simple fix resolves the problem with rollback helper.


johnelliott:
I had this issue today and I think this fix would have prevented my specific situation of rebasing from my signed stable to signed testing that I had to run with rpm-ostree after brh failed. `docker://` was all I needed to add to my ostree command.

PR #3084, opened 2025-08-18
feat: Remove Toolbox

KyleGospo:
In the spirit of removing upstream tools without use, this removes toolbox. Previously we kept this despite Distrobox mogging it due to its removal not saving space, but rechunker means the 100kb this uses can be reclaimed.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -229,6 +229,7 @@ RUN --mount=type=cache,dst=/var/cache \
         ublue-os-update-services \
         firefox \
         firefox-langpacks \
+        toolbox \
         htop && \
     /ctx/cleanup
 


ivxvm:
@KyleGospo 
Ouch, I ran half my services using toolbox. Because `toolbox enter` was faster than creating fedora distrobox and then entering it with explicit name. Would be nice to have some deprecation warning at least.

HikariKnight:
> @KyleGospo Ouch, I ran half my services using toolbox. Because `toolbox enter` was faster than creating fedora distrobox and then entering it with explicit name. Would be nice to have some deprecation warning at least.

we have been actively discouraging the use of toolbox for at least 2 years

miabbott:
> we have been actively discouraging the use of toolbox for at least 2 years

Another `toolbox` user here...

Admittedly, I'm not very active on the Discourse (and almost never on Discord), but I wasn't able to find any forum posts or GitHub issues where that discouragement was clearly communicated.

I'm not saying it is untrue, but would like to better understand the reasons for removing `toolbox`.

KyleGospo:
If you're an active user of it you can layer it back in, nothing stopping that. 

As Hikari said we have never recommended this tool, we have actually actively discouraged its use for the entire history of this project. 

This change means others can't make the same mistake.

PR #3091, opened 2025-08-20
fix(nvidia): fix typo in icd files

Zeglius:
See
https://forums.developer.nvidia.com/t/580-release-feedback-discussion/341205/174 Revert this when fixed upstream.
Good catch, asheriif.


asheriif:
Looks good thanks for the quick patch!

Issue #1994, opened 2024-12-19
GPD Win Max 2021 G1619-03


xKernelPanicx:
### Describe the bug

Gamescope orientation is not set to right orientation. The device-quirks file does not include the modem G1619-03 in the GPD section for Gamescope Orientation. The section needs to include G1619-03 in order for Gamescope to be usable. 

### What did you expect to happen?

I expected Gamescope to be oriented to landscape, or in "right" orientation. 

### Output of `rpm-ostree status`

```shell
State: idle
Deployments:
● ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-deck:stable
                   Digest: sha256:84712e223557d8eb2c3d0ba7fc39070715f6ba796463326405de4bd7d9116a10
                  Version: 41.20241216 (2024-12-16T05:01:17Z)

  ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-deck:stable
                   Digest: sha256:f44bebbf15932d9edc4e0b28b91614be1e172dba004c0cad6f7a3d46c00837eb
                  Version: 41.20241125 (2024-11-26T03:48:17Z)
```

### Hardware

G1619-03

### Extra information or context

I just need G1619-03 added to the device quirks file in the GPD Section. I'm told that I can't permanently modify the file because that would go against what the OS is designed for.

antheas:
`sudo rpm-ostree kargs --append-if-missing video=eDP-1:panel_orientation=left_side_up` try this, reboot, and if it works i will quirk it in the kernel

ThisIsSet-L:
Works with the following

`sudo rpm-ostree kargs --append-if-missing video=DSI-1:panel_orientation=right_side_up`

antheas:
Resolution?

ThisIsSet-L:
Screen is native 800x1280

ThisIsSet-L:
Issue is still not resolved

antheas:
Did you try the command?

ThisIsSet-L:
@antheas yeah I did. It flipped it to upside down. I have the correct orientation in my first comment on this issue

antheas:
Can you post the DMI information so 8 can add it to the kernel?

ThisIsSet-L:
Will get that to you as soon as I can.

PR #3094, opened 2025-08-21
fix(gpd) fix screen rotation on GPD Win Max 2

ThisIsSet-L:
Fixes screen rotation on GPD Win Max 2 devices


ThisIsSet-L:
#1994

antheas:
What is the resolution of the panel?

antheas:
It is 800p over DSI.

antheas:
This is already in our kernel. It needed the rotation shader in https://github.com/bazzite-org/gamescope-session/commit/c65fbffa7306167989e4dd6fe76d6bab3c9d8c30 . I will build bazzite testing in a couple min

PR #3161, opened 2025-09-05
feat: recommend ONLYOFFICE in the Office & Productivity section of Bazaar

tired-runner:
Only office does a better job than libre office at MS Word compatibility in my experience and it seems to be fairly commonly recommended + not so easy to discover for a new user. Seems like a good candidate to recommend as curated.


HikariKnight:
both libreoffice and onlyoffice have better compatibility in different formats.
Personally i have seen documents have better compatibility in onlyoffice while spreadsheets have better compatibility in libreoffice.

HikariKnight (approved):

ledif:
For context, Aurora added Onlyoffice to its config a couple weeks ago: https://github.com/ublue-os/aurora/pull/856

Issue #3163, opened 2025-09-06
KDE menu entries created via `disrtobox-export` do not work


ja02:
### Describe the bug

1. Use `distrobox-export` with Ubuntu latest to export Midnight Commander:
```bash
    distrobox create -i ubuntu:latest -n ubuntu-distrobox
    distrobox enter ubuntu-distrobox
    distrobox-export --app mc
```
_Two new KDE menu entries will appear under _Utilities_ - one for the `mc` app itself, one for the text editor `mcedit`._

2. Select one of the menu entries.

### Error
The terminal starts with title "/usr/bin/distrobox--nter (Failed)" and prints a single error message
`Failed to find executable /usr/bin/distrobox--nter: No such file or directory`.
Then the internal process in the terminal exits.

### What did you expect to happen?

The selected application starts in a terminal.

### Output of `rpm-ostree status`

```shell
State: idle
Deployments:
● ostree-image-signed:docker://ghcr.io/ublue-os/bazzite:stable
                   Digest: sha256:8b18f3c07669141a6f5190d6a65ef73513f9a8851334e7b88efcf689465ed745
                  Version: 42.20250903 (2025-09-03T21:13:39Z)
            LocalPackages: mullvad-vpn-2025.8-1.x86_64

  ostree-image-signed:docker://ghcr.io/ublue-os/bazzite:stable
                   Digest: sha256:e26f9399bff1aeb8c819e932c3c160cac1a4fe6638c6efc5a8fe4dd2cfb76404
                  Version: 42.20250901.1 (2025-09-01T05:54:43Z)
            LocalPackages: mullvad-vpn-2025.8-1.x86_64
```

### Hardware

󱋩  bazzite:stable 
  Linux 6.16.4-104.bazzite.fc42.x86_64
󰻠  Intel(R) Core(TM) i9-10900 (20) @ 5.20 GHz
󰍛  AMD Radeon RX 470 Graphics [Discrete]
󰍛  Intel UHD Graphics 630 @ 1.20 GHz [Integrated]
  10.59 GiB / 31.16 GiB (34%)
  162.70 GiB / 255.97 GiB (64%) - btrfs [Read-only]
󰍹  2560x1440 @ 75 Hz (as 1706x960) in 31" [External] *
󰍹  2560x1440 @ 75 Hz (as 1706x960) in 31" [External]
󰕮  KDE Plasma 6.4.4
  KWin (Wayland)
  bash 5.2.37
  Ptyxis 48.5

PR #3174, opened 2025-09-07
fix: KDE menu entries created via distrobox-export do not work (#3163)

ja02:
The original code in `system_files/desktop/kinoite/usr/bin/kde-ptyxis` replaced all occurrences of `-e` by `--`, which mangled legit paths, e.g. `/usr/bin/distrobox-enter` became `/usr/bin/distrobox--nter`.
After the fix, only standalone occurrences are replaced and all others are ignored.


KyleGospo (approved):
Thanks!

PR #3182, opened 2025-09-09
feat: win mini speaker ujust

xXJSONDeruloXx:
simple ujust to streamline the audio fix from https://github.com/aarron-lee/gpd-win-tricks/blob/main/win_mini_equalizer/README.md


KyleGospo:
This doesn't need to be a ujust and can be automatically applied like the LGO and Ally.

xXJSONDeruloXx:
@KyleGospo good call, added to desktop and deck since its a whole mini pc for some, but lmk if not needed. also, path set for my 2023 win mini but lmk if theres a better way to expose for other win mini models (2024-2025, 7640u 2023)

KyleGospo:
Only deck has the needed services for this, otherwise this is great! Thanks

xXJSONDeruloXx:
@KyleGospo done!

PR #3187, opened 2025-09-09
docs: Fix formatting issues with ToC and headings across READMEs.

say4n:
These seem to have been introduced with a639f03. I have basically gone ahead and re-indented the ToC as well as made the headings be at the correct level for each of the localized README files.


say4n:
Not too sure if the target branch for this should be `testing` or `main`. Happy to change to either as applicable!

xXJSONDeruloXx (approved):

PR #3194, opened 2025-09-11
feat: pause on suspend and wake toggle ujust

xXJSONDeruloXx:
There has been a long-standing proton bug that causes audio hitching on wake from suspend while a game is running. The "pause games" decky plug-in almost completely solves this issue, but that requires the user to install decky with all of the instability that entails. 

This PR introduces a ujust cmd to toggle on/off that core functionality of pausing a proc just before suspend and unpausing right after wake, which should resolve the audio wake bug as well as a few other quirks, without needing decky


SirFerdek:
What if  we used `systemd-suspend.service` and just muted/umuted all audio channels instead?

PR #3199, opened 2025-09-12
fix(ujust): OpenTabletDriver: Add/remove tablet kernel modules

slaugaus:
Fell down this rabbit hole after opening https://github.com/OpenTabletDriver/OpenTabletDriver/issues/4066.

Per [the Flatpak guide](https://github.com/flathub/net.opentabletdriver.OpenTabletDriver/blob/master/docs/Manual-install.md#fedora-atomic), to ensure all supported tablets work properly, you should blacklist 2 kernel-level tablet modules from modprobe.

I've confirmed that running `rpm-ostree kargs --append=modprobe.blacklist=hid_uclogic --append=modprobe.blacklist=wacom` & rebooting makes my XP-Pen Artist 16 (1st gen) work with OTD when it otherwise wouldn't. I'm unsure of the side effects this could cause, as I do not own a Wacom tablet.


KyleGospo:
Can you change the OSTree command to the append if missing variant? Otherwise they can double up.

KyleGospo:
And thanks!

PR #3216, opened 2025-09-14
chore: Update Mutter to 48.5

trbocode:


KyleGospo:
We match Fedora, they are only on 48.4 at this time and this is missing changes between 48.1 and 48.4

PR #3217, opened 2025-09-15
feat: add authors to gh release changelogs

xXJSONDeruloXx:
Minor tweak to tag authors in GitHub release changelog. Nice way to credit each feature or fix (and also quickly get blame if something broke lol)


antheas:
You didn't test it locally, huh

antheas:
Author should go first, your syntax is wrong (there is an excess @), subject can have an arbitrary number of spaces, so author should be put first in the command too. Author can also have spaces, name may be too big

PR #3239, opened 2025-09-19
feat: universal decky plugin installer ujust, refactor to DRY decky installers

xXJSONDeruloXx:
Many decky plugins are not on the store but useful for various compatibility or hardware fixes/control.

Instead of making a ujust for each, I figure its a good time to reduce repetition and make a one size fits all decky plugin installer + remove old versions in favor of universal one.

Should make it easier to add new non-store plugins as we go forward.

Moved decky loader installer and this new universal plugin installation feat into their own justfile for better organization. 

image


antheas:
We do not support these decky plugins: LegionGoRemapper, GPDControl, SimpleDeckyTDP.

Handheld daemon provides functionality for all of these and they tend to introduce issues. If users want them they can install them themselves. We shouldn't encourage their use.

xXJSONDeruloXx:
@antheas good call, removed those

PR #3245, opened 2025-09-21
fix: re-add hidden ujust to fix GPP0 wake on Gigabyte mobos

xXJSONDeruloXx:
Initially implemented first as a ujust, then as a udev rule here: https://github.com/ublue-os/bazzite/pull/3025

then reverted as it had adverse effects on select gigabyte motherboards here: https://github.com/ublue-os/bazzite/pull/3200

to split the difference, heres a hidden ujust so its available for those who need it, but not a footgun for those who it won't apply to. This way its something we can easily point to (or I can apply to my HTPC when I next break my install!), instead of digging up the og reddit discussion and solution this is sourced from:

https://github.com/DAK404/OpenSUSE-Setup-Scripts/blob/main/Scriptlets/Fixes-and-Tweaks/Fix-GigabyteDesktopSleepFix.sh

### DIFF
--- a/system_files/desktop/shared/usr/lib/systemd/system/disable-gpp0-wakeup.service
+++ b/system_files/desktop/shared/usr/lib/systemd/system/disable-gpp0-wakeup.service
@@ -1,0 +1,9 @@
+[Unit]
+Description=Workaround for Gigabyte BIOS sleep/wakeup bug
+
+[Service]
+Type=oneshot
+ExecStart=/bin/sh -c 'if grep 'GPP0' /proc/acpi/wakeup | grep -q 'enabled'; then echo 'GPP0' > /proc/acpi/wakeup; fi'
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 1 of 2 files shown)


grumd:
Fun fact, my ASUS mobo has the same wakeup issue, but I had to disable XHC0-1-2 (one of them, idk which one)

```
[Service]
Type=oneshot
ExecStart=/bin/sh -c 'if grep 'XHC0' /proc/acpi/wakeup | grep -q 'enabled'; then echo 'XHC0' > /proc/acpi/wakeup; fi'
ExecStart=/bin/sh -c 'if grep 'XHC1' /proc/acpi/wakeup | grep -q 'enabled'; then echo 'XHC1' > /proc/acpi/wakeup; fi'
ExecStart=/bin/sh -c 'if grep 'XHC2' /proc/acpi/wakeup | grep -q 'enabled'; then echo 'XHC2' > /proc/acpi/wakeup; fi'
```

PR #3380, opened 2025-10-30
refactor(ujust): skip unnecessary sudo for install-coolercontrol 

DDjivan:
avoid the use of a sudo command when the Terra Repository is already enabled


antheas:
What does this solve?

DDjivan:
This avoids typing a password if the Terra Repository is already enabled 
Also, I changed the shebang so that `env` is used, like other `ujust` recipes do

KyleGospo:
Thanks!

Issue #3382, opened 2025-10-30
Waydroid needs a ujust script adjustment as its making gear lever fail to launch


colelisiongaming:
### Describe the bug

When Waydroid is installed first it doesn't currently mkdir in ~/local/share/ called applications. Instead It makes a desktop entry file called "applications" This will cause apps like gear lever to fail to start as its expecting a folder in that spot. 

Most of the original troubleshooting is in https://discord.com/channels/1072614816579063828/1433452954307919923

### What did you expect to happen?

Another user in the discord amelc suggested the script should make the directory incase it doesn't exist to fix this behavior. That way a file wont be created and gear lever wont break.

### Session Logs

```shell
https://paste.centos.org/view/154a320f
```

### Hardware

Framework 13 Ultra series 1

askpng:
The PR has been merged to testing; I believe maintainers can close this issue.

PR #3386, opened 2025-10-31
refactor(ujust): Create 'applications' folder if it doesn't exist yet in Waydroid initialization script

askpng:
When a user initializes Waydroid immediately after installing Bazzite, the `~/.local/share/applications` directory might not be created yet. So, `/usr/share/applications/waydroid-container-restart.desktop` will be copied as a file called `~/.local/share/applications`, breaking other apps that export to `applications/` like Gear Lever (such as in #3382).

This PR adds a line to create `~/.local/share/applications` before copying `waydroid-container-restart.desktop` to prevent this error.


KyleGospo:
Thanks!

PR #3434, opened 2025-11-05
Update DisplayLink support note in README

ThatHummingBoy:
Clarified DisplayLink support status in README.


antheas:
Did you try to layer it?

PR #3451, opened 2025-11-08
fix(gamescope): update to latest commit hash

porkloin:
I actually don't completely know how this works or if I need to rebase/reapply all of the patches listed here, but I'm hoping to gamescope HDR fixes from [this issue](https://github.com/ValveSoftware/gamescope/pull/1999) into the bazzite build.

I'm happy to do more work if necessary, just took a cursory look and thought maybe it's as simple as updating the commit hash here. It seemed like previous commits to the file only bumped the hash so I figure why not give it a shot :stuck_out_tongue: 

Leaving this in draft just to see if CI freaks out and pending help from anyone who might know what up


porkloin:
Closing since @antheas bumped gamescope elsewhere

PR #3470, opened 2025-11-11
Fix #3399 xbox360 wireless dpad mapping 

### LINKED ISSUE #3399: XBOX 360 wireless d-pad inputs are incorrect
### Describe the bug

A recent update seems to have broken the d-pad mapping for my XBOX 360 wireless controllers.

I have two controllers and the new mapping appears to be:

| d-pad button | direction |
|------------------------|---------------|
|up | left |
|down | right |
|left|up|
|right|down|

where the direction is verified using Steam's gamepad tester.

Both controllers are mapped correctly when I switch back to a Windows 11 partition.

I expected the buttons to correspond to the correction directions, just like they did prior to a few days ago.

| d-pad button | direction |
|------------------------|---------------|
|up | up |
|down | down |
|left|left|
|right|right|

### Session Logs

```shell
https://paste.centos.org/view/16f026ee
```

### Hardware

Lenovo Yoga 7 2-in-1 (model 14AHP9) with a Ryzen 8840HS.

XBOX 360 wireless controllers connected with a Microsoft USB receiver.

The controllers were working fine from February to September of this year.

SirFerdek:
Recent changes to xpad kernel driver in kernel 6.17 broke controller DPad mappings for Xbox 360 wireless for all existing deployments of SDL libraries, including statically linked ones.

Changing button mappings in Steam and enabling Steam Input works as a temporary workaround, but is overwritten by steam itself every time it is restarted (`.steam/steam/config/config.vdf` contains `SDL_GamepadBind` variable with button mappings, but its value is refreshed with SDL-provided mappings if match is found every time the Steam starts).

Try to allieviate this pain by passing a now-correct DPad mapping as `SDL_GAMECONTROLLERCONFIG` environment variable.

In case of more troubles in the future similar to this one, to enable end-users to provide their own mappings as quick-fixes, `bazzite-steam` script was modified to not overwrite an existing `SDL_GAMECONTROLLERCONFIG` environment variable, but to prepend new mappings to it.

Finally a new mapping is injected, similarly to how support for HHD is added.
    
upstream kernel discussion:
https://lore.kernel.org/all/20250702034740.124817-1-vi@endrift.com/

### DIFF
--- a/system_files/desktop/shared/usr/bin/bazzite-steam
+++ b/system_files/desktop/shared/usr/bin/bazzite-steam
@@ -28,12 +28,16 @@ if [[ $IMAGE_NAME =~ "deck" || $IMAGE_NAME =~ "ally" ]]; then
   else
     DECK_OPTION=""
   fi
-
-  # HHD Support
-  if ! /usr/libexec/hwsupport/valve-hardware; then
-      export SDL_GAMECONTROLLERCONFIG="060000000d0f00009601000000000000,Steam Controller (HHD),a:b0,b:b1,x:b2,y:b3,back:b6,guide:b8,start:b7,leftstick:b9,rightstick:b10,leftshoulder:b4,rightshoulder:b5,dpup:h0.1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,leftx:a0,lefty:a1,rightx:a3,righty:a4,lefttrigger:a2,righttrigger:a5,paddle1:b13,paddle2:b12,paddle3:b15,paddle4:b14,misc2:b11,misc3:b16,misc4:b17,crc:ea35,"
-  fi
 fi
+
+# HHD Support
+# Fix XBOX 360 Wireless Controller for kernels >= 6.17 and SDL < 3.4.x
+export SDL_GAMECONTROLLERCONFIG=$(cat <<-END
+060000000d0f00009601000000000000,Steam Controller (HHD),a:b0,b:b1,x:b2,y:b3,back:b6,guide:b8,start:b7,leftstick:b9,rightstick:b10,leftshoulder:b4,rightshoulder:b5,dpup:h0.1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,leftx:a0,lefty:a1,rightx:a3,righty:a4,lefttrigger:a2,righttrigger:a5,paddle1:b13,paddle2:b12,paddle3:b15,paddle4:b14,misc2:b11,misc3:b16,misc4:b17,crc:ea35,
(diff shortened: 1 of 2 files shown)


antheas:
Hm, we have a problem here. I have a secondary override here:
https://github.com/bazzite-org/gamescope-session-steam/blob/ba967fd8c5de7dc6c623b614296b3872255996b0/usr/share/gamescope-session-plus/sessions.d/steam#L126-L127

For deck images you need to change it here

antheas:
Feel free to remove the hw check and merge the blocks. There is no reason steam deck should be skipped anyway. Then secondary PR there.

antheas:
Are you sure this will not cause issues for us? Is there a link for this being a credible fix?

antheas:
Ok I reviewed the SDL PRs. This is good to merge, just tweak and cross-PR.

SirFerdek:
> Ok I reviewed the SDL PRs. This is good to merge, just tweak and cross-PR.

ACK, and yeah, IMHO the kernel change is going to cause so much chaos, I have some doubts if it was worth it.

I politely asked if there is even a slim chance of backporting the SDL fix to stable branches, but will this also fix Steam/Proton/whatever Valve ships? I guess no, so this workaround with env variable is the best we can do for now.

> Are you sure this will not cause issues for us?

Nothing should change for existing users not using XBOX 360 wireless controller. The only users that will be complaining one way or another are the ones that still use the old trusty XBox360 Wireless. This workaround might break once again after SDL is updated, by either Bazzite of Valve if they ship it separately as a part of Proton.

> Hm, we have a problem here. I have a secondary override here:

Ok, please tell me if I understood this correctly:

1. Since this `sessions.d/steam` file determines gamescope session settings/environment variables, we are most probably not even logged in when this "runs",  there is almost zero chance that something user-editable had already set up the `SDL_GAMECONTROLLERCONFIG` variable, so there is no need to retain its current value there? Otherwise we should already have an issue report... :P
2. You just want me to add this Xbox workaround there as an additional mapping.

Thanks

antheas:
Make sure this workaround works with newer SDL versions

No, it's a good idea to have the override still work, just make sure it's in both places 🙂

SirFerdek:
> Make sure this workaround works with newer SDL versions

As long as we are on a kernel >=6.17 and the kernel does not alter the deal any further, we should be fine. At least that's what I understood by studying the simple changes done in libsdl-org/SDL#14339 (effectively removal of all mappings related to Xbox360 wireless). The explicit environment mapping must be provided, however, to override these now invalid, "hardcoded" mappings inside older SDL.

Currently there is no SDL release that would contain the fix. 

Opened https://github.com/bazzite-org/gamescope-session-steam/pull/5

PR #3560, opened 2025-11-27
fix: Apply GSK render fix to all Intel GPUs

HikariKnight:
bug report from discord: https://discord.com/channels/1072614816579063828/1443307686375587860
seems the GSK render issue in hardware accelerated GTK/libadwaita applications did not just affect Xe/Arc cards, it also affects some cards using the i915 driver too.

marking as draft for now while i do some more testing to see if more cards are affected to determine how much this has to be fine tuned.


HikariKnight:
turns out more legacy intel cards are now affected

Issue #3419, opened 2025-11-03
Aptx Codec Not working


n1zmarck:
### Describe the bug

<img width="546" height="317" alt="Image" src="https://github.com/user-attachments/assets/f54b6c1e-9720-4c65-a002-b1baf0dbca67" />

Aptx Codec is not running even though running on 43.20251102.

### What did you expect to happen?

Options to change to aptX HD

### Session Logs

```shell
https://paste.centos.org/view/d1afbada
```

### Hardware

Framework 16

qherreros:
I can confirm, libfreeaptx is here now, but it feels like pipewire/something_in_the_stack has not been compiled with it.

qherreros:
Probably a problem with **pipewire-libs-extra** that has been removed also.

n1zmarck:
Update, new updates have yet to fix the issue.

qherreros:
I tried to install manually pipewire-codec-aptx, which is installed in 42, and it's not even available in 43.

qherreros:
Ok solved:

- Download RPM [here](https://mirror.netsite.dk/rpmfusion/free/fedora/releases/43/Everything/x86_64/os/repoview/pipewire-codec-aptx.html)
- Go to the folder where you downloaded it
- Open terminal
- `rpm-ostree install pipewire-codec-aptx-1.4.8-1.fc43.x86_64.rpm`
- reboot

It feels like it should be easier than that to be honest. Especially for a beginner friendly distribution.

n1zmarck:
New Update has fixed issue as of F43.20251127

PR #3570, opened 2025-11-30
fix: enable aptx again by adding pipewire-codec-aptx to installation list

qherreros:
Trying to solve https://github.com/ublue-os/bazzite/issues/3419.


qherreros:
It might be useless now that `pipewire-libs-extra` is back to ublue-base.

PR #3581, opened 2025-12-02
feat(bazzite-hardware-setup): Add workaround for nvidia flatpak mismatch

ykshek:
This adds two new variables, `$SYSTEM_NVIDIA_VERSION` and `$FLATPAK_NVIDIA_VERSION`. 
The variables will be empty if the system is not nvidia. 
If they have a mismatch, install the matching version on flatpak.

This should fix the nvidia system and flatpak driver mismatch issue, however as yukidream mentioned in #bazzite-dev, there appears to be a better way to fix it.


ykshek:
image

Would fix user complaints such as this one

KyleGospo (approved):

KyleGospo:
This is great, thanks!

HikariKnight (approved):
nice, less work for nvidia users.
sonic-boom-knuckles-the-echidna

PR #3620, opened 2025-12-11
fix(kde): Apply ibus change to ptyxis dbus activation

### LINKED ISSUE #3482: Update from FC42 to 43 breaks dead keys on Terminal with DK layout
### Describe the bug

Most notably, the ~ (tilde) character is no longer generated in the Terminal application,  when using **AltGr** + **tilde button** using a standard DK keyboard layout (pc105).

Other applications like Kate are producing the tilde character as usual.
Please note that **none** of the two dead keys produce _any_ of their supposed characters;
i just mentioned tilde because that's the one I use all of the time...

Having to have Kate open so I can copy-paste it into the Terminal becomes tedious real fast.

I think that's pretty obvious...

### Session Logs

```shell
-
```

### Hardware

Standard x86_64 desktop PC.

Rebasing back to stable-42 gives me functionality back immediately.

HikariKnight:
Should fix #3482


antheas:
Why doesn't the existing service have the correct executable?

HikariKnight:
> Why doesn't the existing service have the correct executable?

because this is only an issue on non gnome desktops, because ptyxis is probably only tested on gnome/gtk desktops upstream?

PR #3628, opened 2025-12-12
fix(ujust): enable Ryzenadj max performance

dkatzdev:
Fix: Corrected the ujust script for enabling and disabling the max-performance mode in ryzenadj.

Previously the script would not run after every plug or unplug due to syntax issues with the "add|change" part of the script, and the fact that the run portion had an invalid bin/sh which would lose single quotes when written, but I corrected it and it now successfully runs and writes all files properly.

It also left an edge case where if you booted the device while unplugged, it would never run the script and you would experience the performance issue until plugging and unplugging again. I added a oneshot boot service to fix that.

### DIFF
--- a/system_files/deck/shared/usr/share/ublue-os/just/85-bazzite-image.just
+++ b/system_files/deck/shared/usr/share/ublue-os/just/85-bazzite-image.just
@@ -75,10 +75,10 @@ enable-ryzenadj-max-performance:
     #/bin/bash
     # credit to adolforegosa on discord for the original fix and Katman for the script corrections
     echo 'this fix sets ryzenadj --max-performance whenever AC status changes to battery from plugged-in'
-    echo 'ACTION=="change", SUBSYSTEM=="power_supply", ATTR{online}=="0", RUN+="/bin/sh -c '/usr/bin/ryzenadj --max-performance'"' | sudo tee "/etc/udev/rules.d/99-ryzenadj-power-source-change.rules" > /dev/null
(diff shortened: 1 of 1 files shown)


antheas:
Can you make due without the service? Should be able to add a second udev rule or something

dkatzdev:
its a lot messier to use udev rules for boot oneshots. technically this whole thing should be a service instead of a rule, since itll trigger and execute more cleanly... when I tried to make it a rule for boot it didnt seem to run because there was no triggering event to do it. it would either have to run constantly or wait for something specific (the unplug)

dkatzdev:
I suppose an option would be to make a oneshot boot service that directly triggers the rule at boot, but thats kinda the same thing

dkatzdev:
udev rules cant guarantee they execute at boot and the workarounds are hacky. even adding the "add" part back to the add|change doesnt guarantee it actually happens

dkatzdev:
Just to clarify, the service is a oneshot type, so it wont be running in the background continuously or anything. Literally runs at boot and then doesnt need to run again because the udev rule takes over on plug events

antheas:
This makes the ujust dirtier in the future to clean up. It would be very good to avoid it if possible

dkatzdev:
What do you mean dirtier? The enable creates it and the disable removes it cleanly. I could make the whole thing operate in the service so it doesn't need the udev rule too, but doing it in udev at all is already not best practice for stuff like this and there's no good way to make it reliably run at boot through udev. like it will literally require plugging and unplugging every time you want it to start

dkatzdev:
@antheas If it's a matter of formatting I can just use multiple echo commands for each line instead of the cat EOF block. That's what some of the other commands do. its just clunkier

PR #3659, opened 2025-12-18
fix: indentation in max performance script so file writes properly

dkatzdev:
Script doesnt work properly when the service contents are indented. This reverts that and solves it.


dkatzdev:
If you dont like the way it looks we can write it out as a series of echo commands. as long as each line is on its own line in the resulting file.

KyleGospo:
This falls just linting, making the build immediately stop.

dkatzdev:
Alright, I'll switch it to echos so it's cleaner and passes

dkatzdev:
It should be good now, I switched to echo commands, and made sure the first one writes a new file instead of appending just in case someone runs it multiple times by accident

KyleGospo (approved):

PR #3666, opened 2025-12-19
feat: Speed up login/out sequence by disabling sound effects

Zeglius:


EPOCHvoyager:
I like the sound effects, adds to the UX polish. -1 from me!

Zeglius:
Yeah, we should discuss if this is too much minmaxing, given there are other issues to deal with, like zstd compression not being applied

Tenama-hub:
On one hand, we get a noticeable boost in logoff and login operations. This is something that Kubuntu, Manjaro and some Arch distributions do by default. Bazzite already suffers from slow boot times, so any tweaking or system change that doesn't affect the end user and benefits the distribution should be welcomed.
On the other hand, it may affect the "user experience" of how the Desktop welcomes the user in the UI. If that is the case, it can be turned on at the user's request in "System Notification" setting.

I feel like this change gives more benefits than setbacks. I doubt people would care that their computer doesn't play a loud noise during boot or shut down, and would rather appreciate the "snappiness" provided by the change.
As a bonus, we should have an option for turning off the splash screen animation at boot. This is something I could do in Fedora KDE, but for some reason it's missing in Bazzite.

KyleGospo:
You can set the system sounds yourself in settings, I think this change is unneeded.

-1

PR #3712, opened 2025-12-25
fix: prevent NM from starting before iwd inits

xXJSONDeruloXx:
when rebasing from stable to testing, the iwd defaulting swap recently is causing wifi to be unusable, with some logs like

`iwd: station: Network configuration is disabled.`

`iwd: station: Quick network configuration is disabled.`

`device (wlp4s0): state change: unavailable -> unmanaged (reason 'unmanaged-link-not-init')`

fix was to rebase back to stable, run `ujust toggle-iwd enable` then rebase back to testing (which removes this recipe entirely)

to make transition smoother for users especially when this goes to main, I think (but could be very wrong, new to the networking services side here) that the fix is to ensure that NetworkingManager does not start before iwd has initiated the wifi device, hence the `unmanaged-link-not-init`.

Soooo I think best way to solve here is add a conf for NM service to force it after init of iwd? If we are all in on only iwd going forward (implied by the removal of the ujust) then this shouldn't be an issue right? Do let me know if theres anything here I have wrong!

### DIFF
--- a/system_files/desktop/shared/usr/lib/systemd/system/NetworkManager.service.d/iwd.conf
+++ b/system_files/desktop/shared/usr/lib/systemd/system/NetworkManager.service.d/iwd.conf
@@ -1,0 +1,4 @@
+[Unit]
+# Ensure iwd is started before NetworkManager when using iwd as WiFi backend
+Wants=iwd.service
+After=iwd.service


KyleGospo:
iwd.service is disabled and started by nm, this would never run

xXJSONDeruloXx:
Something is borked if you don't manually toggle iwd on before the latest changes, I fear when we merge to main this will cause lots of issues for users

PR #3723, opened 2025-12-27
fix: set output img to push repo instead of pull repo

porkloin:
When testing building images on my fork, I realized that the `output_image` variable was being set to use the `PULL_REGISTRY` variable for the repo owner, which was always set to `ublue-os` instead of whatever the actual repo the action is being run from.

Updates the var to use `PUSH_REGISTRY` instead.


KyleGospo (approved):

porkloin:
Had to fix up merge conflict after base change

PR #3728, opened 2025-12-29
fix(crh): correct symlink to relative path

ykshek:
Correct symlink to relative instead of absolute,
Thanks to @asen23 

An absolute blunder on my part - oops!


HikariKnight (dismissed):
lol

PR #3733, opened 2025-12-29
fix(crh): fix a pipefail in print_status

ykshek:
Throw away the results of grep, as that is immediately handled afterwards.

This fixes an issue where the script exits if there is no custom resolutions set, which is presumably not very useful for users.


ykshek:
I think I messed something up with the branching? Probably should've branched from main ahhhh :dispair:

PR #3739, opened 2025-12-29
fix: rechunker being stuck

renner0e:
This reverts commit 57015ad1513a43df1bb35fc58d8188479538ff5d.

for whatever reason this makes the rechunker choke
https://github.com/ublue-os/bazzite/pull/3728/changes
it works for me:
https://github.com/renner0e/bazzite/actions/runs/20579908088/job/59104993713


HikariKnight:
why would that symlink make rechunker choke... it makes no sense as its identical to brh :thinking:

KyleGospo:
Instead of a relative link this should be a direct link to the file in /usr/bin

PR #3764, opened 2025-12-31
feat(just): Add configure-beesd

BWBellairs:
Add new ujust command for configuring beesd (Recently added to the testing tag)
I've chosen a database size recommended by the author of the project [here](https://github.com/Zygo/bees/blob/master/docs/config.md)

Unfortunately, the `--throttle-factor FACTOR` option is not available with the version in fedora repos.

    --throttle-factor FACTOR

> In order to avoid saturating btrfs deferred work queues, bees tracks the time that operations with delayed effect (dedupe and tmpfile copy) and operations with long run times (LOGICAL_INO) run. If an operation finishes before the average run time for that operation, bees will sleep for the remainder of the average run time, so that operations are submitted to btrfs at a rate similar to the rate that btrfs can complete them.
> High values of FACTOR may be desirable when using bees with other applications on the same filesystem.
> 
> The maximum delay per operation is 60 seconds.
[source](https://github.com/Zygo/bees/blob/master/docs/options.md#load-management-options)

### DIFF
--- a/system_files/desktop/shared/usr/share/ublue-os/just/82-bazzite-beesd.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/82-bazzite-beesd.just
@@ -1,0 +1,80 @@
+# Configure beesd for BTRFS filesystems
+[group("system")]
+configure-beesd ACTION="":
+    #!/usr/bin/bash
+    source /usr/lib/ujust/ujust.sh
+    
+    while true; do
+        # Get btrfs filesystem uuids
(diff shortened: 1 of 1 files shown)


KyleGospo:
This is wonderful, thank you! 

I will work on getting that feature in the package in fedora's repos.

KyleGospo (approved):

kressb-gitlab:
While /usr/share/ublue-os/just/82-bazzite-beesd.just seems to have made it into the image, /usr/share/ublue-os/justfile does not include it, so this is not actually able to be used yet.

Also for what it's worth, the systemd service file adds "--no--timestamps" already so adding it to the config files means it's on the command line twice.  Harmless, but cosmetic.

PR #3791, opened 2026-01-02
fix(steamos-update): Prevent first boot gamemode hang

Zeglius:
Introduce exit codes for clarity and refactor `fake_progress` and `upgrade` into private functions within `steamos-update`.

Cleaning up nested if-else conditions.

Implement a fixup mechanism to allow the Steam client's first boot in Deck Mode to proceed without getting stuck in a 'STEAMOS' screen, by exiting `steamos-update` early. This addresses a known issue (see: https://www.reddit.com/r/Bazzite/comments/1psuar7/bazzite_xbox_ally_issue/).

### DIFF
--- a/system_files/deck/shared/usr/bin/steamos-update
+++ b/system_files/deck/shared/usr/bin/steamos-update
@@ -7,84 +7,104 @@
 # steamos-update --enable-duplicate-detection       -- should perform an update
 # steamos-update                                    -- should perform an update
 
+# Exit Codes:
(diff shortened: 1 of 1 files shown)


KyleGospo:
I assume this isn't tested? Only because of the difficulty in doing so

Zeglius:
> I assume this isn't tested? Only because of the difficulty in doing so

Right, probably should spin some isos from testing and have someone try them.

PR #3795, opened 2026-01-02
fix: restore Dualsense haptics functionality

stellaberrant:
override USB-Audio.conf with a known good copy from 43.20251210.1 to fix Dualsense haptics being broken, upstream regression in alsa 1.2.15

### DIFF
--- a/system_files/overrides/usr/share/alsa/ucm2/USB-Audio/USB-Audio.conf
+++ b/system_files/overrides/usr/share/alsa/ucm2/USB-Audio/USB-Audio.conf
@@ -1,0 +1,590 @@
+Syntax 7
+
+Define.ProfileName ""
+Define.MixerRemap ""
+Define.SplitPCMPeriodTime 10000		# 10ms
+
+If.env1 {
+	Condition {
+		Type String
+		Empty "$${env:UCM_USB_PERIOD_TIME}"
+	}
+	False.Define.SplitPCMPeriodTime "${env:UCM_USB_PERIOD_TIME}"
+}
+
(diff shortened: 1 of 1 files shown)


gitty323:
Hello, this commit broke GoXLR audio input and outputs. Is there a way to keep that intact while also having Dualsense haptics? I rebased through the testing versions and in [testing-43.20260102](https://github.com/ublue-os/bazzite/releases/tag/testing-43.20260102) everything works as it has for the past few months. When I rebase to [testing-43.20260102.1](https://github.com/ublue-os/bazzite/releases/tag/testing-43.20260102.1), it breaks and it is broken in the latest stable builds. (That is what made me go looking for the issue. I originally assumed it was an alsa-lib update causing the issue but it is this one commit that seems to have done it).

See https://github.com/GoXLR-on-Linux/goxlr-utility/issues/238 for details on what was broken.

Thank you!

stellaberrant:
i'm hoping the issue will be resolved upstream soon so we can remove the override - it's being tracked here https://github.com/alsa-project/alsa-ucm-conf/issues/677

jwlademann:
All this effectively does is turn off the ucm config for dualsense altogether since there's no condition that matches the controller (and subsequently breaks functionality for any configs using v8 syntax). I believe you could effectively do the same thing by taking a copy of the current USB-Audio.conf file and removing the "If.sony-dualsense-ps5 {" block.

jwlademann:
Implement the above suggestion here: https://github.com/ublue-os/bazzite/pull/3986

PR #3799, opened 2026-01-02
feat(ujust): replace rpm-ostree rmlint with brew

Xarishark:


KyleGospo:
This ujust can be entirely removed, bees does this better.

PR #3800, opened 2026-01-02
chore: remove rmlint justfile

Xarishark:


KyleGospo:
https://github.com/ublue-os/bazzite/blob/main/system_files%2Fdesktop%2Fshared%2Fusr%2Fshare%2Fublue-os%2Fjust%2Frmlint-override.conf

This file can also be removed, and there's a single entry in the Containerfile pointing to the file you removed which also needs to be pulled.

Thanks for doing this!

Xarishark:
> https://github.com/ublue-os/bazzite/blob/main/system_files%2Fdesktop%2Fshared%2Fusr%2Fshare%2Fublue-os%2Fjust%2Frmlint-override.conf
> 
> This file can also be removed, and there's a single entry in the Containerfile pointing to the file you removed which also needs to be pulled.
> 
> Thanks for doing this!

removed both. happy to help!

PR #3819, opened 2026-01-04
feat(just): Use Steam Audio HRTF for surround

EPOCHvoyager:
Valve's Steam Audio library is distributed with a compatible license  — actually the same as Bazzite itself — which allows us to pull their excellent HRTF for this. I've taken the liberty to remove the step-by-step personalization guide too, as this generalized HRTF is far more appropriate for this use case.

The poor quality default user experience, that with the placeholder HRTF, was discussed in the original PR (#1838) — after a posthumous comment by @AlexNPavel. I believe these changes rectify any such issues; the default user experience is now, in my disinterested opinion, splendid.


EPOCHvoyager:
Of note, I'll likely write a page in the Bazzite docs for this — to replace the original step-by-step's headphone correction bit. Volume needs to be lowered a good amount already to avoid clipping here, and with headphone correction equalization, it compounds.

PR #3831, opened 2026-01-04
feat(desktop) increase zram size to match deck builds

stellaberrant:
Increase zram size to half of physical RAM or 4GB if below 8GB. This is to better match deck builds and prevent the system from totally freezing due to running out of RAM which is a bad user experience. This matches the Steam Deck config of 16GB/8GB swap.


KyleGospo:
On a system like mine (64GB RAM) this would be an objectively bad experience and a complete waste of that RAM.

This needs a sensible limit, preferably one that is relative to the max as even the 4GB I use right now is more than I need.

I would also just delete the deck config and make them match it we're choosing to deviate from Fedora behavior.

stellaberrant:
> On a system like mine (64GB RAM) this would be an objectively bad experience and a complete waste of that RAM.
> 
> This needs a sane limit.

How about zram-size=min(ram/2, 8096) ?

PR #3837, opened 2026-01-05
feat: add bazzite-rollback-helper application shortcut for Gnome and KDE

xXJSONDeruloXx:
Add an icon in desktop just like we do for portal and updater, to get uers into the TUI easier, and maybe slightly less intimidating. At least till we make a proper GUI soon TM.

using the regular icon svg for now, but lmk if theres a better one in brand bible, or if we want to make a unique one


HikariKnight (approved):

KyleGospo:
Let's wait on this until the GUI updater/rebaser is ready. One less TUI and one less switcheroo.

PR #3849, opened 2026-01-06
feat(beesd): Implement daily timer with resource limits

BWBellairs:
Hash table is now matched against filesystem size.
To limit resources, the service now exits after 10 minutes and will only start if enough memory is free for bees' hash table.

Would be nice if it only ran when the user is idle but I could not think of a way to implement that.
Hopefully haven't made any mistakes this time around.

### DIFF
--- a/system_files/desktop/shared/etc/systemd/beesd@.service.d/override.conf
+++ b/system_files/desktop/shared/etc/systemd/beesd@.service.d/override.conf
@@ -1,0 +1,7 @@
+[Service]
+ExecStart=
+ExecStart=/usr/bin/timeout 10m /usr/bin/beesd --no-timestamps %i
+Restart=no
+SuccessExitStatus=124
+OOMScoreAdjust=1000
+OOMPolicy=stop
--- a/system_files/desktop/shared/usr/lib/systemd/system/beesd@.timer
+++ b/system_files/desktop/shared/usr/lib/systemd/system/beesd@.timer
@@ -1,0 +1,7 @@
+[Timer]
+OnCalendar=daily
+Persistent=true
+RandomizedDelaySec=1h
+
+[Install]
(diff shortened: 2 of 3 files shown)


KyleGospo (approved):

EPOCHvoyager:
@BWBellairs You could trigger it after the uupd service is done, as that has idle checks.

EPOCHvoyager:
Also, your timeout seems to not be working. 

EDIT: Probably best to remove that, and add a first run to the ujust? Could also limit thread usage like we used to do with duperemove.

EPOCHvoyager:
~~This is how dedup implemented idle detection~~ (actually a connected power supply), would necessitate a wrapper script of some sort here, I believe — https://github.com/ublue-os/bazzite/blob/162732636282e62a771cd314c92590b5832e5c83/system_files/desktop/shared/usr/bin/btrfs-dedup#L147C1-L147C16

CPU limitation is toward the end of the file. Note that any of these could be implemented as an upstream feature as well.

BWBellairs:
> @BWBellairs You could trigger it after the uupd service is done, as that has idle checks.

This would definitely simplify the current implementation

> Also, your timeout seems to not be working.
> 
> EDIT: Probably best to remove that, and add a first run to the ujust? Could also limit thread usage like we used to do with duperemove.

I'd prefer the timeout stayed as it frees the memory back to the system rather than keeping it always.
The thread parameters could be amended.

BWBellairs:
It seems that the override file doesn't exist (From `systemctl cat beesd@.service`)

EPOCHvoyager:
> I'd prefer the timeout stayed as it frees the memory back to the system rather than keeping it always.

That does make sense. However, I guess tangentially, adding that potentially long first run to the ujust would be better UX.

PR #3889, opened 2026-01-11
fix(nvidia): add back VK_hdr_layer for nvidia images

HikariKnight:
seems to still be required by nvidia, see: https://github.com/ublue-os/bazzite/commit/76f5d93e31a64a1d3ad1d95c39dfb55b34726859#commitcomment-173916251

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -799,7 +799,8 @@ RUN --mount=type=cache,dst=/var/cache \
     dnf5 -y copr enable ublue-os/staging && \
     dnf5 -y install \
         egl-wayland.x86_64 \
-        egl-wayland.i686 && \
+        egl-wayland.i686 \
+        VK_hdr_layer && \
     /ctx/install-nvidia && \
     rm -f /usr/share/vulkan/icd.d/nouveau_icd.*.json && \
     ln -s libnvidia-ml.so.1 /usr/lib64/libnvidia-ml.so && \


KyleGospo:
Nvidia needs to get their shit together.

-EDIT-
This package has been unneeded Since Mesa 25.1, and is marked as "Obsolete" in Mesa 26's Fedora package - meaning you can't have both going forward and layering will begin to fail. See here:
https://src.fedoraproject.org/rpms/mesa/blob/rawhide/f/mesa.spec#_350

If there's any remaining issues with the lack of this package it's on Nvidia to fix, we cannot do that for them because it's a closed source driver. We do our utmost to ensure the latest Nvidia driver is what's shipped and will rush out a fix for this whenever Nvidia's driver supports the finalized HDR standards.

In the meantime you should be able to use Gamescope to accomplish HDR on the desktop and we ship ScopeBuddy to make that as easy as possible.

Sorry folks, I know that sucks. There's not a lot of good options. My apologies for the poor communication in this pull.

HikariKnight:
> Nvidia needs to get their shit together

agreed and very fair

Koeleskab:
I have to chime in here because this move feels counter-intuitive to the Bazzite mission. By removing this, you are placing the burden of friction directly on the end-user. specifically those with Nvidia GPUs and HDR monitors who might already face an uphill battle on Linux when first joining the community.

While I understand the frustration with Nvidia’s lack of cooperation with the linux community, they aren’t the ones who will feel this change; the users will. They'll never prioritize this within any sensible timeline. 

So far, the reasoning provided has been limited to one-liners that don't explain the technical necessity of breaking features for thousands of users. If there are massive stability risks or technical blockers that make the current implementation untenable, please share them. Without that transparency, this feels less like a technical necessity and more like a philosophical stance that sacrifices usability for a point Nvidia won't even notice. Is this kind of tailoring not exactly the point of having bespoke images for Nvidia?

I'll just go layer it. the next 500 won't. they'll be clogging the discord.

dannymate:
> I have to chime in here because this move feels counter-intuitive to the Bazzite mission. By removing this, you are placing the burden of friction directly on the end-user. specifically those with Nvidia GPUs and HDR monitors who might already face an uphill battle on Linux when first joining the community.
> 
> While I understand the frustration with Nvidia’s lack of cooperation with the linux community, they aren’t the ones who will feel this change; the users will. They'll never prioritize this within any sensible timeline.
> 
> So far, the reasoning provided has been limited to one-liners that don't explain the technical necessity of breaking features for thousands of users. If there are massive stability risks or technical blockers that make the current implementation untenable, please share them. Without that transparency, this feels less like a technical necessity and more like a philosophical stance that sacrifices usability for a point Nvidia won't even notice. Is this kind of tailoring not exactly the point of having bespoke images for Nvidia?
> 
> I'll just go layer it. the next 500 won't. they'll be clogging the discord.

Honestly I read the one line reply and assumed it was being added at some later point. I assumed as much because it's the kind of thing I thought Bazzite was striving to be. Not including the vk_hdr_layer because "**_Nvidia_**" only pushes the burden downstream to users or people like me who do their best to support others. 

A couple of other points to add to the quoted. In regards to layering being the solution, which it effectively is here, the docs heavily discourage this usage to the point I've had people scared to layer anything at all. 

I specifically have people I help switch to Linux who I will then have to tell directly to drop into the CLI to install this which is not an insignificant amount. This makes for a terrible first impression and something I avoid with all my might. I already have guided people into layering VK_hdr_layer after getting the inevitable "HDR is bad pls help" which isn't a good experience for newcomers. On top of that Gamescope does not carry great Nvidia support. In my experience it's been easier to explain to Nvidia folk to use GE-Proton with Wayland over Gamescope in general.

KyleGospo:
Sorry, we're not interested in continuing to ship this workaround.

-EDIT-
This package has been unneeded Since Mesa 25.1, and is marked as "Obsolete" in Mesa 26's Fedora package - meaning you can't have both going forward and layering will begin to fail. See here:
https://src.fedoraproject.org/rpms/mesa/blob/rawhide/f/mesa.spec#_350

If there's any remaining issues with the lack of this package it's on Nvidia to fix, we cannot do that for them because it's a closed source driver. We do our utmost to ensure the latest Nvidia driver is what's shipped and will rush out a fix for this whenever Nvidia's driver supports the finalized HDR standards.

In the meantime you should be able to use Gamescope to accomplish HDR on the desktop and we ship ScopeBuddy to make that as easy as possible.

Sorry folks, I know that sucks. There's not a lot of good options. My apologies for the poor communication in this pull.

Issue #3868, opened 2026-01-08
i18n support for system-update.desktop and other desktop files


AsciiWolf:
I may be wrong, but it looks like that the _Name_ and _Comment_ strings in [system-update.desktop](https://github.com/ublue-os/bazzite/blob/main/system_files/desktop/shared/usr/share/applications/system-update.desktop) and possibly other desktop files are English-only at the moment.

Please consider making them translatable so that Bazzite can be easily used by non-English speaking users. Thanks!

AsciiWolf:
https://github.com/ublue-os/bazzite/pull/3858

Related.

(But it would be much better to handle this on Weblate or a similar translation platform.)

theMimolet:
I agree, but it's the best we can do for now.

Xarishark:
+1 for weblate as it makes the system much more accessible for people that can provide i18n but dont know how to fork pr etc

PR #3891, opened 2026-01-11
feat: add Czech translation to the desktop files

AsciiWolf:
Helps: #3868

Also fix the incorrect CRLF line terminators in discourse.desktop (Linux uses _LF_) and a small typo in the French translation of bazzite-steam-bpm.desktop.


AsciiWolf:
Wow, that was quick! Thanks! :-D

PR #3920, opened 2026-01-15
 feat: Add CMS check to block install

Xarishark:
I disabled decorations so the user can only on the shutdown button.


KyleGospo (approved):

Xarishark:
@KyleGospo again one of my commit is unverified. Not sure what I'm doing wrong as I literally push them from the same exact IDE and committed directly to the same branch...

PR #3932, opened 2026-01-17
feat: Add installation script for asusctl

Xarishark:


KyleGospo:
I have other plans for this, at this point I'd like to never add another layering ujust ever again.

-1

PR #3962, opened 2026-01-19
Add global DLSS toggle for Proton

moi952:
Adds a new ujust toggle-global-dlss command to enable or disable DLSS upgrade globally for Proton. Integrated into the Yafti portal alongside existing FSR4 toggles for consistency.

### DIFF
--- a/system_files/desktop/shared/usr/share/ublue-os/just/80-bazzite.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/80-bazzite.just
@@ -642,6 +642,47 @@ toggle-global-fsr4-rdna3 ACTION="":
             echo "No changes made."
             ;;
     esac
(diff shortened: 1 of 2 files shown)


system_files/desktop/shared/usr/share/ublue-os/just/80-bazzite.just
```diff
@@ -642,6 +642,46 @@ toggle-global-fsr4-rdna3 ACTION="":
             echo "No changes made."
             ;;
     esac
+    
+# Toggle global Proton DLSS upgrade; updates DLSS DLLs for Proton
```
xXJSONDeruloXx: I believe this requires a proton like GE or cachy, like in the fsr 4 ones can you note that here

PR #3986, opened 2026-01-22
fix(alsa): Update USB-Audio.conf override from upstream

jwlademann:
Replace existing override with updated version from upstream with dualsense condition removed
Addresses: https://github.com/ublue-os/bazzite/pull/3986


jwlademann:
Apparently a fix is in testing for this so I'll close

PR #3996, opened 2026-01-24
Add com.github.Matoking.protontricks to Flatpak lists and simplify Bazzite Portal installation logic

Xarishark:
Include com.github.Matoking.protontricks in the GNOME and KDE Flatpak lists. Simplify the logic for installing the Bazzite Portal local .flatpak. Remove redundant checks for existing installations. Remove Protontricks from Flatpak manager script.


Xarishark:
this will have to wait until @xXJSONDeruloXx also merges the changes made [here ](https://github.com/xXJSONDeruloXx/yafti-gtk/pull/10) and [here](https://github.com/xXJSONDeruloXx/yafti-gtk/pull/11)

PR #4014, opened 2026-01-26
feat(shell): add command not found suggestions

Zeglius:
Introduce a command_not_found_handle function to provide users with command suggestions when a command is not found.

This feature enhances the shell experience by using fzf to perform a fuzzy search on the mistyped command against a list of available ujust recipes and system commands, suggesting possible alternatives.


renner0e:
just want to point out that KDE Linux has something similar to this and it might be worth looking into

https://invent.kde.org/kde-linux/kde-linux/-/blob/dbdbb552e94e7d8766d89e1a77e60447d4b5eaaf/mkosi.extra/usr/lib/command-not-found-handler.py

Zeglius:
> just want to point out that KDE Linux has something similar to this and it might be worth looking into
> 
> https://invent.kde.org/kde-linux/kde-linux/-/blob/dbdbb552e94e7d8766d89e1a77e60447d4b5eaaf/mkosi.extra/usr/lib/command-not-found-handler.py 

Interesting. Some sort of handle for alternatives would be nice for dnf/apt/pacman.

Will try to add that.

PR #4043, opened 2026-01-30
Clean up leftover kernel-devel and kernel-common packages

Xarishark:
Remove leftover kernel-devel and kernel-common packages after installation to ensure a cleaner environment. Cleanup includes removing associated source files.


KyleGospo:
Merging this but it may not be needed after the swap to OGC.

Xarishark:
Love this I can bookmark it so I can remember to revert at a later date ? Or is there is another way for temporary code ?

PR #4044, opened 2026-01-30
Remove documentation folder to save space

Xarishark:


renner0e (commented):
This would probably break this functionality:
https://github.com/ublue-os/bazzite/blob/1545b4c1e952c1af533b57952241b673c4dde386/system_files/desktop/kinoite/usr/share/templates/MangoHud.desktop

this change saves  ~190 MiB

KyleGospo:
> This would probably break this functionality:
> https://github.com/ublue-os/bazzite/blob/1545b4c1e952c1af533b57952241b673c4dde386/system_files/desktop/kinoite/usr/share/templates/MangoHud.desktop
> 
> this change saves  ~190 MiB

This could be fixed by moving that file and editing the desktop file with sed

KyleGospo:
> I think khelpcenter and the gnome equivalent should be removed from the image as well with this as those apps will be essentially useless/broken without local docs, double check on the gnome one please, not sure if it only does local docs

I think we already removed the GNOME equivalent. Did we verify this change doesn't remove any license files?

Xarishark:
image

Xarishark:
I will happily move the conf if you have a location to recommend me pls

Xarishark:
I can see the mangohud conf being referenced in [here too](https://github.com/ublue-os/bazzite/blob/testing/system_files/desktop/shared/usr/libexec/bazzite-user-setup)

renner0e:
> Did we verify this change doesn't remove any license files?

this would remove license files, we should probably not touch this then.

rough estimate on the kde image:
```
rpm -qla | grep '^/usr/share/doc' | grep -E "COPYING$|LICENSE$|GPL|BSD" | wc -l
214
```

I think that those must be packaging errors then If I understand [docs on this correctly](https://docs.fedoraproject.org/en-US/packaging-guidelines/LicensingGuidelines/#_license_text)

We could explicitly exclude those files, but this could quickly become an unmaintainable mess (that still deletes license files)

one easy win would be removing `/usr/share/doc/HTML` (60 MiB) on the kde image, this doesn't have any license files in it

renner0e (commented):
yelp should probably also be removed here, if it's no longer part of the image

https://github.com/ublue-os/bazzite/blob/1545b4c1e952c1af533b57952241b673c4dde386/system_files/desktop/silverblue/etc/dconf/db/distro.d/01-bazzite-desktop-silverblue-folders#L15

Xarishark:
> I think that those must be packaging errors then If I understand [docs on this correctly](https://docs.fedoraproject.org/en-US/packaging-guidelines/LicensingGuidelines/#_license_text)

According to this LISENCEs should be in  /usr/share/licenses/ then. I could:

`mkdir -p /usr/share/licenses`

then

```
for d in /usr/share/doc/*; do
  pkg=$(basename "$d")
  mkdir -p "/usr/share/licenses/$pkg"
  find "$d" -maxdepth 2 -type f \( -iname '*license*' -o -iname 'copyright*' -o -iname 'copying*' \) \
    -exec cp -a {} "/usr/share/licenses/$pkg/" \;
done
```
and then 

`rm -rf /usr/share/doc/`

PR #4059, opened 2026-02-01
chore(beesd): Throttle them further

EPOCHvoyager:
Usage was lowered to ~7% of a Ryzen 3700X in the last commit regarding this. While that's fairly low, it could likely still get some fans kicking, so throttle down the service even further (though it might still get them kicking on low-powered devices). Also, missed that override file last time.

### DIFF
--- a/system_files/desktop/shared/etc/systemd/system/beesd@.service.d/bees-timeout.conf
+++ b/system_files/desktop/shared/etc/systemd/system/beesd@.service.d/bees-timeout.conf
@@ -1,6 +1,6 @@
 [Service]
 ExecStart=
-ExecStart=/usr/bin/timeout 20m /usr/bin/beesd --no-timestamps %i
+ExecStart=/usr/bin/timeout 30m /usr/bin/beesd --no-timestamps %i
 Restart=no
 SuccessExitStatus=124
 OOMScoreAdjust=1000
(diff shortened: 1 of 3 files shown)


PVermeer:
Hi there,

Thanks for your work. I recently switch to beesd from duperemove before your PR's and found it to be much better. I've set it up for my secondary drives that hold a lot of collections of data that benefit massively from deduping.

I do wonder why you have a timeout on the service. The service is intended to always be run in background and, with some tweaking, handles system resources nicely so it only runs when idle and when there is actually something to be done. In my opinion this one of better features when comparing to the previous duperemove implementation.

In my case for 12 hardware threads I have:
```
OPTIONS="--strip-paths --thread-count 2 --loadavg-target 2"
```
It only kicks in when cpu `load average: <2` which is about <15%. This could be calculated based on the number of CPU's the system has.

So why have a timeout at all?

Also some suggestions that come to mind:
Maybe a new service for the bazzite setup `bazzite-beesd@.service` would be preferable, to keep the original service intact. A pre-check for the original service status (running / enabled) and with a trigger that stops / starts it when run on battery / power.

EPOCHvoyager:
> So why have a timeout at all?

I had wondered about this as well — it's to free up the decent amount of RAM the service uses once it's done. On my system, it's about half a GB.

EPOCHvoyager:
> […] I recently switch to beesd from duperemove before your PR's and found it to be much better. […]

By the way, I'm not the original author of the beesd PRs. That would be @BWBellairs; I'm only filling in due to extraneous circumstances on his end.

PVermeer:
Thanks for the reply,

I see why the timeout is there. Bees loads the whole db in memory (not swapable), which for me is 1.1GB on a 8TB drive. I've now overriden the service for the drive to not use the timeout. I'm enabling the service now only when a significant amout of data has changed.

There was some discussion about a `--oneshot` option in [this issue](https://github.com/Zygo/bees/issues/106) but it never got implemented.

KyleGospo:
+1 on this

PR #4064, opened 2026-02-02
feat: add image verification logic for unverified images

ZachrielForever:
Adding in logic to tell if a user is on a unverified image, and queues a deployment for the same image, only verified.

### DIFF
--- a/system_files/overrides/etc/profile.d/verify_motd.sh
+++ b/system_files/overrides/etc/profile.d/verify_motd.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+# 1. Get the current image reference
+CURRENT_REF=$(rpm-ostree status --json | jq -r '.deployments[0]["container-image-reference"] // empty')
+
+# 2. Define Identifiers
+UNVERIFIED_TAG="ostree-unverified-registry"
+OFFICIAL_TAG="ghcr.io/ublue-os/"
+
+# 3. The "Smart" Check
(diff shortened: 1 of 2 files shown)


KyleGospo:
A user might have a reason for being on an unverified image, including that they're using a custom image they themselves made that doesn't have a signed key applied to it.

Could this perhaps be done from the MOTD? We already show a flashing red lock to unverified users. Perhaps it can also give instructions on what to do?

ZachrielForever:
> A user might have a reason for being on an unverified image, including that they're using a custom image they themselves made that doesn't have a signed key applied to it.
> 
> Could this perhaps be done from the MOTD? We already show a flashing red lock to unverified users. Perhaps it can also give instructions on what to do?

It may work better as a ujust instead, I also wasnt thinking about the custom image scenario. 
Do custom images have any type of identifier on them? If so, I can try to add in something that ignores this check if on a custom. Otherwise i can reformat this into a ujust for ease of use, and have a MOTD notif that gives them the ujust upfront.

ZachrielForever:
Changed the whole idea. Added in a warning on the MOTD that gives red text saying the image is unverified. This wont affect custom images, as it looks specifically for the base ublue prefixes on the image identifiers. Ujust script preforms a basic rebase command, keeping to the exact image the user is on.

KyleGospo:
@ZachrielForever can you revert the change to hardware setup? Then I'm ready to +1 this

ZachrielForever:
> @ZachrielForever can you revert the change to hardware setup? Then I'm ready to +1 this

Turns out im still a moron at times. Change was reverted, should be good. Sorry for the issue.

KyleGospo:
I think you merged main into this, can you also remove the two commits that aren't yours from this PR?

PR #4065, opened 2026-02-02
feat: remove enable-supergfxctl ujust

Synby:
one line command doesn't need a ujust

related: https://github.com/ublue-os/docs.bazzite.gg/pull/362


Xarishark:
instead of removing the just, it would be better to copy one of the other just scripts and converting this to an enable disable menu and changing it to `ujust toggle-supergfx`.
Also just recipes are there for ease of use. Much easier for a user to open terminal and write ujust * instead of remembering each one line commant especially a service stacks like systemctl *.
So instead of removing it I would expand on the commands for ease of use for the general user. Most laptops come with nvidia chips behind a mux switch.

PR #4076, opened 2026-02-03
feat: dankpods proofing attempt#2

stellaberrant:
- liveiso: show a warning dialog if the user has nvidia and is trying to install an incorrect image, such as Nvidia card and AMD image, or old Nvidia card/ new Nvidia image
- show a different warning if the user has an unsupported (Kepler or earlier) GPU with a link to the documentation (locally served)
- show another warning about unrecognized Nvidia GPU, recommending not to install Bazzite
- three buttons: 
-- install anyway
-- power off
-- show gpu information (mostly for support requests)
- reduce the number of popups by 1 by removing the separate Nvidia driver warning and making the warning in the welcome prompt more clear

Nvidia GPU detection based on information from lspci and architecture codenames. image detection based on podman (which only works in the live iso)
-->


KyleGospo (dismissed):

KyleGospo:
Please pull out the revert commit from your branch, it's one of 4 in this PR.

KyleGospo:
+1 otherwise, thanks!

KyleGospo:
Disregard main is good per Zeglius, +1

installer/titanoboa_hook_postrootfs.sh
```diff
@@ -445,6 +449,163 @@ if [[ ! -d /sys/firmware/efi ]]; then
...
+fi
+}
+
+detect_nvidia
+result=$?
+if [ $result -eq 0 ] || [ $result -eq 1 ] || [ $result -eq 124 ]
+then
+  echo 'launch welcome dialog'
+    welcome_dialog
+fi
```
tulilirockz: Lets export this script into a `.sh` script close to the titanoboa hook instead of making it be inline, this kills readability on that file

stellaberrant: /usr/libexec?

PR #4122, opened 2026-02-10
chore(announcements): Add new sunshine service announcement

Zeglius:


KyleGospo:
I can adjust this message, but let's adjust this slightly. Disabling sunshine.service is trivial for us to do, and we can instruct people using it to re-enable it using the ujust.

PR #4144, opened 2026-02-12
fix(liveiso): disable dnf kernel protection in initramfs build

Zeglius:


KyleGospo:
@Zeglius @Xarishark you two have been cranking out these PRs like a fire hose, just wanted to say I really appreciate it. Good work here.

PR #4168, opened 2026-02-18
Add a service to automatically compile and install the sc0710 driver

kressb-gitlab:
Add a service to automatically pull the source for the Elgago sc0710 driver and compile it.  Intended to run on boot, it will only recompile the driver if the kernel version and the driver version differ.  Basically a mini DKMS.

The service is shipped but not enabled, idea being a user could enable it if they have this hardware.


KyleGospo:
Sorry this is a no, this must be implemented the proper way in akmods.

I appreciate the PR and will look at bringing it there as an akmod in the future.

PR #4170, opened 2026-02-18
feat: re-add konsole to the image

dreamyukii:
re-add Konsole to the image when Konsole got container feature


stellaberrant:
do we know when new Konsole is coming? I see the PRs have been merged but i'm not sure when we get a new release with that

PR #4176, opened 2026-02-19
Change how we configure default pins in Plasma for 6.6

kressb-gitlab:
Fedora updated to Plasma 6.6 breaking how we do our default pins.

Removed the (no longer functional) 6.5 mechanism as it was breaking the build.  Added a new 6.6 mechanism in the form of javascript that Plasma Shell will run once.  The script checks to see if "loaders" is defined in any panel and if they are not sets an explicit default list.

Plasma 6.5 users that have set their pins will not be affect.  Users that have not set them will get the default Bazzite list.
New users or fresh installs will also get the default Bazzite list.

The one case that isn't as nice any more is an explicit "Add Default Panel" in Plasma Shell.  This, unfortunately still gets you the Plasma default set of pins.  This could possibly be improved in a future PR.


KyleGospo:
This is great, thanks for fixing this

KyleGospo (approved):

PR #4207, opened 2026-02-24
fix(virt): dont use legacy libvirt daemon

### LINKED ISSUE #4086: Issues due to use of legacy monolithic libvirtd.service instead of modular components.
### Describe the bug

The "Enable Virtualization" action within `setup-virtualization` ujust script works by using a oneshot service to enable and start the `libvirtd.service` unit on the next reboot.  This is a legacy means of running libvirt on modern systems and I noticed it being the root of a few issues I've ran into trying to use virtualization on a new bazzite install.

Since Fedora 35, the documented way of enabling libvirt is through the modular service components.

Ref: 
- https://libvirt.org/daemons.html
- https://docs.fedoraproject.org/en-US/fedora-server/virtualization/installation/#_activation_and_startup_with_fedora_34
- https://docs.fedoraproject.org/en-US/fedora-server/virtualization/installation/#_activation_and_startup_with_fedora_35_and_up

A couple of the problems I had which I attributed to this:

- QEMU/KVM user sessions were not accessible. Still not entirely sure why, but it had something to do with the service socket not being reachable.
- The one shot service created by `setup-virtualization`, `bazzite-libvirtd-setup.service`, was not successfully enabling the libvirtd service on reboot, but would run it only that one time. This resulted in libvirtd running on that first reboot, but not on subsequent startups. I would have to manually run `systemctl start libvirtd.service`. 
- `libvirtd.service` could effectively not be enabled, because a number of the modular services that are enabled by default have `Conflicts=libvirtd.service` in their unit definition.

The libvirt service would be enabled with the first boot after running the `setup-virtualization` ujust and run reliably during subsequent boots.

QEMU/KVM System and user sessions would be accessible through virt-manager/virsh.

### Session Logs

```shell
-
```

### Hardware

-

This is how I went about fixing my install to disable/mask the monolithic libvirtd.service, and enable the modular service components. The `setup-virtualization` ujust should probably updated to do similarly.

Disable monolithic services
```
systemctl disable --now libvirtd.service
systemctl disable --now libvirtd{,-ro,-admin,-tcp,-tls}.socket
```

Mask the monolithic services
```
systemctl mask libvirtd.service
systemctl mask libvirtd{,-ro,-admin,-tcp,-tls}.socket
```

Enable the modular services
```
for drv in qemu interface network nodedev nwfilter secret storage
  do
    systemctl unmask virt${drv}d.service
    systemctl unmask virt${drv}d{,-ro,-admin}.socket
    systemctl enable virt${drv}d.service
    systemctl enable virt${drv}d{,-ro,-admin}.socket
  done
```

Start the sockets for the modular services
```
for drv in qemu network nodedev nwfilter secret storage
  do
    systemctl start virt${drv}d{,-ro,-admin}.socket
  done
```

Zeglius:
Fixes #4086

### DIFF
--- a/system_files/desktop/shared/usr/lib/systemd/system/bazzite-libvirtd-setup.service
+++ b/system_files/desktop/shared/usr/lib/systemd/system/bazzite-libvirtd-setup.service
@@ -5,8 +5,6 @@ ConditionPathExists=/usr/lib/systemd/system/libvirtd.service
 
 [Service]
 Type=oneshot
-# TODO: Rewrite this whenever systemd allows to queue ephemeral commands for next boot without modifying kernel args
-ExecStart=/usr/bin/bash -c "systemctl enable --now libvirtd; restorecon -Rv /var/log/libvirt /var/lib/libvirt; systemctl disable %n"
-
-[Install]
-WantedBy=multi-user.target
+# See https://github.com/ublue-os/bazzite/issues/4086#issue-3904621119
+# We dont need to enable modular services manually, these seem to come preenabled when installing qemu and libvirt.
+ExecStart=/usr/bin/bash -c "systemctl disable --now libvirtd.service; systemctl disable --now libvirtd{,-ro,-admin,-tcp,-tls}.socket"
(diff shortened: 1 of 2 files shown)


Zeglius:
You might know better than I do regarding this @HikariKnight

PR #4216, opened 2026-02-25
Change from legacy_rechunker to the rpm-ostree rechunker

kressb-gitlab:
The legacy rechunker does three things:

1) Copies entries from /etc/passwd and /etc/group into their corresponding files in /usr/lib.  Replace this by doing it in build_files/finalize.  As an added bonus it means you can use locally build images that haven't been rechunked for a faster dev/test cycle and they don't break from missing users.

2) The actual rechunk itself.  This is now done by rpm-ostree.

3) Removes the base image labels and adds the Bazzite ones.  The workflow now does this.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -348,6 +348,10 @@ jobs:
             sudo buildah config --label "$line" "$container"
           done <<< "$IMAGE_LABELS"
 
+          filter='{ "packages": [inputs | split(" ") | select(length==2) | { (.[0]): .[1] }] | sort_by(keys) | add }'
+          manifest=$(sudo podman run --rm "$img" rpm -qa --qf '%{NAME} %{VERSION}\n' | jq -Rnc "$filter")
+          sudo buildah config --label "dev.hhd.rechunk.info=$manifest" $container
+
           sudo buildah commit --identity-label=false --rm $container $img
 
           echo "version=$version" >> "$GITHUB_OUTPUT"


kressb-gitlab:
rpm-ostree compose build-chunked-oci also supports a --reference referring to the last iteration of an image to minimize diffs.  Using this unfortunately crashes rpm-ostree hard.  It can be relatively easily added to this workflow when it's fixed.

kressb-gitlab:
There is some interest in getting rid of /lib/group and /lib/passwd.  But that is a seperate issue to changing the rechunker, and for now since the legacy_rechunker did it, and Bazzite images expect it, this change does it as well.

KyleGospo:
I think we should adopt chunkah instead or at least look into switching. RH is pushing that

tulilirockz:
I think its still valuable to move off of the legacy rechunker because there are a few migration pains when doing so. Chunkah itself has some bugs that make it a bit funky to use on our images right now so I think we could move later (should be a lot easier after this migration.)

kressb-gitlab:
Summary of the current state of chunkah research:

Chunkah breaks the association between the ostree data and the manifest and produces images rpm-ostree will not accept.
Stripping the ostree information from the container and turning it into a plain OCI container results in an image that WILL install, but corrupts the ostree repo somehow.
This is not specific to chunkah, installing ANY plain OCI container image seems to corrupt the repo in the same way.

But these together and at the moment I see no way to actually USE an image rechunked with chunkah.

Per discord discussion, marking this for inclusion for testing, possibly in unstable branch.

p5:
> Chunkah itself has some bugs that make it a bit funky to use on our images right now

@tulilirockz please could you point me to the issues you're facing?

kressb-gitlab:
You might be talking about the fact that it chokes on the cosign and tailscale packages.  This is a bug in the rpm-qa crate that's fixed.  The latest release of chunkah hasn't been updated to use the fixed rpm-qa crate yet.  But the dev version has.  

This is independent of actually getting an image usuable by rpm-ostree though.

p5:
> This is a bug in the rpm-qa crate that's fixed

I filed that bug :)

I'm just wanting to make sure everything else is filed so we can work on it

renner0e:
I want to point out that the changelog generation currently depends on the labels set by the rechunker.

In Aurora we have been using SBOMs for our changelogs since the beginning of the year. [It broke recently due to some change in recor?](https://github.com/ublue-os/aurora/pull/1816)

A similar change could be done before this is merged to ease the transition.

Lumaeris:
While I'm in favor of this change I really want to point out one more thing.

`build-chunked-oci` only assigns RPM packages as individual layers but not external files that might appear from different sources, say Homebrew tarball or Steam bootstrap. Since those files aren't assigned in any way then they'll end up in `unpackaged` layer so you'd need to re-download them over again during the update (screenshot is not mine):

image

Luckily there's a solution. `build-chunked-oci` accepts `user.component` attribute to individual files and maybe even directories, I haven't tried the latter. It is described in rpm-ostree's docs: https://coreos.github.io/rpm-ostree/build-chunked-oci/#assigning-files-to-specific-layers

So you can add `setfattr -n user.component -v "homebrew" /usr/share/homebrew.tar.zst` at any point (after getting it from Brew OCI image of course) and `setfattr -n user.component -v "steam-bootstrap" /usr/share/gamescope-session-plus/bootstrap_steam.tar.gz` after getting Steam bootstrap from Nobara folks. Just some examples from the top of my head :slightly_smiling_face: 

If done correctly, the rechunker will mention new layers in logs (example from my personal repo): https://github.com/Lumaeris/ophiros/actions/runs/22088308380/job/63828034735#step:9:113

By the way, `user.component` is accepted by Chunkah as well, so whenever it would be appropriate to move to it, all is needed is to change workflow file and nothing else.

KyleGospo:
> While I'm in favor of this change I really want to point out one more thing.
> 
> `build-chunked-oci` only assigns RPM packages as individual layers but not external files that might appear from different sources, say Homebrew tarball or Steam bootstrap. Since those files aren't assigned in any way then they'll end up in `unpackaged` layer so you'd need to re-download them over again during the update (screenshot is not mine):
> image
> 
> Luckily there's a solution. `build-chunked-oci` accepts `user.component` attribute to individual files and maybe even directories, I haven't tried the latter. It is described in rpm-ostree's docs: https://coreos.github.io/rpm-ostree/build-chunked-oci/#assigning-files-to-specific-layers
> 
> So you can add `setfattr -n user.component -v "homebrew" /usr/share/homebrew.tar.zst` at any point (after getting it from Brew OCI image of course) and `setfattr -n user.component -v "steam-bootstrap" /usr/share/gamescope-session-plus/bootstrap_steam.tar.gz` after getting Steam bootstrap from Nobara folks. Just some examples from the top of my head 🙂
> 
> If done correctly, the rechunker will mention new layers in logs (example from my personal repo): https://github.com/Lumaeris/ophiros/actions/runs/22088308380/job/63828034735#step:9:113
> 
> By the way, `user.component` is accepted by Chunkah as well, so whenever it would be appropriate to move to it, all is needed is to change workflow file and nothing else.

This is done now for all files we explicitly download.

Files in the repo itself remain to be done, wallpapers are what immediately come to mind.

kressb-gitlab:
Added attrs for homebrew and the ublue-os docs, we can add more as we go.

kressb-gitlab:
Added some logic to generate the OCI property that changelog.py is expecting so this PR doesn't result in losing that functionality.

We could also change how we generate changelogs if we want, but that could be a separate PR, trying not to boil the ocean in this one.

PR #4224, opened 2026-02-26
feat: re-arrange categories in the Bazzite Portal

nicknamenamenick:
This rearranges the different tabs on the Bazzite Portal so it goes:

```
Setting Up Bazzite --> Customization and Tweaks --> Media Applications --> Brew --> brh GUI
```
This just makes more sense to new users in my opinion, but anyone can argue that it doesn't.  Also, this PR was edited on the Github website and was NOT tested on Bazzite. Please test it properly before merging if these changes are accepted.

Cc: @Xarishark

### DIFF
--- a/system_files/desktop/shared/usr/share/yafti/yafti.yml
+++ b/system_files/desktop/shared/usr/share/yafti/yafti.yml
@@ -1,44 +1,11 @@
 title: Bazzite Portal
-screens:
-  - title: "System Management"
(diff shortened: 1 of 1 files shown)


Xarishark:
Ack by me

eramdam:
It seems this broke Portal on `testing` because the identation on line 56 is wrong 
image
image

PR #4232, opened 2026-02-27
Remove references to konsole as it is not in main yet

kressb-gitlab:
Remove references to konsole as it is not in main yet


KyleGospo:
I merged the konsole change, it should be in the next main along side Ptyxis so people can begin to transition if they have custom Ptyxis configs.

PR #4236, opened 2026-02-28
Remove Ptyxis Terminal as default terminal

Xarishark:
This should be merged when the change actually happens.


thisischrys:
org.gnome.Ptyxis.desktop is still set as "default" here
https://github.com/ublue-os/bazzite/blob/main/system_files/desktop/kinoite/usr/share/kde-settings/kde-profile/default/xdg/kicker-extra-favoritesrc

Xarishark:
Yes this should not be merged yet. Still its just a readme change. so small thing really.

PR #4244, opened 2026-03-01
fix(portal): fix indentation in yafni.yml

eramdam:
I don't have a good way of testing right now but as I can tell, the indentation is correct which should fix the crash introduced by https://github.com/ublue-os/bazzite/pull/4224


nicknamenamenick:
Just wanted to say thank you for the fix. Next time I will draft the PR and undraft it AFTER I test it. 😆

PR #4250, opened 2026-03-02
Remove Ctrl-Alt-T shortcut for Ptyxis

kressb-gitlab:
Remove the additions to the container file that force Ptyxis to use the Ctrl-Alt-T shortcut.  This conflicts with the Konsole one.


kressb-gitlab:
Any user that rebased to any image that had both claiming the same shortcut will have some garbage in ~/.config/kglobal* that has to be cleaned out.

Any user that never saw images that claimed both shortcuts will have a seemless change.

PR #4251, opened 2026-03-02
fix: remove leftover initramfs

renner0e:
The initramfs from the base image is not properly removed, makes the image about 200MiB smaller.

Using rpm this way is better than dnf because it doesn't remove the virtualbox-guest-additions package like dnf. I don't think this was an intentional change. It's actually faster to use rpm in a for loop here instead of doing it all at once for whatever reason. This has been used like this in Aurorafin for quite a while.

### DIFF
--- a/build_files/install-kernel
+++ b/build_files/install-kernel
@@ -15,7 +15,12 @@ printf '%s\n' '#!/bin/sh' 'exit 0' > 50-dracut.install
 chmod +x  05-rpmostree.install 50-dracut.install
 popd
 
-dnf5 -y remove --no-autoremove kernel kernel-core kernel-modules kernel-modules-core kernel-modules-extra kernel-tools kernel-tools-libs
+for pkg in kernel kernel{-core,-modules,-modules-core,-modules-extra,-tools-libs,-tools}; do
+    rpm --erase "${pkg}" --nodeps
+done
+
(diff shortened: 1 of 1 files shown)


renner0e:
yeah this worked pretty much like I expected:

```
bash-5.3# ls /usr/lib/modules
6.17.7-ba25.fc43.x86_64

bash-5.3# rpm -qa | grep -i kernel
kernel-core-6.17.7-ba25.fc43.x86_64
kernel-modules-core-6.17.7-ba25.fc43.x86_64
kernel-modules-6.17.7-ba25.fc43.x86_64
kernel-modules-extra-6.17.7-ba25.fc43.x86_64
kernel-headers-6.18.3-200.fc43.x86_64
kernel-devel-6.17.7-ba25.fc43.x86_64
kernel-tools-libs-6.17.7-ba25.fc43.x86_64
kernel-tools-6.17.7-ba25.fc43.x86_64
kernel-devel-matched-6.17.7-ba25.fc43.x86_64
kernel-6.17.7-ba25.fc43.x86_64
kernel-modules-akmods-6.17.7-ba25.fc43.x86_64
kernel-common-6.17.7-ba25.fc43.x86_64
kernel-srpm-macros-1.0-27.fc43.noarch
```

PR #4258, opened 2026-03-03
fix!: temporary fix for recursive drkonqi coredump

tulilirockz:
We've got some reports about 100% disk usage going on because of DrKonqi coredumping and launching when it coredumps so it coredumps again (...)

This should be a _TEMPORARY!_ fix where we just removedrkonqi from the image as a mitigation step.


tulilirockz:
Fix already got pushed by @KyleGospo. Thank you!

PR #4259, opened 2026-03-03
chore(bazaar): add russian strings to jetbrains hook

Lumaeris:
Adding russian strings to new JetBrains hook in Bazaar config as described in https://github.com/kolunmi/bazaar/blob/main/docs/overview.md#translations-in-yaml-configs


ZachrielForever:
LGTM

PR #4266, opened 2026-03-03
Add zh-tw translation and update README links

rockleona:
Thank you for creating this project, I'm also one of the users as well, and I would like to stand out for the first time contribution to this project.

This pull request contains a `zh-tw` translation which is using Traditional Chinese with Taiwanese common words.

I believe Taiwanese is a big gamer community as well, from my perspective, we also have a lot of **bazzite users and promoters**  in the commuity. As for why `zh-tw` is needed, you can also check [a PR from famous FastAPI](https://github.com/fastapi/fastapi/pull/10505#issuecomment-1786540636) which explain the similarity and difference in Chinese (which is `zh`).


KyleGospo:
Thank you!!

PR #4279, opened 2026-03-05
Mask systemd-remount-fs.service as its nonsensical for bazzites context

Xarishark:
Mask the `systemd-remount-fs.service` to prevent it from being activated as it never worked under bazzite. Its out of the specs and scope of bazzite anyway and it will take care both current and new installs.


renner0e:
here is some more extra info on what this change actually is working around:
https://discussion.fedoraproject.org/t/root-mount-options-are-ignored-in-fedora-atomic-desktops-42-and-later/148562

PR #4287, opened 2026-03-05
Add users to initramfs to silence errors

kressb-gitlab:
Every boot Bazzite has a lot of alarming (to some users) errors in the journal from missing users and groups in initramfs.  This is because Bazzite reads users and groups from altfiles, and those files aren't in initramfs.

So add them.  Silences the errors from initramfs.

### DIFF
--- a/system_files/desktop/shared/usr/lib/dracut/dracut.conf.d/99-bazzite.conf
+++ b/system_files/desktop/shared/usr/lib/dracut/dracut.conf.d/99-bazzite.conf
@@ -1,0 +1,1 @@
+install_items+=" /etc/nsswitch.conf /usr/lib/passwd /usr/lib/group "


tulilirockz (commented):
Imho its valuable to let upstream deal with these kinds of issues, if this happens on bazzite it most likely also happen on Silverblue/Kinoite, and if thats the case then this logic should be there instead.

kressb-gitlab:
No, it doesn't happen upstream.  This is a result of how we do our users with NSS altfiles.

Kinoite and Silverblue do not do this.

m2Giles:
> No, it doesn't happen upstream. This is a result of how we do our users with NSS altfiles.
> 
> Kinoite and Silverblue do not do this.

What is the difference between our usage and upstreams? My understanding is that silver blue has an empty /etc/group and a more or less empty /etc/passwd by default. Due to rechunk, we have everything in altfiles which should effectively be the default.

kressb-gitlab:
Oh, you're right this is a kinoite/silverblue thing.  I thought the altfiles thing was only Bazzite/bluefin/aurora.
I wonder if they'd want this...

PR #4309, opened 2026-03-07
Include release in changelogs

kressb-gitlab:
Changelogs are supposed to include the release (without the ".fc43" part).  Currently they do not.

Add it.


kressb-gitlab:
Tightened the regex to only remove "\.fc\d+$" given that not all rpms are fedora.

PR #4312, opened 2026-03-08
feat: rebrand bootloader entry from Fedora to Bazzite

richstokes:
The `build_files/image-info` script already extensively rebrands `/usr/lib/os-release` (changing `PRETTY_NAME`, `NAME`, `ID`, `BOOTLOADER_NAME`, etc.), but it doesn't touch `/etc/system-release`, which still contains `Fedora release XX (...)`.

This matters because `/etc/default/grub` on Fedora derives `GRUB_DISTRIBUTOR` from `system-release`:
```
GRUB_DISTRIBUTOR="$(sed 's, release .*$,,g' /etc/system-release)"
```

So when users run `ujust regenerate-grub` or otherwise invoke `grub2-mkconfig`, their boot entries show "Fedora" instead of "Bazzite". This is confusing — especially in dual-boot setups where users need to distinguish Bazzite from a stock Fedora install.

The installer ISO path (`titanoboa_hook_postrootfs.sh`) already rebrands `system-release`, but users who rebase to Bazzite (without going through the installer) don't get this fix.

This one-line `sed` in `image-info` replaces `Fedora` with the `IMAGE_PRETTY_NAME` ("Bazzite") in `/etc/system-release` during the image build, ensuring the bootloader is properly branded regardless of install method.


KyleGospo:
This is great, thank you!

PR #4322, opened 2026-03-08
fix(ci): make dnf cache work

renner0e:
Basically a follow-up of https://github.com/ublue-os/bazzite/commit/f39e268089417dde5bcbcb93dc8c9935886a43d3. This makes local builds way less
painful.

Setting keepcache=0 to revert back to the default so downstream images
don't get surprised all of a sudden :)

We have a 10GiB limit on cache per repo, we must be more efficient here
if we want to have cache on a main branch and testing branch. -deck
images are the fattest and derived from the normal builds so those are
the ones that are allowed to write to the cache. Everyone else just
reads the cache.

This means that we have 2 caches (per branch) with roughly 1.3GiB.

Making a separate cache for kde and gnome makes sense as there are some
specific packages that are only installed on each variant. Bringing
nvidia into the mix doesn't make sense as we get all the big nvidia rpms
from the akmods image anynway, the rest doesn't matter all that much.
The non nvidia builds are more robust to random breakage.

This is in my opinion the best compromise within the current
limitations.

If we ever ever need to reduce our cache size for whatever reason then
we can only use one cache for everything and then decide on one blessed
build that is allowed to write to the cache.

In Aurora we have realized that this actually doesn't work super well
for our usecase as we basically never build images off of the main
branch (where all other branches) get their cache from as every other
branch is and should be treated as untrusted. as we use the merge queue
for everything.
See: https://github.com/actions/cache/issues/1537
Related PR: https://github.com/ublue-os/aurora/pull/1761

Also did a little cleanup, the ostree container commit we don't need,
this is old stuff that can be replaced by bootc container lint (which
we do) and dnf5-plugins is already in the ci-test base image.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -256,7 +256,6 @@ RUN --mount=type=cache,dst=/var/cache \
         v4l-utils \
         yad \
         f3 \
-        pulseaudio-utils \
         lzip \
         p7zip \
         p7zip-plugins \


renner0e:
not sure if the we way do it in aurora would scale for bazzite as the cache for bazzite is probably way bigger and it's generally more difficult because of the bigger image matrix, we have a 10GiB limit per repo after all

renner0e:
this is not 100% representative and is heavily skewed by inconsistencies elsewhere in the pipeline and github runner jank but first results can be compared here:

[no cache](https://github.com/ublue-os/bazzite/actions/runs/22843776232/job/66255285443)

[with cache](https://github.com/ublue-os/bazzite/actions/runs/22846129698/job/66262871781?pr=4322)

raw build step comparisons:

bazzite went from 13m12s to 8m42
bazzite-gnome-nvidia-open went from 15m 7s to 9m 21s
bazzite-deck-nvidia went from 19m 39s to 12m 10s

I have no idea why the difference is so big here, this doesn't do as much for aurora

KyleGospo:
This is awesome, thank you so much!

PR #4334, opened 2026-03-09
fix(sunshine): switch off of sunshine-kms

Zeglius:


moi952:
Hi, sorry to ask, but it seems to me that Sunshine KMS was implemented a few weeks ago, and I see that it's no longer being used.

I don't know exactly why, but what's the reason? 
Did the Sunshine developers change the way the service is activated?

It's simply to try to understand

Thanks in advance for your reply 😊

KyleGospo:
You'd have to ask the sunshine developers what they're doing.

Our problem is they haven't pushed a single stable build for Fedora 43, which forces us to use their beta repo. The team is aware and it's a single button press to fix it. Your guess is as good as mine as to why it hasn't been done.

PVermeer:
Lot's of massive code changes in sunshine again...
See: https://github.com/LizardByte/Sunshine/pull/4779

A small description explaining PR's would be helpful.

There are now two services:
 - `app-dev.lizardbyte.app.Sunshine.service`: unprivileged (new portal grab)
 - `sunshine.service`: old kms grab

Imo this PR should have included a change back from `sunshine-kms.service` to `sunshine.service`, if it was enabled. It is weird that there are often massive refactors in the sunshine project compared to actual fixes / features. It seems a bit overengineerd. Maybe better to have a own COPR for bazzite for a stable sunshine.

PR #4344, opened 2026-03-10
Add section for Copr packages we maintain

Xarishark:
Introduce a new section detailing the Copr packages maintained by the project, organized by source for clarity. 
This will help devs and maintainers know what package changes need to be notified to Kyle so he can initiate a rebuild so the changes will appear on next Bazzite build


noelmiller:
Is there a way we can pull the list of packages programmatically? My fear is this list won't stay up to date in the future.

Xarishark:
Not sure this can be done in the future tbh but as of right now its much better than nothing as I personally had no idea that the kde presets file was an RPM that was getting installed from our copr. I patched the Update kcm and I found out 3 months later that the patch was never actually applied to bazzite because of the whole copr thing.

renner0e:
> Is there a way we can pull the list of packages programmatically? My fear is this list won't stay up to date in the future.

not sure what the best way would be to do this automatically but we should probably be having SBOMs anyway and they could also be used for this purpose *somehow*.

No information regarding this is probably better than outdated wrong information.

Xarishark:
should we also enable the Wiki tab at some point for a SBOMs because the docs website is not the best place for the people that need this info tbh

also @renner0e if you are ok with this it needs one ack.

Xarishark:
@KyleGospo should I close this then?

PR #4348, opened 2026-03-11
feat(Portal): Add GUI autologin fix for KDE users after rebase from gamemode to a Desktop image

Xarishark:
Implement a GUI fix for user so they can start using the autologin option again. This is a new button in the portal.


KyleGospo:
We should be able to automatically fix this via the hardware setup service

Xarishark:
Will this fix the current installs the experience this bug? I feel like we still would be benefit from a gui button for users to be able to find.

KyleGospo:
Yes, just a matter of checking that we're not on Deck image and deleting a file if it exists

Xarishark:
Where is this script at? I would be happy to implement the automation there instead then

PR #4359, opened 2026-03-12
feat(gaming): add global Steam Deck override toggle

moi952:
- Added `ujust toggle-global-steamdeck` command to enable/disable Steam Deck mode globally
- When enabled, `SteamDeck=0` is set to force games to run in PC mode
- Integrated Steam Deck toggle into Yafti portal under "Customizations and Tweaks"

### DIFF
--- a/system_files/desktop/shared/usr/share/ublue-os/just/80-bazzite.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/80-bazzite.just
@@ -722,6 +722,46 @@ toggle-global-dlss ACTION="":
             ;;
     esac
 
+# Toggle global SteamDeck environment override
+[group("gaming")]
+toggle-global-steamdeck ACTION="":
+    #!/usr/bin/bash
+    set -euo pipefail
+    source /usr/lib/ujust/ujust.sh
+
+    FILE="$HOME/.config/environment.d/99-steamdeck-override.conf"
+
+    CURRENT="Disabled"
+    if [[ -f "$FILE" ]]; then
+        CURRENT="Enabled"
+    fi
+
+    OPTION="{{ ACTION }}"
+    if [[ -z "$OPTION" ]]; then
+        echo -e "${bold}Global Steam Deck Override:${normal} $CURRENT"
+        echo ""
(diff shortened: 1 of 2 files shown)


KyleGospo:
We really want to disincentivize this as must as possible, this breaks all kinds of things and should only be done for games that abuse this environment variable to do things Valve already explicitly says not to do.

moi952:
> We really want to disincentivize this as must as possible, this breaks all kinds of things and should only be done for games that abuse this environment variable to do things Valve already explicitly says not to do.

Ah, I understand. I use it for some games. If I set SteamDeck=0, I can enable FG DLSS. In Ratchet & Clank and The Last of Us, it's great, but in Horizon Zero Dawn Remastered, I get a lot of stuttering, and even my GPU only reaches 85-90% utilization for 100fps (with AMD or Nvidia FG). Whereas without SteamDeck=0, I get 120fps without stuttering at 90-95% utilization. 

It's strange. But despite that, I find it simpler to have SteamDeck=0 globally and set it=1 for the few games that cause problems (so far, only one).

image

xXJSONDeruloXx:
@KyleGospo i think you're mixing this up with the deck alias steam launch changes, this is a per game env var, and not destructive to steams state. 

That said, I did similar previously and found that as a global override this didn't work properly as this is overridden by steams own -steampal and other similar args, unlike the PROTON_FSR4_UPGRADE cousin. 

If there's an approach that works then +1 but I don't think the current will trigger in the correct order to affect the game init.

HikariKnight:
> @KyleGospo i think you're mixing this up with the deck alias steam launch changes, this is a per game env var, and not destructive to steams state.
> 
> That said, I did similar previously and found that as a global override this didn't work properly as this is overridden by steams own -steampal and other similar args, unlike the PROTON_FSR4_UPGRADE cousin.
> 
> If there's an approach that works then +1 but I don't think the current will trigger in the correct order to affect the game init.

i think the only method that works is applying it through launch arguments, at that point you can have scopebuddy handle it or use the `sd0` command in the launch options, neither is an easy toggle but both are less to type than `SteamDeck=0 %command%`

PR #4380, opened 2026-03-14
liveiso: add nvidia-deck warning

stellaberrant:
- added nvidia-deck image detection warning
- added image information and Discord server link to third button 
-some refactoring and general cleanup

Screenshot_20260314_181907
Screenshot_20260314_181859


Zeglius:
Will say there is qrencode available in the live environment, just to let you know. I would imagine an hyperlink wouldn't be as helpful as would require connecting to the internet.

PR #4401, opened 2026-03-16
feat: prepare branding for plasma-setup

renner0e:
- https://invent.kde.org/plasma/plasma-setup/-/issues/72
- https://github.com/ublue-os/packages/pull/1191
- https://github.com/ublue-os/packages/pull/1199

[Aurora PR](https://github.com/ublue-os/aurora/pull/1933), which has to be implemented here as well

image

PR #4406, opened 2026-03-17
feat: utilize plasma setup and gnome initial setup for desktop

Lumaeris:
Draft for now since Plasma Setup should appear on F44 release.

I'm also replacing start-here-symbolic.svg which is being used by GNOME Initial setup, it does not support coloring

Screenshot From 2026-03-17 20-43-54

Related: https://github.com/ublue-os/bazzite/pull/4401


EPOCHvoyager:
@Zeglius Couldn't this work be used to simplify the installer?

Lumaeris:
It will except for deck users, Steam expects already existing user account to start itself

Lumaeris:
I have an idea on how to make Plasma Setup and GNOME Initial Setup work for deck users but in order to do that Bazzite needs to switch off of SDDM for session switching to something else. As of recent SteamOS Preview build Steam now uses steamos-manager for session switching which may no longer using SDDM under the hood for this purpose

Lumaeris:
Anyway, I'm closing this PR since it doesn't really do anything hard that anyone else can do freely. I'll focus on something else

PR #4422, opened 2026-03-19
fix(liveiso): use older pin replacement for kde

Lumaeris:
New solution apparently doesn't work right now

image


renner0e:
This config file you reference here got removed with 6.6, which is the reason why we had to change all this.

https://bugs.kde.org/show_bug.cgi?id=511560

PR #4424, opened 2026-03-19
chore(ci): rechunk image only after applying labels

Lumaeris:
I didn't liked that it added 1 custom layer on top of rechunked image so I moved this step further

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -393,7 +393,7 @@ jobs:
           sudo buildah commit --identity-label=false --rm $container raw-img
 
           sudo podman run --rm --privileged --volume /var/lib/containers:/var/lib/containers \
-              quay.io/centos-bootc/centos-bootc:stream10 \
+              localhost/raw-img \
               rpm-ostree compose build-chunked-oci \
               --bootc --max-layers 96 --format-version 2 \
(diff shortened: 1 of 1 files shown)


kressb-gitlab:
The old version cleared all the labels before running the rechunker.  The new version doesn't clear them.
This means any labels present from the base image (that aren't rewritten) will still be present on the output image.  Unclear how much of an issue this is, but it's a semantic change.

Lumaeris:
> The old version cleared all the labels before running the rechunker. The new version doesn't clear them. This means any labels present from the base image (that aren't rewritten) will still be present on the output image. Unclear how much of an issue this is, but it's a semantic change.

Oh that's why that line was there, let me add it back

kressb-gitlab:
I think removing line 376 there was an issue as you're continuing to use $container after you rmed it.

Lumaeris:
I moved it to 345 line

kressb-gitlab:
345 creates it.
347 deletes it.
383 uses it without recreating it.

Lumaeris:
So it has to be used again? Sure, I'll add it one more time

bsherman (approved):

noelmiller (approved):

PR #4435, opened 2026-03-20
Warn if the user is about to run into error 1

kressb-gitlab:
If the user tries to install Bazzite reusing an EFI that happens to contain an old (possibly not even used) copy of Grub, anaconda fails with the infamous error 1.  The current installer will protect the user from a copy of grub installed with the subdirectory "fedora" but not any others. 

This leads to users that had previous tried say Linux Mint and are now installing Bazzite to get this rather bad failure.

So when we boot the installer, let's take a quick peek at any EFI partitions and warn the user if GRUB is lurking there.
We specifically ignore "fedora" and "boot" since the ISO has that covered.

### DIFF
--- a/installer/system_files/shared/usr/bin/on_gui_login.sh
+++ b/installer/system_files/shared/usr/bin/on_gui_login.sh
@@ -164,6 +164,40 @@ nvidia_hardware_helper() {
         done
     fi
 }
+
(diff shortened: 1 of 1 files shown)


kressb-gitlab:
image

stellaberrant:
:100: % approval from me

PR #4441, opened 2026-03-21
feat: add a spec for gnome initial setup

Lumaeris:
Instead of replacing GNOME icon in start-here-symbolic (like in #4406) let's just add Bazzite to the list of distros in language page. That's just a Fedora spec for F44 with added Patch1 variable


renner0e:
I'm not sure this is the best way to handle this. This patch could be upstreamed to Gnome, although I'm not sure if they accept this for every distro under the sun. It's definitely worth a try.

Another alternative would be to change the logo name to fedora-logo-icon which is probably less friction in the long run.

I think both of these solutions are way less of a maintenance burden compared to patching and maintaining this forever.

PR #4450, opened 2026-03-22
feat: persist NTFS/exFAT notification dismissal across reboots

dylanmtaylor:
## Summary

Previously, clicking "I understand" on the NTFS/exFAT warning only incremented an in-memory counter that reset on every service restart. Users had to dismiss the notification 5+ times per session before they could stop it, and it would return after every reboot.

## Changes

- On clicking "I understand", a marker file is written to `$XDG_DATA_HOME/bazzite/ntfs_nag_dismissed` and the service stops immediately
- On subsequent service starts, the script checks for this file and exits early
- Removed the in-memory counter and all `firstbutton_ignore` translation strings (no longer needed)

## How to re-enable

Delete `~/.local/share/bazzite/ntfs_nag_dismissed` to see the warning again.

### DIFF
--- a/system_files/desktop/shared/usr/libexec/ntfs_exfat_monitor_script
+++ b/system_files/desktop/shared/usr/libexec/ntfs_exfat_monitor_script
@@ -1,16 +1,20 @@
 #!/usr/bin/env bash
 
+#check if previously dismissed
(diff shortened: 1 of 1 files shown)


KyleGospo:
This notification is intentionally annoying

PR #4468, opened 2026-03-25
feat(ujust): add toggle-reisub

ykshek:
REISUB is disabled by default on fedora, a ujust to easily re-enable it would be helpful.


ykshek:
image
image
image

ykshek:
Oh that was slightly messy
at least I got my commits signed now

ykshek:
Also add an entry for bazzite portal

PR #4477, opened 2026-03-26
feat: add back openrazer-daemon

dreamyukii:
add openrazer-daemon to the base image and remove openrazer just recipe


Cm4nXD:
It conflicts with openRGB. it was a pain to manually remove this when it was in the base image. having a ujust recipe to install this is fine.

PR #4478, opened 2026-03-26
feat: add toolbox override

Lumaeris:
Adds a simple warning similar to dnf override

image

I added it to override dir so it doesn't actually get removed on dnf remove transaction


KyleGospo:
This prevents layering toolbox back in which I imagine some users did, since this is a CLI only utility and not exactly a common one I'm not sure this needed.

PR #4507, opened 2026-03-30
fix: OCR in unstable not working

stellaberrant:
for some reason for the OCR to work in unstable, KDE now needs the tesseract-devel package


stellaberrant:
> This was always needed for OCR, maybe this should go straight to main instead?
> 
> [ublue-os/aurora#1788](https://github.com/ublue-os/aurora/pull/1788)

I agree… but for some reason it was working until recently

PR #4515, opened 2026-03-31
feat(sunshine): added sunshine spec + ujust changes

PVermeer:
# Motivation
Due to numerous ignored issues about their stable repo I created my own COPR test repo to see why stable isn't built. I found numerous issues that would prevent a solid build on all targets. I suggested a new build and package setup for COPR to the sunshine maintainer where stable could be build on all targets without issues. The maintainer however has no intention of rebuilding stable every 6 months for the major Fedora update. So this is a dead end.

The only solution, to keep Sunshine on Bazzite, is with an external build COPR so I did. 

# New COPR
I transfered my work to a new [COPR](https://copr.fedorainfracloud.org/coprs/pvermeer/sunshine/monitor) and this [Github repo](https://github.com/PVermeer/copr_sunshine).

## Packages
- [sunshine](https://copr.fedorainfracloud.org/coprs/pvermeer/sunshine/package/sunshine/): This is the stable release and my CI will notify me if there is a new release or Fedora has been forked so I can review the changes and warn / update Bazzite before releasing.
- [sunshine-beta](https://copr.fedorainfracloud.org/coprs/pvermeer/sunshine/package/sunshine-beta/): CI will check once a week for beta updates and update the beta package. No warnings here since it's rolling.

## Testing
Vaapi encoder has been tested by me and has no issues.
Nvidia encoder has been tested by me and has no issues (cuda).
Vulkan encoder has been tested by me and has no issues (mesa / nvidia builds).

~~### Nvidia needs testing!~~
~~Nvidia I could not test since I have no recent hardware for it. This package uses a different strategy to build the cuda parts to support encoding on Nvidia hardware. I don't expect any issue since the code compiles fine, but I would like for someone to test it before releasing this PR.~~

# Nvidia
Right now i'm using the latest nvcc cuda compilers because it's easy to maintain. The cuda parts of the code will only work with a compatible driver / hardware. Right now it must be 580 or higher. I know Bazzite is using recent drivers so this would be no issue.

~~However, I would like to know if there are Bazzite images that use a lower nvidia driver version (<580) and/or supports the GTX 10** or lower?~~

~~If so, I will set the cuda compiler to a lower version. The cuda compiler is forwards compatible so it will support newer drivers but drops off older drivers (and hardware). The range it should currently support is RTX 20xx - RTX 50xx >. If the GTX 10xx is still supported on Bazzite I will switch it to a lower version.~~

# Changes with Sunshine COPR
Removed a lot of unnecessary code so it builds on all targets, including:
- Testing suites, code is already tested. Testing can cause issues in different chroots and is not needed for packaging.
- Compiler switching on targets because of gcc - cuda compatibility. This is fixed with a [conda package](https://anaconda.org/channels/conda-forge/packages/cuda-nvcc/overview) that has all the necessary compilers to compile the cuda files. Sunshine itself is build with the host compiler for max compatibility and the cuda part is linked with.
- Removed unneeded build packages.
- Removed all runtime dependencies. `rpmbuild` can sort this out automatically. This also reduced the total packages install size.

Removing all this complexity resulted in a reduction of 289 lines in the spec file.

# To do
- compile with -Werror on:  One dependency will give warnings but It's harmless. The next stable build will fix it or I will patch it soon.

# Changes to Bazzite
- Reverted the removal of LizardByte/Sunshine
- Updated to my copr: pvermeer/sunshine
- Updated install Sunshine to sunshine (no capital)
- Removed the `setcap`. The RPM is already doing this or does this not work when building the container...?
- Service is `sunshine.service` again

I've added my own copr here but I would be great if Bazzite can fork this, or create it's own copr builds using this spec. I intend to maintain this for as long as Sunshine doesn't build stable properly.

# Conclusion
I think it's a shame for Bazzite to loose Sunshine because of these issues. It's a great project and one of reasons I moved to Bazzite, long ago. I understand dealing with these issues can be frustrating. Issues arising with this build can now be directed to me. Also new stable builds will personally be checked by me before releasing.

https://copr.fedorainfracloud.org/coprs/pvermeer/sunshine
https://github.com/PVermeer/copr_sunshine


andy10115:
+1 from where i sit.  Probably better than the workarounds we've been using. I would think this means putting it back in image though.

KyleGospo:
@PVermeer Sorry for the long wait, just getting through the backlog. If you can fix the conflict I'll get this merged!

PVermeer:
Sorry for the close, I resetted this branch because rebasing/merging had too many conflicts and force pushing with no commits auto-closes it.

**Updates**:
Nvidia has been tested and it still works with older hardware. Also sunshine stable now has the vulkan api encoding paths, so this is a nice fallback.

**Questions before I can proceed**:
- Would this replace flatpak / brew (preferable imo)?
- If it needs to run alongside the current implementations, I think there is a binary conflict with brew? @ykshek

andy10115:
My vote would be to replace. Up to Kyle though.

ykshek:
@PVermeer 
> Would this replace flatpak / brew (preferable imo)?

Ideally flatpak or brew would be better as they do not block or slow down system updates. Given the current running issues on both of those methods, if there is an rpm to make it just work™, it would be very nice, and tbh the current way of managing install scripts and stuff for Sunshine(especially brew) in a massive bash script is kinda messy.

Though given the flatpak does work in most cases by using desktop portal or Kwin screencast, I'd

> If it needs to run alongside the current implementations, I think there is a binary conflict with brew? @ykshek

It should be fine, though I haven't gotten around to testing it. The brew package is very experimental and finicky to begin with, so I definitely support telling users to layer it through Bazzite Portal on `-deck` images

## Some more thoughts about this:

Maybe you could add the spec into `/spec_files/` under the repo so that it can go under [`ublue-os/bazzite`](https://copr.fedorainfracloud.org/coprs/ublue-os/bazzite/)?

Imo it should be a layerable package for users to add through the portal and not in-image, similar to `asusctl`, because that would allow the user to layer stable/beta versions at will. Ultimately, it is up to Kyle to decide on this.

andy10115:
If this is going to be maintained would it not be better to add it back into the image rather than layer it?

ykshek:
Adding the package back into the image would likely block installation of the sunshine beta package due to package conflicts*, and additionally bind every image to a specific version of sunshine(which was one of the additional reasons for sunshine to be thrown out, besides the package not being built for f44 and breaking numerous times)

*well technically I think it could be possible to override it with `rpm-ostree override remove`, but that is really finicky and I am sure no one wants that.

PVermeer:
> Adding the package back into the image would likely block installation of the sunshine beta package due to package conflicts*, and additionally bind every image to a specific version of sunshine(which was one of the additional reasons for sunshine to be thrown out, besides the package not being built for f44 and breaking numerous times)

Thanks, that was my concern by just adding it back to the image.

> *well technically I think it could be possible to override it with `rpm-ostree override remove`, but that is really finicky and I am sure no one wants that.

I think this is fine. Using (any) beta software will break things sooner or later. User should know how to fix issues like this if they want to use the beta version. Otherwise they can file an issue for the stable version.

Bazzite is about simplicity and it should just work. As you might know I'm not a big fan of all these `ujust` scripts that can break things and some of them are inherently insecure (running external scripts with sudo).

So there are there three options:
1. Leave as is with brew / flatpak, users can still layer the copr package (complex)
2. Add the layering to ujust with checks to ensure only 1 is installed (added complexity)
3. Add to the image and remove the other implementations (reduce complexity)

**My opinion**
From a technical perspective I would keep things simple and separated. If Flatpak / Brew / Rpm versions are causing issues, and thus add complexity to the project, create your own package where these issues are addressed and install that.

From a user perspective I would say I just want things to work. I don't want to worry about using using different versions. I do not like to run some elevated scripts to hopefully makes things work. This should already be done in a controlled way when building the image.

My preference would go to option 3.
I have removed a lot of complexity from the official Sunshine rpm build and it's now pretty much stable and hopefully only needs some dependency updates now and then.

andy10115:
Well, I'll defer to others judgement here, but I prefer not to layer when
possible, and only use the HTPC images. Flatpak will never work for how my
PC operates.

Layering does fix that, but will eventually break and need to be reset.

Brew currently side steps that. I think the complexity is unavoidable in
any of these cases. We could certainly remove the beta brew option and use
stable again.

I thinking given the range of use cases and ways in which people use their
PCs, choice is more prudent even if it's more work to maintain.

Layering and resetting isn't the biggest deal ever, but if the config also
resets it's going be a major pain point for average users, and the portal
and just scripts will need to a quick fix for folks that don't like the
terminal as were trying to move away from terminal first solutions.

I get the idea, and as I said I'll defer, but don't fully agree with a
layer first approach.

I'd also like to add, that regardless, your efforts to help here are appreciated, period.  And I'll just be happy to have a permanent solution.

ykshek:
> I would offer it as a layer
Assuming it's realistic
Don't add to image again since it can break image building

@PVermeer This was what Kyle said on Discord

I believe the plan after some discussion on #bazzite-dev is have desktop=flatpak, deck/desktop(hdr)=layer. I think you can add the specfile under `/spec_files/` to bring it under the [ublue-os/bazzite](https://copr.fedorainfracloud.org/coprs/ublue-os/bazzite/) repo and I'll make changes to the ujust install script and Bazzite docs.

andy10115:
@PVermeer  I'd also change your branch to testing instead of unstable as that's where the action is these days and where the ujust and portal changes will land when done.

PVermeer:
Alright. I need to make some changes to the spec file because I'm using my own tooling in the community package.

I will update when it's ready and tested.

spec_files/sunshine/sunshine.spec
```diff
@@ -0,0 +1,154 @@
...
+Release: 1%{?dist}.bazzite
+Summary: Self-hosted game stream host for Moonlight.
+License: GPLv3-only
+URL: https://github.com/LizardByte/Sunshine
+
+Source0: sunshine-service-override.conf
+
+Patch0: disable-std-static-linking.patch
+Patch1: cuda-use-external-deps.patch
+Patch2: disable-codecov.patch
```
PVermeer, line 14: Not necessary to disable codecov, but it's definitly not needed. It also uploads some stats to god knows where during the build.

spec_files/sunshine/sunshine.spec
```diff
@@ -0,0 +1,154 @@
...
+  "-DSUNSHINE_EXECUTABLE_PATH=%{_bindir}/sunshine"
+  "-DSUNSHINE_ENABLE_X11=ON"
+  "-DSUNSHINE_ENABLE_WAYLAND=ON"
+  "-DSUNSHINE_ENABLE_DRM=ON"
+  "-DSUNSHINE_ENABLE_PORTAL=ON"
+  "-DSUNSHINE_ENABLE_VULKAN=ON"
+  "-DSUNSHINE_ENABLE_KWIN=ON"
+  "-DSUNSHINE_PUBLISHER_NAME=copr:ublue-os:bazzite"
+  "-DSUNSHINE_PUBLISHER_WEBSITE=https://copr.fedorainfracloud.org/coprs/ublue-os/bazzite/sunshine"
+  "-DSUNSHINE_PUBLISHER_ISSUE_URL=https://github.com/ublue-os/bazzite/issues"
```
PVermeer, line 95: This will be printed in the sunshine logs.

PVermeer:
I've created the spec file. This is what actually is needed to build and package sunshine at this time. As mentioned above, I've removed a lot of complexity from the official spec file (mainly fixes ontop of fixes that where not needed in the first place).

- Cuda deps are fetched from mamba (anaconda) packages. The cuda patch is to make sure the final linker forgets about the "mamba lib includes" and only uses what it is actually missing. This increases compatibility with Fedora.
- The main service file is `sunshine.service`. The FQDN service `app-dev.lizardbyte.app.Sunshine.service` can also be used.
- Gnome autostart service has been fixed.

I will still maintain the community packages and I've added this spec to my test builds. In this repo I'm keeping track of the changes with beta. Before a major Fedora release and after a stable update I will update this spec and patches with a PR.

ykshek:
@PVermeer You can merge https://github.com/PVermeer/bazzite/pull/2 to include other changes into this PR

andy10115:
@KyleGospo @PVermeer bumping this one again.

KyleGospo:
@PVermeer Would you be open to this moving to the terra repo where we have auto updating based on tags? That'll let more than just us benefit from this and we can retool this PR ever so slightly to use the package from there.

PVermeer:
> @PVermeer Would you be open to this moving to the terra repo where we have auto updating based on tags? That'll let more than just us benefit from this and we can retool this PR ever so slightly to use the package from there.

@KyleGospo I'm not open to maintaining that. Sunshine is PITA project to maintain because of constant breaking changes in the build system and code. I need a fully controlled (by me) environment with a testing playground. With COPR I can easily do that.

1. `auto updating based on tags`: There is no way to determine releases based on git tags / branches or any ref in git. Only GH releases contain information about a release. I had to create a custom updater to check for new release via the GH api.
2. The released source on GH is incomplete / missing deps. Besides that, the build (cmake) is using external commands, including git commands and needs internet. So offline builds and building outside of the git repo (or with depth=1) is not possible. 
4. I'm building beta once a week, not because I want to provide this package (but people can still use it), but mainly to keep up with the changes in smaller chunks.

These are some of issues I encountered and from time to time something else pops up. I think auto-updating sunshine is just a bad idea. It needs to be (manually) tested before release.

PR #4516, opened 2026-03-31
fix: Repo owners with upper-case names can build images on their forks

FairlySadPanda:
Prevents a gotcha with people with capital letters in their usernames not actually being able to fork and build the project.

It's possible to (ab)use Bash to do this, but marking strings with little crab stalks (`;;`) to do it is weird Bash magic. This project already to-lowers the usernames this way in two places, so it makes sense to just do that. Happy to do the crab stalks trick instead though where suitable.


KyleGospo:
Thanks!

PR #4526, opened 2026-04-01
chore: bump Fedora to 44

dylanmtaylor:
Bump Fedora version to 44 in Containerfile and build workflow.


dylanmtaylor:
Blocked on bazzite-multilib COPR being rebuilt for F44. Packages like pipewire, bluez, NetworkManager, and xorg-x11-server-Xwayland in the COPR are older than F44 base and cause conflicts during the swap step.

PR #4534, opened 2026-04-02
fix(ujust): have brew bundle output install progress when enabling bazzite-cli

shahoob:
kinda self explanatory but this fixes a non-issue where i can't see what homebrew is doing while it installs the tools for the experience
also sorry i forgot to use conventional commits =(


shahoob:
thank you!

PR #4542, opened 2026-04-03
chore: inline build scripts into Justfile

dylanmtaylor:
Migrates bazzite's local build system towards the pattern used by aurora/bluefin.

Changes:
- Inlines 8 shell scripts into Justfile recipes
- Adds `_resolve_image` and `_container_mgr` as reusable just helpers
- Resolves kernel version dynamically via `skopeo inspect` instead of relying on Containerfile defaults
- Passes all build args explicitly to `podman build`
- Updates ISO scripts to use just helpers instead of sourcing shell scripts


dylanmtaylor:
just --list --unsorted:

```
Available recipes:
    [Just]
    just-check                           # Check Just Syntax

    [Image]
    build target="" image=""             # Build image
    run-container target="" image=""     # Run Container [alias: run]

    [ISO]
    build-iso-release target="" image="" # Build ISO [alias: build-iso]
    build-iso-git target="" image=""     # Build ISO using ISO Builder Git Head
    run-iso target="" image=""           # Run ISO

    [Utility]
    list-images                          # List Images
    clean-images                         # Clean Images
    clean-isos                           # Clean ISOs
```

m2Giles:
This is awesome. I would recommend using patterns from Aurora and the archived Cayo repos. Akmods also has some good insights as well.

dylanmtaylor:
> This is awesome. I would recommend using patterns from Aurora and the archived Cayo repos. Akmods also has some good insights as well.

I think this is awfully close to what they do, any specifics?

renner0e:
If there is still motivation to do this, I recommend adopting this new pattern:
- https://github.com/ublue-os/aurora/issues/2436

PR #4556, opened 2026-04-05
chore(liveiso): add simple motd and remove verified image check

Lumaeris:
This is enough so it doesn't throw rpm-ostree error upon startup twice

Screenshot From 2026-04-05 14-01-18

UPD: The second rpm-ostree error actually came from verify_motd.sh which doesn't really need to be there for Live ISO


ZachrielForever:
Looks good, but whats with that weird L in the word 'please'?

Lumaeris:
> Looks good, but whats with that weird L in the word 'please'?

That's just Maple Mono on my system, ignore it :P

PR #4566, opened 2026-04-05
feat(gnome): Add template for empty file.

BlueGen3sis:
It would be a nice QoL addition for new GNOME users.

### DIFF
--- a/system_files/desktop/shared/usr/libexec/bazzite-user-setup
+++ b/system_files/desktop/shared/usr/libexec/bazzite-user-setup
@@ -167,8 +167,8 @@ else
   if [[ ! -f "$TEMPLATES_DIR/MangoHud.conf" && -f "/usr/share/doc/mangohud/MangoHud.conf.example" ]]; then
     ln -s "/usr/share/doc/mangohud/MangoHud.conf.example" "$TEMPLATES_DIR/MangoHud.conf"
   fi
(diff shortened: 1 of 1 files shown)


HikariKnight:
the name `document` might make the template be mistaken for an actual rich document file like odt, docx, etc
maybe change it to text file or just `Empty File`?

also `$TEMPLATES_DIR` is undefined and will be empty

BlueGen3sis:
Good idea, but how do you mean that `$TEMPLATES_DIR` is undefined? 
It seems to be defined with ` TEMPLATES_DIR=$(/usr/bin/xdg-user-dir TEMPLATES || true)`, with  `TEMPLATES_DIR="$HOME/Templates"` as a fallback

HikariKnight:
> Good idea, but how do you mean that `$TEMPLATES_DIR` is undefined? It seems to be defined with ` TEMPLATES_DIR=$(/usr/bin/xdg-user-dir TEMPLATES || true)`, with `TEMPLATES_DIR="$HOME/Templates"` as a fallback

was looking at it at the middle of the night and didnt see that github didnt load all the lines above the commit when i told it to load them when i checked 😂 my bad

HikariKnight (dismissed):

HikariKnight (approved):

PR #4570, opened 2026-04-06
fix(ci): use github.repository for kickoff dx

Lumaeris:
Apparently github.action_repository simply doesn't exist, github.repository does though and it should be enough


HikariKnight (approved):

renner0e (approved):

FairlySadPanda:
My bad:
image

`github.action_repository` seems to reference the action's repo when running an action. I must have grabbed the wrong var when searching for the one I needed. (`action_repository` could be interpreted as "the repository where the action is running").

PR #4574, opened 2026-04-06
feat(sunshine-brew): add announcement for active sunshine users

ykshek:
Adds a conditional announcement if sunshine is being actively used, to notify users of the changes to sunshine, and point them to a page for instructions to do so. Requires https://github.com/ublue-os/docs.bazzite.gg/pull/426

Not sure if the switching instructions would be better off as an announcement on Ublue/Bazzite Discourse rather than in bazzite docs though. This can be updated if it does need to move elsewhere.


ykshek:
@PVermeer might want to take a look at this

KyleGospo (approved):

ykshek:
This PR should hit stable at least one image before sunshine removal from base image occurs, to avoid existing sunshine users ending up with no stream

PR #4598, opened 2026-04-09
fix(sunshine): fix brew sunshine losing KMS capture

ykshek:
Sort of a refactor of brew update sunshine

> I guess a temp fix would be to pin the install
> otherwise we would need setcap each time sunshine is updated
> actually doesn't seem like that bad of an idea
> we could add the postinst script ~~as a git submodule~~ ~~subtree?~~ ghcurl it and put it somewhere like /usr/libexec 
> then add a update|upgrade option in the sunshine ujust
> and pin it by default
> hell yeah


ykshek:
one more commit to fix xdg portal capture because brew sunshine doesn't carry the service file, and their build system is way too complicated for me to add it for them upstream to add a 5 second delay. Therefore just override it locally instead, with the same contents on their repo

The service file exists as a .service.in on their repo, which seems to be auto-generated by gh actions for the actual file so unfortunately had to hardcode this instead of ghcurling like the postinst script.

PR #4599, opened 2026-04-09
fix: Use FEDORA_VERSION and ARCH in KERNEL_VERSION

kressb-gitlab:
fix: Don't hardcode "43" in the Container file, use the FEDORA_VERSION arg.

Also use ARCH while we're at it.


KyleGospo:
@kressb-gitlab will merge, but has conflicts atm

PR #4608, opened 2026-04-10
feat(portal): add portal to prepinned apps

Xarishark:
This PR adds the Bazzite Portal kickstart too to the prepinned apps so users can have it ready to go out of the box instead of searching for it.


HikariKnight (approved):

Zeglius (approved):
i

Issue #4262, opened 2026-03-03
In desktop mode on screen keyboard no longer works in terminal


coryrauch:
### Describe the bug

I am on the latest release. When I am in desktop mode the steam on screen keyboard no longer works in the terminal. Everytime I type the terminal window loses focus.

### What did you expect to happen?

To be able to type in terminal window

### Session Logs

```shell
N/A
```

### Hardware

Lenovo Legion Go

coryrauch:
Sorry unable to run terminal command without an screen keyboard working at the moment. Once I am in front of my home computer I will update ticket. The issue appears to be a window docus issue to me though I was reading that the terminal app was changed so maybe related.

nicknamenamenick:
Didn't test this myself yet because also not home, but like you said it's probably due to the KDE Bazzite images switching back to Konsole.

ImranR98:
At least for me, this is not limited to the terminal. It happens in all text inputs I've tried.

xkotlet12:
Also experiencing the same issue. Cannot use the virtual keyboard in any app as the window will lose focus the second a key is pressed.

ConnorS1110:
I also am having this happen for all input fields. Have to use the thumbstick to manually select what I want to type, but focus is lost the instant I tap a character on screen

lonsdaleite:
Same here on Asus Xbox Ally X. The bug exists in [43.20260303](https://github.com/ublue-os/bazzite/releases/tag/43.20260303) and [43.20260309.1](https://github.com/ublue-os/bazzite/releases/tag/43.20260309.1), had to rebase back to [43.20260217](https://github.com/ublue-os/bazzite/releases/tag/43.20260217)

I've also tested [testing-43.20260221](https://github.com/ublue-os/bazzite/releases/tag/testing-43.20260221) (the closest one to 43.20260217), it also has the bug.

lonsdaleite:
The bug still exists in 43.20260313

DIeHarke:
I found a reddit thread discussing the issue: https://www.reddit.com/r/Bazzite/comments/1rne3ao/after_update_on_screen_keyboard_not_working/

lonsdaleite:
Did anybody test it on some of the latest testing firmwares? Was it fixed?

coryrauch:
I will try later today to see and report back.

RoboChoco776:
Just update to the latest release, still didn't fix the problem

DIeHarke:
[...]

PR #4614, opened 2026-04-11
fix: patch Steamdeck-KDE-Presets for functional Steam keyboard

stellaberrant:
fixes https://github.com/ublue-os/bazzite/issues/4562 https://github.com/ublue-os/bazzite/issues/4262 
after this is merged, steamdeck-kde-presets needs to be rebuilt


spec_files/steamdeck-kde-presets/steamdeck-kde-presets.spec
renner0e:
```suggestion
Release:        2%{?dist}
```

 needs to be bumped in order dnf to pull it

HikariKnight (approved):

PR #4671, opened 2026-04-18
fix: reenable terra

playviewer:


KyleGospo:
This is intentionally disabled

PR #4677, opened 2026-04-19
feat: add steam first run window

Lumaeris:
It simply adds a new window that tells unaware users that Steam is about to download itself for the first time. It's the least we can do right now to prevent confused people to assume that Steam is broken. Suggested by Xarishark

(I've changed "Please be patient" to "Please wait" since making those screenshots)

Screenshot From 2026-04-19 14-16-23

Screenshot From 2026-04-19 14-17-04


wolfyreload:
This is a fantastic idea! I've seen so many posts on Reddit complaining that "Steam isn't working" on Bazzite. This will make it much clearer what is happening in the background

HikariKnight:
wont this run each time you manually run steam due to `/usr/share/applications/steam.desktop` launching `bazzite-steam`

i see nothing that edits that to be compatible with this

Lumaeris:
> wont this run each time you manually run steam due to `/usr/share/applications/steam.desktop` launching `bazzite-steam`
> 
> i see nothing that edits that to be compatible with this

I don't see an issue with it. The script checks for a file in Steam installation, once found it silently exits. If needed I can copy the check to bazzite-steam also so it can happen instantly

PR #4706, opened 2026-04-24
feat(KDE): improve Traditional Chinese and Indic languages support ootb for KDE image

tingweiwan:
This PR adds the most commonly used Traditional Chinese input method module (fcitx5-chewing), and the tesseract language packs for Traditional Chinese to the KDE Plasma image. This should improve the ootb experience of KDE images for Traditional Chinese users. 
Also, the M17N input method module for fcitx5 is added to bring feature parity with GNOME based images, as M17N for ibus is shipped by default with fedora.
Finally, the English language pack of tesseract is included as well.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -435,6 +435,7 @@ RUN --mount=type=cache,dst=/var/cache \
             kio-extras \
             krunner-bazaar \
             krdc \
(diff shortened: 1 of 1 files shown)


KyleGospo:
Thanks!

PR #4717, opened 2026-04-26
chore: update xwayland allow-commits revert patch for F44

dylanmtaylor:
Regenerate the SteamOS allow-commits revert patch for xwayland 24.1.10.

The upstream code is now wrapped in `#ifdef XACE` guards, causing the old patch to fail.

Tested: full rpmbuild in F44 container passes.


dylanmtaylor:
This patch is regenerated with git format-patch against the latest source code. The issue was that the #ifdef XACE guards were added around the access restriction code in 24.1.10, causing the old patch context to not match.

PR #4729, opened 2026-04-27
chore(gnome): explicitly set nodejs22

Lumaeris:
unstable now has a leftover of NodeJS because of package name change happened in F44. I removed locale gen support in #4728 but figured I shouldn't do two changes at once


dylanmtaylor:
Why not 24 which is the new LTS? Incompatibility?

Lumaeris:
It was automatically selected by dnf when trying to install `nodejs-npm`. If npm will throw an error while trying to build at some point in the future then sure I'll change it. I doubt it's highly important for RebootToUEFI

PR #4758, opened 2026-04-29
fix(build): remove libaacs and libbdplus (post-RPM Fusion cleanup)

dylanmtaylor:
The build was failing because libaacs and libbdplus are no longer available in the repositories after dropping RPM Fusion. This PR removes them from the Containerfile to restore CI stability.

Log error:
```
No match for argument: libaacs
No match for argument: libbdplus
```

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -174,8 +174,6 @@ RUN --mount=type=cache,dst=/var/cache \
     dnf5 -y install \
         libfreeaptx && \
     dnf5 -y install --enable-repo="*fedora-multimedia*" \
-        libaacs \
-        libbdplus \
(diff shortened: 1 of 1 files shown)


KyleGospo:
These should be present and provided by negativo, better to fix by identifying why they're missing

dylanmtaylor:
> These should be present and provided by negativo, better to fix by identifying why they're missing

These don't seem to be packaged by negativo under this name. The closest I can identify is `makemkv`

dylanmtaylor:
Take a look at https://negativo17.org/repos/multimedia/fedora-44/x86_64/

PR #4759, opened 2026-04-29
feat: add MakeMKV and symlink libmmbd for Blu-ray playback

dylanmtaylor:
"MakeMKV comes with a library called libmmbd. This library exposes an AACS- and BD+-compatible API but routes all decryption to MakeMKV engine in the background. If you replace the system's libaacs and libbdplus with libmmbd, then any program that uses the open-source libbluray library for Blu-ray playback (like VLC or mplayer) will magically start to play all encrypted discs."
https://forum.makemkv.com/forum/viewtopic.php?f=3&t=7009

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -175,7 +175,12 @@ RUN --mount=type=cache,dst=/var/cache \
         libfreeaptx && \
     dnf5 -y install --enable-repo="*fedora-multimedia*" \
         libbluray \
-        libbluray-utils && \
+        libbluray-utils \
+        makemkv && \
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
This is consistent with how Arch Linux does it: http://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=makemkv-libaacs

```
package() {
  install -d $pkgdir/usr/lib
  ln -s libmmbd.so.0 $pkgdir/usr/lib/libaacs.so.0
  ln -s libmmbd.so.0 $pkgdir/usr/lib/libbdplus.so.0
}
```

PR #4785, opened 2026-04-30
revertme: disable dx kickoff for the time being

Lumaeris:
There's no reason to do DX builds since -deck is still on F43


dylanmtaylor:
I was trying to do some refactoring to merge dx into the main repo; I think we may want to drop this entirely and pull in dx to a single streamlined CI process.

dylanmtaylor:
Can we rewrite the commit message to just "ci: disable dx kickoff"? That's semantic and open for interpretation as far as what we do with it later on.

PR #4788, opened 2026-04-30
feat: hide makemkv desktop entry

renner0e:
The intention with this was to abuse the shipped libmmbd in makemkv to make open source programs use the makemkv provided decryption engine in 3f4b939826 which should make it possible to play commercial/decrypted blu-rays with them.

I think it was not intended to actually expose the full application here. Hiding the desktop entry seems to be the most maintainable option in my opinion as it would still allow users to unhide the desktop entry, if they know it exists in the first place, without undoing significant amounts of work of the packaging of makemkv here in Bazzite just to ship those 3 symlinks.

Although I'm a little bit confused what specific usecase this commit mentioned above actually fixes as applications like VLC would be installed as a flatpak which will not use system provided libraries.

@dylanmtaylor thoughts?

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -177,6 +177,7 @@ RUN --mount=type=cache,dst=/var/cache \
         libbluray \
         libbluray-utils \
         makemkv && \
+    desktop-file-edit --set-key=Hidden --set-value=true /usr/share/applications/makemkv.desktop && \
     ln -sf /usr/lib64/libmmbd.so.0 /usr/lib64/libaacs.so.0 && \
     ln -sf /usr/lib64/libmmbd.so.0 /usr/lib64/libaacs.so.0.7.2 && \
     ln -sf /usr/lib64/libmmbd.so.0 /usr/lib64/libbdplus.so.0 && \


dylanmtaylor:
As someone who doesn't play Blu-ray discs on my computer, I'm not too invested in this one one way or another. We ran into a tough spot when we removed RPMfusion. From my understanding of the situation, MakeMKV is one of the only ways to decrypt and play this media on Linux. It's a tricky situation because it's not completely free and open source software, but on the other hand, users who play blu-rays and don't want to lose the ability to do so will appreciate it being there. Users may need to go into the GUI and reconfigure things on occasion from what I was seeing in Discord. There is some sort of key system for the program itself but it is currently free at the moment. I don't love that. We don't have a good alternative though. With copy protection schemes, especially on Linux, I don't anticipate there being something that works which is free and open source and can decrypt Blu-ray BD+ media, especially as new discs come out with new keys, etc. Blu-ray playback and ripping is kind of a mess. This was shipped entirely as a replacement for the previous libraries. I think right now it's probably between shipping this software or removing Blu-ray support entirely. Removing the ability to play and decode Blu-ray would be a regression from 43.

PR #4791, opened 2026-04-30
feat: add rocm-smi back to the image

AlexNPavel:
The `rocm-smi` package was removed from the base images along with the whole ROCm suite in the Fedora 44 update. This broke reporting of GPU usage stats in `btop` for AMD GPUs. The `rocm-smi` package is tiny and doesn't depend on the rest of the ROCm suite. I did a local build with this change and found that adding the `rocm-smi` package back only added 2.9MB to the image size.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -248,6 +248,7 @@ RUN --mount=type=cache,dst=/var/cache \
         webapp-manager \
         btop \
         amdsmi \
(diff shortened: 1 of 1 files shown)


KyleGospo:
This isn't returning, use nvtop or wait for btop to support the modern replacement from AMD.

Issue #4792, opened 2026-04-30
Revert back to sunshine beta rpm with freeze


Zeglius:
It seems that homebrew one is too finicky. A rpm version freeze mechanism would be a better middle ground between stable and beta rpm, rather than having to deal with beta sudden changes.

Zeglius:
We can also discuss whenever simply returning back to beta rpm without freeze.

Zeglius:
Cc @Xarishark @KyleGospo

PR #4794, opened 2026-04-30
feat: switch sunshine to frozen beta RPM and restore F43 ujust behavior

dylanmtaylor:
https://github.com/ublue-os/bazzite/issues/4792


dylanmtaylor:
Closing based on chat w/ Kyle:
"I don't wanna use a beta sunshine rpm again
This is their problem now, we can't keep covering for their bad packaging
Nor should we be responsible to figure out which cursed beta version works well enough to ship
Direct any support requests to sunshine or tell users to use an alternative"

PR #4795, opened 2026-04-30
fix: plasma-setup branding

renner0e:
This gives plasma setup the bazzite wallpapers. Needs
https://github.com/ublue-os/packages/pull/1231

We need to install it via this stupid way to avoid a possible mismatch
between plasma packages, as basically nobody tests mixing them. This is
especially important on major releases. we are maintaining our own
version and need to keep it updated, a failing build will force us to do
that.

The reason why we have patches in the first place is because of
https://invent.kde.org/plasma/plasma-setup/-/work_items/72

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -406,6 +406,8 @@ RUN --mount=type=cache,dst=/var/cache \
             kcharselect \
             kde-partitionmanager \
             plasma-discover && \
+        dnf5 -y swap --repo=copr:copr.fedorainfracloud.org:ublue-os:staging \
+            plasma-setup plasma-setup-"$(rpm -q --qf "%{VERSION}" plasma-desktop)"-*.bazzite && \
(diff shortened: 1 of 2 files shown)


renner0e:
One little issue with the current ISOs is that plasma-setup gets triggered on first boot. It benefits nobody.

We are still using anaconda for the user creation, we could be using plasma-setup for the user creation now and have one less page in anaconda, which would be the same way how gnome works.

renner0e:
I have not tested this change beyond trying things out via bootc install

renner0e:
This needs some more changes, deck images can't use plasma-setup

Lumaeris:
As of now on GNOME side Initial Setup package gets deleted so we can't just hide accounts screen.

I did a simple PR for adding Bazzite icon to Initial Setup, additionally it swaps to it. https://github.com/ublue-os/bazzite/pull/4441 Maybe it needs to be moved to ublue/packages repo instead?

PR #4819, opened 2026-05-02
chore: Re-Add HID-TMFF2

trbocode:
It was deleted due to another, unnecessary driver it pulled (Terra issue), but the driver itself is needed for wheel support


KyleGospo:
Thanks!

PR #4835, opened 2026-05-04
fix: setup-virtualization overlay libvirt packages

roworu:
That will add overlaying of libvirt packages in setup-virtualization ujust script, as they were removed from base image in 44 release.

I was also thinking, if it would be better to just show a deprecation notice for such use case and offer to rebase to DX variant. Let me know, if such variant would be better, I will change the PR.

Thanks!


roworu:
Or maybe it is worth completely removing the setup-virtualization from Bazzite image?...

roworu:
Will be reimplemented later with Flatpak QEMU version.

PR #4846, opened 2026-05-05
feat(portal): add resolve to the portal

Xarishark:


Lumaeris:
Pointing out that it needs to be merged after https://github.com/ublue-os/packages/pull/1236

Xarishark:
@Lumaeris Actions already existed but they were not working properly. Also this will stay on the testing channel so there is no need to hold it back as testers are supposed to know whats going also its mostly us. Only thing that will time out is the status query but that does not matter as 3 seconds will pass and the user will still get the window with the enable disable choices.

PR #4902, opened 2026-05-12
chore(ci): replace retry action

Lumaeris:
Retry action doesn't work as intended[1] and skopeo can't extract[2] digest[3] key[4] so replacing the whole step entirely and bringing the code @tulilirockz was using for Zirconium for several months now, slightly adjusted for our needs.

Also moving Cosign install step before the actual image push so users wouldn't end up with unsigned image

[1]: https://github.com/ublue-os/bazzite/actions/runs/25670171643/job/75353034807
[2]: https://github.com/ublue-os/bazzite/actions/runs/25681930815/job/75395592341
[3]: https://github.com/ublue-os/bazzite/actions/runs/25681937840/job/75395492760
[4]: https://github.com/ublue-os/bazzite/actions/runs/25681937840/job/75395492782

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -493,7 +493,7 @@ jobs:
           ALIAS_TAGS: ${{ steps.generate-tags.outputs.alias_tags }}
           OUTPUT_IMAGE: ${{ steps.base.outputs.output_image }}
           VERSION: ${{ steps.relabel.outputs.version }}
-          MAX_RETRIES: 3
+          MAX_RETRIES: 5
           WAIT_SECONDS: 15
         run: |
           set -x
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -479,44 +479,32 @@ jobs:
           echo ${{ secrets.GITHUB_TOKEN }} | sudo podman login ghcr.io -u ${{ github.actor }} --password-stdin
           echo ${{ secrets.GITHUB_TOKEN }} | docker login ghcr.io -u ${{ github.actor }} --password-stdin
 
-      # Push the image to GHCR (Image Registry)
-      - name: Push to GHCR
-        id: push
-        if: github.event_name != 'pull_request'
-        uses: nick-fields/retry@ad984534de44a9489a53aefd81eb77f87c70dc60 # v4.0.0
-        env:
-          ALIAS_TAGS: ${{ steps.generate-tags.outputs.alias_tags }}
-          OUTPUT_IMAGE: ${{ steps.base.outputs.output_image }}
-          RECHUNK_REF: ${{ steps.rechunk.outputs.ref }}
-          VERSION: ${{ steps.relabel.outputs.version }}
-        with:
-          max_attempts: 3
-          retry_wait_seconds: 15
-          timeout_minutes: 10
-          command: |
-            log_sum() { echo "$1" >> $GITHUB_STEP_SUMMARY; }
-            log_sum '# Push to GHCR result'
-            log_sum '```'
-
-            dest_image="${OUTPUT_IMAGE}:${VERSION}"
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -479,44 +479,32 @@ jobs:
...
+          MAX_RETRIES: 3
+          WAIT_SECONDS: 15
+        run: |
+          set -x
 
-            digest=$(sudo skopeo inspect --format '{{.Digest}}' docker://${OUTPUT_IMAGE}:${VERSION})
-            echo "digest=${digest}" >> $GITHUB_OUTPUT
+          for tag in ${VERSION} ${ALIAS_TAGS}; do
+            for i in $(seq "${MAX_RETRIES}"); do
+              sudo podman push --digestfile=/tmp/digestfile "localhost/chunked-img" "${OUTPUT_IMAGE}:$tag" && break || sleep $((${WAIT_SECONDS} * i));
```
renner0e: Why the change of mind to use a digestfile here, zirconium does this because they are doing multiarch.

renner0e (commented):
I like this, we probably shouldn't rely on actions for things that can easily replaced with bash.
Did you test this on a fork already?

.github/workflows/build.yml
Lumaeris: podman push only has this flag related to digest, I remember Zirconium used it before building multiarch

Lumaeris: While skopeo generally works, I don't think it's stable enough to get digest of already pushed image, see footnotes

Lumaeris:
> Did you test this on a fork already?

Yeah, that's what I'm using in my new repos now. Not Bazzite fork specifically, though I should probably check just in case

Lumaeris:
So I tried it in my fork and I actually saw a bunch of 502 errors in some logs, so it clearly works! :slightly_smiling_face: 

https://github.com/Lumaeris/bazzite/actions/runs/25719213176

But maybe just to be safe I'll bump the max retries number to 5

UPD: And here's with summary, I added log_sum function back https://github.com/Lumaeris/bazzite/actions/runs/25723316438

.github/workflows/build.yml
renner0e: yeah that makes sense to me, awesome if this removes one more step where things could go wrong

renner0e:
Could we give the retry flags of push another try?
There was some reason why we have this action in the first place

renner0e:
> Could we give the retry flags of push another try? There was some reason why we have this action in the first place

Oh yeah, I know why we didn't use --retry before, podman is old af in ubuntu 24.04, the flag doesn't exist. newer versions have it. This will have to wait until we jump to new runners or install a newer version.

KyleGospo:
> > Could we give the retry flags of push another try? There was some reason why we have this action in the first place
> 
> Oh yeah, I know why we didn't use --retry before, podman is old af in ubuntu 24.04, the flag doesn't exist. newer versions have it. This will have to wait until we jump to new runners or install a newer version.

Do you know when that jump might occur?

renner0e:
> > > Could we give the retry flags of push another try? There was some reason why we have this action in the first place
> > 
> > 
> > Oh yeah, I know why we didn't use --retry before, podman is old af in ubuntu 24.04, the flag doesn't exist. newer versions have it. This will have to wait until we jump to new runners or install a newer version.
> 
> Do you know when that jump might occur?

No idea, I stole this from bootc actions repo and changed it to Ubuntu 26.04 (bootc did 25.04?) and that also works
https://github.com/ublue-os/aurora/blob/main/.github/workflows/reusable-build.yml#L76

Podman is new enough there.

The reason why we want new podman besides that is
https://github.com/ublue-os/aurora/blob/main/Justfile#L784

PR #4924, opened 2026-05-15
feat(smb): install mount manager

Xarishark:
This PR adds smb mount manager to Bazzite so our users will have a better more stable experience adding their shares to the system as both KIO and GIO provide an unstable experience when it comes to SMB shares. 

Mount manager is under my git now and not under ublues org. So it can be treated as a preinstalled app exactly like protonplus for example.

EDIT: PR changed to draft and will be converted to brew if accepted.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -289,6 +289,7 @@ RUN --mount=type=cache,dst=/var/cache \
         wlr-randr \
         gmodpatchtool \
         bazzite-portal \
+        smb-mount-manager \
         ls-iommu && \
     ln -s /dev/null /etc/NetworkManager/dispatcher.d/04-iscsi && \
     systemctl mask iscsi && \


tulilirockz (changes requested):
NACK, should be optionally done or not at all

Xarishark:
> NACK, should be optionally done or not at all

Would ublues tap be a preferable path for this? As its a tool specifically for bazzite in its current form.

renner0e:
This would be perfect as a flatpak on Flathub to benefit everyone.

If the app matures more and proves to be useful then it still can be shipped on bazzite.

Xarishark:
Flathub is a different beast tbh but I understand what you mean tho I would prefer this thing to stay as a tool until plasma/gnome get a proper CIFS implementation in order as I feel that smb shares should be managed fully by the file manager (but still mount on a low level instead of being a framework thing specific to the DE). 

So in this context making it a flatpak release feels wrong to me as I dont want to do the work to manage it for everyone (that uses flatpak) but I am totally ok doing the work to manage it for Bazzite(and the rest of ublue ofc)

So would a cask on our tap be a better solution? That makes it fully optional to the user and also its not part of the image (other than being in the portal as a choice)

Xarishark:
Just in case I already started work for the tap path as I dont know how comparable to publishing on Terra it is, tho it looks close in logic from the other casks I am checking.

I converted this PR to draft and if I get a verification that the brew path is ok I will just change it to add the button in the Portal if that is acceptable.

PR #5036, opened 2026-06-05
fix(deck): fix: set capture=kms and system_tray=disabled for sunshine-beta install

andy10115:
When installing Sunshine Beta via the Brew package, the service would fail to start because sunshine.conf was missing required configuration for Deck/KMS environments. This fix writes capture = kms and system_tray = disabled to ~/.config/sunshine/sunshine.conf before the service starts, ensuring the beta install works out of the box for Deck users without requiring manual config edits.

### DIFF
--- a/system_files/desktop/shared/usr/share/ublue-os/just/82-bazzite-sunshine.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/82-bazzite-sunshine.just
@@ -105,6 +105,18 @@ setup-sunshine ACTION="":
           exit 1
         fi
         sudo /usr/libexec/sunshine-postinst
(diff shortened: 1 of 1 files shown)


andy10115:
Merged to incorrect branch.

KyleGospo:
This is fine on main if you reopen, I can get it where it needs to be

andy10115:
reopened, thanks. should probably go into testing as well with the deck users coming in. Let me know if i need to submit specifically to that, sorry 100% new at this.

HikariKnight (approved):
LGTM

ykshek:
+1

KyleGospo:
+1

PR #5050, opened 2026-06-07
feat(sunshine): add notice for deck images

ykshek:
- Add and updated notices for deck and desktop users.
- Also adds a conditional to make #5036 activate on deck only.
- Add yellow coloring to any "Sunshine" that refers to the installed package of Sunshine.
- Add bold to links and keywords.
- Add an option to keep data on delete for flatpak installations.


andy10115:
+1

PR #5075, opened 2026-06-10
fix(just): add sudo to lsblk in configure-beesd for non-root detection

xlionjuan:
The `configure-beesd` recipe uses `lsblk --f` to detect BTRFS filesystems, but lsblk requires root to read filesystem info on Fedora 44. When run without sudo, it returns empty and reports `No BTRFS filesystems found.`, even though the system has BTRFS volumes.

Prepend sudo to the lsblk command so the recipe works regardless of whether it's invoked directly or through the Bazzite Portal (yafti) which runs ujust as the regular user.


KyleGospo:
Thanks!

PR #5076, opened 2026-06-11
chore(Portal): Give welcome tab a second pass

EPOCHvoyager:
Another pass on text introduced in #5052. The goal was to improve readability, consistency with the current state of Portal, and read a little warmer.

I've run this by @andy10115, and got a “Ship it bro!” — which I'll consider an ACK.


andy10115:
+1

PR #5103, opened 2026-06-15
fix: restore privileged user setup service enablement

S1mplector:
## Summary

- Finish the `steam-web-debug-portforward@...` enable command in `bazzite-privileged-user-setup` so the script parses and runs again.
- Use the unescaped systemd instance specifier for the port-forward service user, avoiding the invalid `/user` expansion from `%f`.
- Quote the username before passing it to `systemd-escape`.

## Why

`bazzite-privileged-user-setup` currently ends with a trailing `&& \`, so `bash -n` reports an unexpected EOF and the privileged first-login setup path cannot run. After completing that command, the enabled template also needs `User=%I`; `User=%f` expands to a path-like `/instance` value and systemd rejects the unit.

## Testing

- `bash -n system_files/desktop/shared/usr/libexec/bazzite-privileged-user-setup`
- shell syntax sweep over executable shell entrypoints
- `systemd-analyze verify system_files/deck/shared/usr/lib/systemd/system/steam-web-debug-portforward@.service`
- `just just-check`
- `git diff --check`

### DIFF
--- a/system_files/deck/shared/usr/lib/systemd/system/steam-web-debug-portforward@.service
+++ b/system_files/deck/shared/usr/lib/systemd/system/steam-web-debug-portforward@.service
@@ -1,9 +1,9 @@
 [Unit]
-Description="Portforward CEF remote debug port for user %f"
+Description="Portforward CEF remote debug port for user %I"
 
 [Service]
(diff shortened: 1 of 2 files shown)


KyleGospo:
Thank you!

PR #5152, opened 2026-06-26
fix: try to work around an issue with unusable controllers in Plasma 6.7

stellaberrant:
Plasma 6.7 enabled 'allow using as pointer and keyboard' again by default which makes nearly every game unplayable. This is an attempt to fix this issue.

### DIFF
--- a/system_files/desktop/kinoite/usr/share/kde-settings/kde-profile/default/xdg/kwinrc
+++ b/system_files/desktop/kinoite/usr/share/kde-settings/kde-profile/default/xdg/kwinrc
@@ -1,0 +1,6 @@
+[Wayland]
+InputMethod[$e]=/usr/share/applications/com.github.maliit.keyboard.desktop
+VirtualKeyboardEnabled=true
+
+[Plugins]
+gamecontrollerEnabled=false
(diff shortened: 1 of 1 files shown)


renner0e:
Is this related to this?
https://invent.kde.org/plasma/kwin/-/commit/ffdf99a2ecbf0990630f25f930c8148ab241bd44

system_files/desktop/kinoite/usr/share/kde-settings/kde-profile/default/xdg/kwinrc
```diff
@@ -0,0 +1,6 @@
+[Wayland]
+InputMethod[$e]=/usr/share/applications/com.github.maliit.keyboard.desktop
```
renner0e: Does bazzite even use maliit?

Here is the stock fedora config:
https://forge.fedoraproject.org/kde/kde-settings/src/branch/rawhide/usr/share/kde-settings/kde-profile/default/xdg/kwinrc

stellaberrant: my bad

stellaberrant:
> Is this related to this? https://invent.kde.org/plasma/kwin/-/commit/ffdf99a2ecbf0990630f25f930c8148ab241bd44

unsure

stellaberrant:
should i close the PR since it seems like KDE 6.7.1 disables it by default again?

renner0e:
> should i close the PR since it seems like KDE 6.7.1 disables it by default again?

Let's assume the regression that was talked about in the commit I mentioned earlier gets fixed in plasma and will be in 6.8, maybe this is still something that bazzite explicitly wants to set anyway?

KyleGospo:
Iirc kwinrc is in some of our kde config packages, this might need to be changed there

KyleGospo:
Is this still needed?

PR #5176, opened 2026-07-01
feat(just): FW16 Matrix LED setup in 80-bazzite.just

dnkmmr69420:
Adding FW16 Matrix setup
For some context
Led Matrix part: https://frame.work/products/16-led-matrix?v=FRAKDE0001
Input module: https://github.com/FrameworkComputer/inputmodule-rs/blob/main/ledmatrix/README.md
Matrix control software: https://github.com/MidnightJava/led-matrix
Another guide: https://knowledgebase.frame.work/led-matrix-controls-on-linux-ByJKHV7axg

### DIFF
--- a/system_files/desktop/shared/usr/libexec/bazzite-hardware-setup
+++ b/system_files/desktop/shared/usr/libexec/bazzite-hardware-setup
@@ -6,7 +6,7 @@ IMAGE_BRANCH=$(jq -r '."image-branch"' < $IMAGE_INFO)
 FEDORA_VERSION=$(jq -r '."fedora-version"' < $IMAGE_INFO)
 
 # SCRIPT VERSION
-HWS_VER=71
+HWS_VER=72
 HWS_VER_FILE="/etc/bazzite/hws_version"
 HWS_VER_RAN=$(cat $HWS_VER_FILE)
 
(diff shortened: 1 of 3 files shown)


dnkmmr69420:
Do whatever changes you want with this

ykshek:
I think it'd be more appropriate to use `/tmp/` rather than `~/.tmp/`

If possible, the services and executables can be added in the Containerfile, then the just script would be just a toggle to enable/disable the service.

Edit: 
I looked at it some more and I think the installation script could be put in Containerfile with `/ctx/ghcurl`, though ideally it should be packaged as an .rpm

Installation of udev rules should probably live under [/usr/libexec/bazzite-hardware-setup](https://github.com/ublue-os/bazzite/blob/main/system_files/desktop/shared/usr/libexec/bazzite-hardware-setup) in a new `# FRAMEWORK MATRIX LED SETUP` section near `# FRAMEWORK AMD FIXES`

Then the just recipe would just be similar to `enable-framework-fan-control:`, simple toggle to enable/disable the service.

dnkmmr69420:
> I think it'd be more appropriate to use `/tmp/` rather than `~/.tmp/`
> 
> If possible, the services and executables can be added in the Containerfile, then the just script would be just a toggle to enable/disable the service.
> 
> Edit: I looked at it some more and I think the installation script could be put in Containerfile with `/ctx/ghcurl`, though ideally it should be packaged as an .rpm
> 
> Installation of udev rules should probably live under [/usr/libexec/bazzite-hardware-setup](https://github.com/ublue-os/bazzite/blob/main/system_files/desktop/shared/usr/libexec/bazzite-hardware-setup) in a new `# FRAMEWORK MATRIX LED SETUP` section near `# FRAMEWORK AMD FIXES`
> 
> Then the just recipe would just be similar to `enable-framework-fan-control:`, simple toggle to enable/disable the service.

are you able to do any of that because that is outside my scope of knowledge

dnkmmr69420:
it would be nice to have it be part of the image where it can be toggled on and off

PR #5213, opened 2026-07-11
chore(Bazaar): Give the pt_BR locale another pass

EPOCHvoyager:
Cursory read these while working on the previous PRs (#5211 and #5212) and figured it was worthy of some more polish. Modify and correct these locale strings. For specifics on the corrections — one title was verbal instead of nouns (as the others are), and a subtitle was missing a required preposition for a verb.

@koitorin, as you did the original translation work here, would love to hear your thoughts.


koitorin:
Looks good to me. Thanks, @EPOCHvoyager!

PR #5221, opened 2026-07-13
feat(ujust): update fsr4 just script and yml

ykshek:
image

image
image
image


andy10115:
+1 I'd guess GE-Proton will have this in next release as well.

These are going to stay in flux for a bit, but I'll keep an eye on them.

PR #5223, opened 2026-07-13
fix: Add ignore configs for external controllers

HikariKnight:
Ideally this will be temporary and we would remove files as inputplumber gains support for the controllers. We cannot blanket ignore all "xinput" controllers as this would then affect the ROG Ally controls and maybe other handhelds.

I will try make an ujust to generate "per-controller" ignore configs which grabs vendorID and productID that users can have on their system and even provide to us to update these lists.

You can find the needed vendorID and productID by:
1. connecting your controller (on the deck image or after having gone into the gamemode session once in the dx image)
2. run `inputplumber devices list` and note down the id for the controller you connected
3. run `inputplumber device ID info` (replace `ID` with the id you noted down)
4. run `udevadm -a SOURCEDEVICE` replace `SOURCEDEVICE` with the source device shown by inputplumber, from this info we can grab the vendorID and productID to add to our ignore lists (and possibly generate) an ignore config for the user.


HikariKnight:
going to mark this as ready so we have a baseline and i can open a new PR to add missing xinput controllers

PR #5227, opened 2026-07-14
feat: add gnome SearchProvider for Bazzite Portal

ykshek:
GNOME side of https://github.com/ublue-os/bazzite/pull/5129

PR #5129, opened 2026-06-21
feat: krunner-bazzite-portal-plugin

ykshek:
Requires 2 other things:
- ✅~~https://github.com/ublue-os/yafti-gtk/pull/19~~ merged with https://github.com/ublue-os/yafti-gtk/pull/21, waiting for a new release
- ✅~~The [krunner-yafti package](https://github.com/ykshek/krunner-yafti) either added as additional [COPR](https://copr.fedorainfracloud.org/coprs/starfish/krunner-yafti/) in Containerfile or moved into [ublue-os/packages](https://github.com/ublue-os/packages), or~~ #5228 , merged, waiting for a build on ublue-os/bazzite

PR #5228, opened 2026-07-14
feat: add specs for search packages for bazzite portal

ykshek:
alternate requirement for https://github.com/ublue-os/bazzite/pull/5129 and https://github.com/ublue-os/bazzite/pull/5227


ykshek:
Just found out about https://copr.fedorainfracloud.org/coprs/ublue-os/bazzite/ and moving the specs here

PR #5235, opened 2026-07-15
feat(announcements): add nv flatpak mismatch notif

ykshek:
image


ykshek:
This *should* work, but needs someone on Nvidia to actually test it as I don't have an Nvidia device anymore.

ykshek:
Can be seen as a continuation of https://github.com/ublue-os/bazzite/pull/3581

ykshek:
Closed in favor of https://github.com/ublue-os/bazzite/pull/5242

PR #5241, opened 2026-07-16
feat(portal): Add Android TV Wakeup and UnShade Shader Cache Cleaner

andy10115:
2 tools on this one
1. One in "install apps", open to suggestion on where it should go
ATV Couch wake use the Android debug bridge to wake TV and change inputs when PC wakes or boots, and changes to desired input. Also turns tv off on sleep and shutdown.  Mimicing CEC behavior.  Guided user setup is included in onboarding.

2. In trbouleshooting section.  Added Unshade Shader Cache cleaner. Will clean shader cache on AMD GPU's install via curl auto updates everytime it's run.


andy10115:
Apologies, I meant to submit a seperate PR to fix the RyzenAdj just recipte, but did not know it would add it to this PR. Let me know if i need to resubmit. I reverted the change to the just file.

andy10115:
Cancelling, these projects have been combined into one GTK app.

PR #5265, opened 2026-07-20
feat: vmon for streaming(KWin only)

ykshek:
The scripts in /usr/libexec are edited from https://www.reddit.com/r/MoonlightStreaming/comments/1tg1dnc/guide_sunshine_on_kde_wayland_virtual_display/

This adds an action in ujust/Bazzite Portal to add an application entry in Sunshine, which adds the do/undo scripts.


ykshek:
~~also needs https://github.com/ublue-os/docs.bazzite.gg/pull/483~~ merged

system_files/desktop/kinoite/usr/libexec/sunshine-start-vmon
jordan1818, line 1: On line 29, asking AI help get the same answer. However, playing with this a bit more only specifying the primary display to be the virtual one works way better than disabling them.

Trust me, it is better to not disable.

jordan1818, line 1: Speaking on virtual monitors, i read a little about it. This only supports SDR. Which is great, just wonder if there will be thought into HDR support with this tool or upgradable to one with HDR support.

ykshek: Currently this relies on KWin ScreenCast and kfrb-virtualmonitor, and KWin ScreenCast currently doesn't support HDR. HDR with KWin ScreenCast would probably require effort from upstream, or if someone finds a workaround for it.

ykshek: I did do some testing on my own machine, and whether or not I disable the monitor appears to blank out the internal display, though this could be a case of GPU driver bug.

Leaving physical displays enabled has the side effect of losing your cursor and apps more easily, so in this case I believe it is better to leave it disabled.

jordan1818: Good to know! Thanks!

jordan1818: That is true!i did notice that. What i wanted to accomplish was to have a seems non-system involvement. But it is hard to do so with desktop mode. 

Good to know others see this as well!

PR #5279, opened 2026-07-22
Modify Surface fixes to use startswith for Vendor ID

### LINKED ISSUE #5278: Microsoft Surface hardware detection
https://github.com/ublue-os/bazzite/blob/5c62bccc36d1d8e3b3004ffa37a83d5931ba9330/system_files/desktop/shared/usr/libexec/bazzite-hardware-setup#L232

The chassis vendor string on my Surface Pro 9 reports "Microsoft Corporation" not Microsoft.
Suggest changing the detection condition to capture "Microsoft" in the VEN_ID perhaps - 

```
...
if [[ "$VEN_ID" == *Microsoft* ]]; then
...
```

astrobytez:
Update Surface fixes to check for Vendor ID starting with 'Microsoft'.

close #5278

### DIFF
--- a/system_files/desktop/shared/usr/libexec/bazzite-hardware-setup
+++ b/system_files/desktop/shared/usr/libexec/bazzite-hardware-setup
@@ -228,8 +228,8 @@ if [[ ":Framework:" =~ ":$VEN_ID:" ]]; then
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


KyleGospo:
Thanks!

PR #5353, opened 2026-08-01
fix(bazzite_portal): default file name for downloaded decky plugins in ujust

gabripo:
- fixes the Bazzite ujust command to install Decky Framegen


KyleGospo:
Thanks!

PR #5429, opened 2026-08-12
feat(cardwire): Add docs and GUI button to shim

ykshek:
Update switcherooctl shim to have 2 buttons, one to open docs page, one to open cardwire-gui

image


ykshek:
image

PR #5431, opened 2026-08-12
fix(cardwire): add back missing cardwire list exec

HikariKnight:
again makes the shim do the bare minimum expected behavior while nagging about migrating to cardwire


HikariKnight:
the list command was removed in #5429 when @ykshek added additional functionality

we can add a set date to remove the shim in the future like lets say fedora 46 or something?

ykshek:
oops!

imo given it is just a simple script I feel like this should be left in as long as people keep seeing(and posting) `switcherooctl` on protonDB

tbh I am thinking of doing something similar with `gamemoderun` too

PR #5449, opened 2026-08-13
fix: mask appstream refresh service causing slow boot

ykshek:
In the systemd unit:

`Description=Workaround for Atomic Destkops to refresh the appstream cache`
`Documentation=https://gitlab.com/fedora/ostree/sig/-/work_items/103`

thus should be safely maskable as Bazzite uses uupd and doesn't use discover.

Funnily enough it doesn't actually run on my machine due to 

```console
bazzite systemd[1]: [🡕] multi-user.target: Found ordering cycle: fedora-atomic-desktop-appstream-cache-refresh.service/start after graphical.target/start after multi-user.target/start - after fedora-atomic-desktop-ap>
bazzite systemd[1]: [🡕] multi-user.target: Job fedora-atomic-desktop-appstream-cache-refresh.service/start deleted to break ordering cycle starting with multi-user.target/start
```
though some users seems to have it, and take one and a half minute extra to boot.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -297,6 +297,7 @@ RUN --mount=type=cache,dst=/var/cache \
     ln -s /dev/null /etc/NetworkManager/dispatcher.d/04-iscsi && \
     systemctl mask iscsi && \
     systemctl mask systemd-remount-fs.service && \
+    systemctl mask fedora-atomic-desktop-appstream-cache-refresh && \
     mkdir -p /usr/lib/extest/ && \
     /ctx/ghcurl "$(/ctx/ghcurl https://api.github.com/repos/ublue-os/extest/releases/latest | jq -r '.assets[] | select(.name| test(".*so$")).browser_download_url')" -Lo /usr/lib/extest/libextest.so && \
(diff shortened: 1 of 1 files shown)


ykshek:
ok user I was helping installed bazzite on their HDD
either way this should be harmless

renner0e (commented):
Odd, I've never seen this. Do you by any chance have seen/filed a bug report upstream somewhere/know how to reproduce it?

Also if this workaround is specific to discover because it uses appstream to check for major system updates then it might be a good idea to only run this unit when discover is actually on the system and contribute this upstream so we don't have to carry this change and everyone benefits.

ykshek:
The upstream issue for adding the original service unit is this https://gitlab.com/fedora/ostree/sig/-/work_items/103. There was an additional one that updates the mandb too, actually. In the end the user fixed the slow boot problem by reinstalling to his/her SSD.(Sidenote, when it was still on HDD, greenboot took half a minute to complete)

I guess an issue can be created upstream to move the unit to the `plasma-discover` package? But since this appears to be a fedora-atomic specific workaround, I am not sure if it would be appropriate to do that(perhaps something like `WantedBy=ostree-finalize-staged`? It's really jank that way though).

PR #5458, opened 2026-08-14
fix(portal): Update VSCodium Brew Script

andy10115:
Script now trusts and taps ublue-os/tap and then install vscodium.


ZachrielForever:
LGTM

PR #5486, opened 2026-08-19
feat(brh): make `current` output prettier

ykshek:


ykshek:
Before:
image
After
image

PR #5522, opened 2026-08-21
chore(Portal): Give HDMI 2.1 toggle another pass

### LINKED ISSUE #5070: Blur my Shell extension on bazzite-gnome now has a dependency
### Describe the bug

When opening a popup in gnome, with the Blur my Shell extension enabled, popups have square corners, instead of rounded

Image

The blur from blur my shell should be going around the corners cleanly, offering the normal gnome rounded corners.

### Session Logs

```shell
https://paste.centos.org/view/20bd3d79
```

### Hardware

[Hardware](https://paste.centos.org/view/7b99dabe)
This should be applicable to any Gnome image though

The Blur my Shell extension, which is included by default in bazzite, now has a [dependency](https://github.com/kancko/gnome-rounded-blur), which allows Blur my Shell to blur the corners of applications/popups : 

This library is not provided by fedora packaging, so would need to be packaged in copr or built from source

EPOCHvoyager:
Revise phrasing for the HDMI 2.1 toggle. Use imperative for the starting verb, and be specific about kernel arguments applying on boot; also use “induce”, as opposed to “cause”, given that the flickering isn't contingent on HDMI 2.1 itself, but on current issues. 

From what I've read on the [freedesktop issue](https://gitlab.freedesktop.org/drm/amd/-/work_items/5649), flickering is dependent on VRR being enabled — use “and/or”. Finally, also be specific about instability applying to the video signal.

### DIFF
--- a/system_files/desktop/shared/usr/share/yafti/yafti.yml
+++ b/system_files/desktop/shared/usr/share/yafti/yafti.yml
@@ -680,8 +680,8 @@ screens:
             label: "Disable Autoupgrading FSR"
             script: "ujust global-fsr4 disable"
       - id: "configure-amd-hdmi21"
-        title: "Enable HDMI 2.1 for AMD Graphics Cards"
-        description: "Sets a karg to unlock additional functionality on AMD graphics cards. However, this may also cause flickering and instability on some displays."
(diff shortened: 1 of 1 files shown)


andy10115:
LGTM

PR #5524, opened 2026-08-21
fix(deck): persist steamosctl login mode

Smalls1652:
This PR changes how the `bazzite-autologin` script works, so that it doesn't override what the user sets the default login mode to with `steamosctl set-default-login-mode`.

Currently, when `bazzite-autologin` runs at every startup it will:

1. Delete `/etc/sddm.conf.d/zz-holo-autologin.conf` (Along with the old config files).
2. Evaluate what the default session should be based off it's checks.
3. Writes/overwrites the file at `/etc/sddm.conf.d/zz-bazzite-autologin.conf` with that determined session and the username of the user (With the `1000` UID).

This change will have `bazzite-autologin`:

1. Delete the old config files.
2. Evaluate what the default session should be based off it's checks. 
3. Writes/overwrites the file at `/etc/sddm.conf.d/zz-bazzite-autologin.conf` with the username of the user (With the `1000` UID).
4. Write the file at `/etc/sddm.conf.d/zz-holo-autologin.conf`, **if it does not exist**, with the determined default session.

This will allow users to set their preferred login mode with `steamosctl` and have it persist across reboots.

### DIFF
--- a/system_files/deck/shared/usr/libexec/bazzite-autologin
+++ b/system_files/deck/shared/usr/libexec/bazzite-autologin
@@ -5,19 +5,19 @@ BASE_IMAGE_NAME=$(jq -r '."base-image-name"' < $IMAGE_INFO)
 
 USER=$(id -nu 1000)
 
-BAZZITE_CONF='/etc/sddm.conf.d/zz-bazzite-autologin.conf'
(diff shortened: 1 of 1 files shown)


KyleGospo:
Thank you, looks great!

KyleGospo (approved):

Issue #5529, opened 2026-08-21
ujust update:  '/var/lib/flatpak/exports/share' is not in the search path


ourichermath:
### Describe the bug

With Bazzite v44.20260820, ujust update is being executed as sudo.  I'm now getting the following when it hits the flatpak portion:
```
Note that '/var/lib/flatpak/exports/share' is not in the search path
set by the XDG_DATA_HOME and XDG_DATA_DIRS
environment variables, so applications may not
be able to find it until you set them. The
directories currently searched are:

- /root/.local/share
- /usr/local/share/
- /usr/share/
```
But, if I do:

`echo $XDG_DATA_DIRS`

I get:

`/home/dave/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share:/usr/share`

IOW, the path is there.  According to what I can find online, it looks like that's a result of not all environment variable being able to be passed with sudo.  I have no idea if that's correct or what the solution might be.  Perhaps return ujust update to non-sudo?

### What did you expect to happen?

I was expecting not to see a path issue pop up when doing ujust update.

### Session Logs

```shell
System Information:
https://paste.centos.org/view/9935ff35

Logs This Boot:
https://paste.centos.org/view/978a4f27
```

### Hardware

Operating System: Bazzite 44
KDE Plasma Version: 6.7.4
KDE Frameworks Version: 6.29.0
Qt Version: 6.11.1
Kernel Version: 7.2.0-ogc4.1.fc44.x86_64 (64-bit)
Graphics Platform: Wayland
Processors: 16 × AMD Ryzen 7 5800X3D 8-Core Processor
Memory: 16 GiB of RAM (15.5 GiB usable)
Graphics Processor: AMD Radeon RX 9070 XT
Manufacturer: Gigabyte Technology Co., Ltd.
Product Name: B550 VISION D-P
System Version: -CF

ourichermath:
For whatever reason, ujust update updated me to 44.20260824.  Since I've got it, I ran ujust update again to see what happened.  All I got was:

`uupd needs to be invoked as root.`

So, it looks like removing sudo doesn't work as expected.

ourichermath:
The revised fix for this is in 44.20260825.  It looks good so far.

Should I close this?  Or, does it get closed automatically because it's in a Stable release?

PR #5566, opened 2026-08-23
Fix: removed sudo to fix issue-5529

HuntedRaven7:
In [issue-5529](https://github.com/ublue-os/bazzite/issues/5529) they report that

"With Bazzite v44.20260820, ujust update is being executed as sudo. I'm now getting the following when it hits the flatpak portion:

```
Note that '/var/lib/flatpak/exports/share' is not in the search path
set by the XDG_DATA_HOME and XDG_DATA_DIRS
environment variables, so applications may not
be able to find it until you set them. The
directories currently searched are:

- /root/.local/share
- /usr/local/share/
- /usr/share/
```
"

Removed sudo from ujust update in system_files/desktop/shared/usr/share/ublue-os/just/10-update.just.
uupd handles its own privilege escalation internally, so running it via sudo was unnecessary and caused environment variables like XDG_DATA_DIRS to be stripped, triggering the Flatpak path warning.

### DIFF
--- a/system_files/desktop/shared/usr/share/ublue-os/just/10-update.just
+++ b/system_files/desktop/shared/usr/share/ublue-os/just/10-update.just
@@ -6,7 +6,7 @@ alias upgrade := update
 [group("system")]
 update:
     #!/usr/bin/bash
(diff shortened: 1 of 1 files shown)


KyleGospo:
Thanks much!

Issue #5551, opened 2026-08-22
"Get Media Apps" menu in Bazzite Portal is confusing


TeaDrinkingProgrammer:
### Describe the bug

When I click the button to install Netflix in the Bazzite Portal under "Get Media Apps", a terminal flashes quickly.

### What did you expect to happen?

I didn't know exactly what to expect, maybe that it installs some kind of Netflix app or a shortcut. Turns out it adds a shortcut to _steam_ but I didn't see this at first because I am using Desktop mode, so that is not the first thing that comes to mind.

I also expected some kind of feedback along the lines of "shortcut added to Steam", instead of a terminal flash.

### Session Logs

```shell
N.A
```

### Hardware

Operating System: Bazzite 44 (Desktop Mode)
KDE Plasma Version: 6.7.4
KDE Frameworks Version: 6.29.0
Qt Version: 6.11.1
Kernel Version: 7.2.0-ogc4.1.fc44.x86_64 (64-bit)
Graphics Platform: Wayland

PR #5567, opened 2026-08-23
fix(portal): explain media app Steam shortcuts

nileshchakraborty:
## What

Clarifies that Bazzite Portal media apps add shortcuts to Steam and shows a desktop notification only after a shortcut is created successfully.

## Why

Addresses #5551, where the terminal flashes and users do not know what happened after selecting Netflix or another service.

The recipe now checks `steamos-add-to-steam` directly:

- success: show the generic Steam-library notification and success message
- failure: show an error, return non-zero, and do not display a success notification

## Validation

- `git diff --check`
- Yafti YAML remains unchanged
- extracted web-app recipe passes `bash -n`
- Docker `just --unstable --fmt --check` passes
- stubbed success path confirms notification and success output
- stubbed failure path confirms non-zero status, error output, and no false notification
- no host or immutable image changes made


nileshchakraborty:
Added explicit success/failure handling around `steamos-add-to-steam`. A failed add now returns non-zero and cannot show the notification or success message. Bash syntax, Docker just formatting, and stubbed success/failure paths pass.

PR #5575, opened 2026-08-24
chore(deck): add announcements

ykshek:
completely forgot to do this for the release :sadge:

Better late than never I suppose


ykshek:
also snuck in a fix for the other announcement message in the same commit

technically should be separated into another commit but oh well

PR #5587, opened 2026-08-25
fix(just): replace uupd with bazzite-updater with jq parsing

numero1024:
In the current state, ujust update fails because it lacks privileges 
this fix makes use of bazzite-updater and correctly parses the JSON with jq
Co-authored-by: biebel <badk666@hotmail.com>


numero1024:
#5592 yeah this is not great UX

PR #5589, opened 2026-08-25
fix: replace outdated tap reference in bazzite-cli

ToffeeFox:
bbrew is now a part of homebrew-core and no longer needs to be tapped from external.

Remove tap for `valkyrie00/bbrew`
Correct reference to `valkyrie00/bbrew/bbrew` to simply `bbrew`

This lets bazzite-cli be enabled without having to explicitly trust the valkyrie00/bbrew tap
(More accurately, this change is *required* to enable bazzite-cli at this time)


ToffeeFox:
Need to fix my commit message to comply with contribution guidelines
Will be resubmitting shortly

PR #5603, opened 2026-08-26
fix: get brew image like akmods and bump hash

Lumaeris:
Renovate treats ARGs as normal text which made Bazzite continuosly get 6 month old image of Brew OCI image missing out newest changes (and Brew itself ofc).


KyleGospo:
Thank you :)

PR #5795, opened 2026-09-12
feat(power): dynamic game-aware Wi-Fi power saving, TDP restore fix, and Strix Point support

nileshchakraborty:
## Motivation & Context

This pull request addresses several critical power management, TDP restoration, hardware detection, and Wi-Fi latency issues across Bazzite handhelds and desktop images:

1. **Wi-Fi Latency Spikes & Stuttering ([#5760](https://github.com/ublue-os/bazzite/issues/5760))**: Default 802.11 power saving causes periodic 50–200 ms latency spikes, packet jitter, and audio/video micro-stuttering during Moonlight, Sunshine, Chiaki, Steam Remote Play, and online multiplayer.
2. **Battery Life vs. Low-Latency Balance**: Disabling Wi-Fi power saving unconditionally reduces battery life when browsing or on the go. Users need a dynamic, zero-configuration solution that provides ultra-low latency during gaming/streaming and on AC power, while conserving battery when idle or when in Power-Saver mode.
3. **`bazzite-tdpfix` Restoration Syntax Error**: When restoring default TDP (>45W), `bazzite-tdpfix` contained a bash command substitution bug (`"$(cat "${POWER_CAP_PATH}_default")" > ...`) that attempted to execute the TDP integer as a command, causing command-not-found errors and leaving the GPU stuck at 15W. In addition, TDP limits and permissions were not reapplied on wake from sleep.
4. **Strix Point APU Handheld Hardware Detection ([#5788](https://github.com/ublue-os/bazzite/issues/5788))**: Missing DMI vendor/board entries in `powerstation-hardware` prevented proper power station management on Strix Point handhelds.

---

#### 1. Dynamic Power-Setting & Game-Aware Wi-Fi Power Save Auto-Switcher
- **Script**: `/usr/libexec/bazzite-wifi-game-powersave`
- **Unit**: `/usr/lib/systemd/system/bazzite-wifi-game-powersave.service` (enabled by default in `Containerfile`)
- **Behavior**:
  - Dynamically inspects running processes, power profiles (`power-profiles-daemon` / CPU EPP), and AC/Battery power source every 3 seconds with negligible overhead (<1.5 MB memory, 0% CPU).
  - Automatically disables 802.11 power saving (`iw dev <iface> set power_save off`) during gameplay/streaming, on AC power, or when Performance profile is selected.
  - Automatically restores power saving (`on`) when running on battery in desktop/idle mode or when Power-Saver profile is active.
[...]

PR #5811, opened 2026-09-13
fix(input): suppress input-remapper autostart when service is disabled

nileshchakraborty:
### Problem
When Bazzite disables `input-remapper.service` by default in favor of `inputplumber.service`, systemd's XDG autostart generator still processes `/etc/xdg/autostart/input-remapper-autoload.desktop`. On user session login, `app-input-remapper-autoload@autostart.service` executes `input-remapper-control --autoload`, immediately fails because the service is inactive, and flags a persistent failed user unit (`systemctl --user --failed`).

### Solution
1. In `Containerfile`, add `Hidden=true` to `/etc/xdg/autostart/input-remapper-autoload.desktop` so the XDG autostart generator skips it by default (matching the existing `NoDisplay=true` treatment on `/usr/share/applications/input-remapper-gtk.desktop`).
2. Update `ujust restore-input-remapper` across desktop and deck justfiles to copy and unhide `input-remapper-autoload.desktop` into `~/.config/autostart` when enabled, and remove it on disable.

### Validation
- Verified on ROG Xbox Ally X on Bazzite 44.
- Confirmed `systemd --user` no longer generates a failing service on login, and `ujust restore-input-remapper enable` correctly unhides both the GTK menu entry and the autostart hook.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -703,6 +703,7 @@ RUN --mount=type=cache,dst=/var/cache \
     ; fi && \
     printf "[scx]\nscx_service = \"scx_loader.service\"\n" >> /usr/share/steamos-manager/platform.toml && \
     sed -i 's@\[Desktop Entry\]@\[Desktop Entry\]\nNoDisplay=true@g' /usr/share/applications/input-remapper-gtk.desktop && \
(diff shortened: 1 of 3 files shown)


KyleGospo (approved):

nileshchakraborty:
@KyleGospo could you please review this [pr too](https://github.com/ublue-os/bazzite/pull/5795), this fixes a major wifi latency bug

PR #5830, opened 2026-09-15
fix(hwsupport): detect SuiPlay0X1 as a handheld

archlinuxbits:
SuiPlay0X1 is an Ayaneo 2S rebadge (vendor: Mysten Labs, Inc.).

InputPlumber already has 50-suiplay_0x1.yaml, but bazzite-hardware-setup treats it as an HTPC and masks inputplumber because non-valve-handheld-hardware returns false.

This adds the DMI product/vendor so InputPlumber stays enabled.

DMI from the device:
product_name: SuiPlay0X1
sys_vendor: Mysten Labs, Inc.
valve-hardware: exit 1
non-valve-handheld-hardware: exit 1


KyleGospo:
Great fix thanks, this will land on testing soon if you want to try it out.

PR #5831, opened 2026-09-15
feat: enable default user account creation on deck images

Zeglius:
Create an account with UID 1000 when one is not pressent at boot up


system_files/desktop/shared/usr/libexec/hwsupport/powerstation-hardware
```diff
@@ -17,14 +17,19 @@ if [[ "$PRODUCT_NAME" == *AOKZOE* ]]; then
 	exit 0
 fi
 
+# Mysten / SuiPlay (Ayaneo 2S rebadge)
+if [[ "$sys_vendor_lc" == *"mysten"* ]] || [[ "$PRODUCT_NAME" == "SuiPlay0X1" ]]; then
```
orowith2os, line 21: why is SuiPlay0X1 listed both here and down below? Shouldn't it hit either case, so you only need one?

system_files/deck/shared/usr/lib/systemd/system/bazzite-create-default-user.service
```diff
@@ -0,0 +1,14 @@
+# Create default user if none with UID 1000 exists already.
+
+[Unit]
+Description=Create Default User
+After=systemd-sysusers.service
```
orowith2os, line 5: This should run on/after `systemd-firstboot`, which runs when `/etc/machine-id` doesn't exist AFAIK. Does need a double check to see if that case is actually met on the first boot of a Bazzite system. Otherwise it'll run on every boot and, on systems without a user of ID 1000 (for example, on my system running a dx-gnome image with systemd-homed managed users) it'll make a user that, intentionally, should not exist.

PR #5836, opened 2026-09-16
feat(CEC): Add CEC input mapping for deck-based images

addmix:
Leverages the existing CEC control mappings in InputPlumber to allow UI navigation through CEC. https://github.com/ShadowBlip/InputPlumber/blob/main/rootfs/usr/share/inputplumber/capability_maps/cec.yaml

I have marked this PR as a draft because I would like feedback about unmasking/enabling InputPlumber for all deck images. The CEC control mappings are only active when CEC is configured with a valid connection to a CEC device.


addmix:
Following the discussion here: https://discord.com/channels/1072614816579063828/1237926270151688192/1549606753937788950

I'm going to remove the draft label from this PR, as it doesn't seem like there are any known issues with unmasked InputPlumber on HTPC builds, where only Steam Deck hardware needs InputPlumber to be disabled/masked.

Issue #5746, opened 2026-09-07
Gaming Mode / Desktop Mode switching broken after updating from 44.20260820 to 44.20260902


miikkikiikki:
### Describe the bug

After updating Bazzite Deck GNOME to 44.20260902, switching between GNOME Desktop Mode and Gaming Mode became unreliable/broken.

My setup uses GNOME through GDM as the normal desktop session with the GDM login screen, and SDDM/Gamescope for Gaming Mode.

This setup worked correctly on previous Bazzite 44 versions.

After updating to 44.20260902, switching from Desktop Mode to Gaming Mode often fails. The graphical session closes, but instead of entering Gaming Mode, the system returns to the Desktop/login session.

Sometimes it gets into a loop where I cannot properly enter Gaming Mode or return to the expected Desktop session. In some cases, only a couple of reboots allowed me to recover and get back into the correct mode.

I initially suspected my configuration, but I tested older deployments on exactly the same machine:

- 44.20260902 — broken
- 44.20260820 — works correctly
- 43.20260420 — works correctly

After rebasing from 44.20260902 to 44.20260820, Desktop ↔ Gaming Mode switching immediately started working normally again.

### Additional HDMI issue

I also noticed an HDMI regression after the same update.

When entering Gaming Mode with my TV connected over HDMI, the TV sometimes loses the HDMI signal and reports "No signal".

This problem also disappears after reverting to 44.20260820 and does not occur on 43.20260420.

This may be related to the existing HDMI/AMDGPU regression reported in #5691, although I do not know whether the HDMI problem and the session-switching problem have the same root cause.

### What did you expect to happen?

Switching from GNOME Desktop Mode to Gaming Mode should reliably start the Gamescope/Steam Gaming Mode session.

Switching back should reliably return to GNOME Desktop Mode.

HDMI output to the TV should continue working when Gaming Mode starts.

### Session Logs

```shell
Unable to provide logs from the broken deployment at the moment because I reverted to 44.20260820, which immediately resolves the issue
```

### Hardware

Image: bazzite-deck-gnome
Working version: 44.20260820
Broken version: 44.20260902
Also tested working: 43.20260420
GPU: AMD Radeon RX 9070 XT
CPU: AMD Ryzen 7 7700
Motherboard: Gigabyte B650M AORUS ELITE AX ICE
Desktop environment: GNOME
Gaming Mode: Gamescope / Steam

### Extra information or context

[...]

Issue #5691, opened 2026-09-01
No HDMI output on boot since 44.20260825 update — amdgpu REG_WAIT timeout in optc314_disable_crtc


smildren:
### Describe the bug

After updating from Bazzite 43.20260420 to 44.20260825, HDMI output to my TV fails completely on every boot. The Bazzite splash screen displays correctly, then the display briefly changes resolution, the TV reports a lost/no signal (with an audible beep from the TV at the moment of drop), and the system continues booting completely headless (SSH still works fine — this is a display-only failure).

I've since tested the latest available update (as of 2026-09-02, ostree checksum starting ea6c61f3...) and the exact same failure occurs. Rolling back to 43.20260420 (`sudo rpm-ostree rollback`) immediately and reliably fixes it every time — same hardware, same TV, same cable, only the OS version differs.

Versions:
- Working: 43.20260420 (digest sha256:d63c82406dfb8c0dbd451b579150175fad27a39f9a8354811d00906b61853b8)
- Broken: 44.20260825 (digest sha256:035fbf36e555ac596bad1f5260b60981e0ea0d9d12f7731625028536cea6cb1b)
- Also broken: latest available update as of 2026-09-02 (ostree checksum ea6c61f32cea8ee4b8a121b5ef6d01635c0377bae7786d3b0b29418af35434be)

Steps to reproduce:
1. Cold boot the machine with the TV on and connected via HDMI.
2. Bazzite splash screen appears normally on the TV.
3. Just as the boot hands off toward the desktop/gamescope session, the display briefly changes resolution and the TV loses signal entirely ("No Source" / "HDMI no signal").
4. System continues booting headless.

Key log line (journalctl -b 0 -k):
amdgpu 0000:c4:00.0: [drm] REG_WAIT timeout 1us * 100000 tries - optc314_disable_crtc line:146

Surrounding context:
amdgpu 0000:c4:00.0: [drm] HDMI-A-1: PSR support 0, DC PSR ver -1, sink PSR ver 0 DPCD caps 0x0 su_y_granularity 0
amdgpu 0000:c4:00.0: [drm] REG_WAIT timeout 1us * 100000 tries - optc314_disable_crtc line:146

What I've already tried (none of it prevents the drop):
- Full cold power cycle (power off, unplugged 60s, reseated HDMI) — no change.
[...]

PR #5862, opened 2026-09-19
fix(gamescope): prefer steamosctl in steam session recovery

vorosdev:
`sessions.d/steam` defines a `short_session_recover()` that calls`steamos-session-select desktop` after 5 consecutive `gamescope-wl`crashes. On Bazzite that wrapper exists but delegates to `/usr/libexec/os-session-select`, whose case statement does not recognize `desktop` and exits with `Unrecognized session 'desktop'` and exit 1.
The recovery function then silently no-ops and `gamescope-session-plus` keeps retrying the broken compositor in
an infinite loop (the only escape is `rpm-ostree rollback`).

The fix adds the use of `steamosctl switch-to-desktop-mode` with Fallback to the `steamos-session-select` wrapper (deprecated).

## Changes in `sessions.d/steam`

```diff
-    steamos-session-select desktop
+    if command -v steamosctl >/dev/null; then steamosctl switch-to-desktop-mode; else steamos-session-select desktop; fi
```

This does not fix the underlying gamescope problem, but it prevents the user from getting stuck on a black screen by dropping
them to desktop mode. Tested in the version testing (`testing-44.20260919.2`,`sha256:59af8aac7b051732d0ba335fb637f1197ad1dbea724c0c3f22120b4574c23b5e`).

## Related

- #5691
- #5746

## Test plan

1. Build an image with the patch on a machine where `gamescope-wl` reproducibly crashes on session start with the
   ogui-steam client.
2. Confirm the compositor retries ~5 times and then drops to KDE.


KyleGospo:
Thanks, this one's already fixed upstream and will be pushed shortly

PR #5864, opened 2026-09-20
fix: skip image verification MOTD check in non-interactive shells

Ricket:
`verify_motd.sh` currently runs `rpm-ostree status --json` during non-interactive shell startup. In restricted environments such as Codex's sandbox, this produces errors even when the requested command is unrelated to system status:

> error: Loading sysroot: Connecting to system bus: Could not connect: Operation not permitted

Add an early return for non-interactive shells. This avoids the unnecessary query for automated commands while preserving the image-verification warning in interactive terminals.

Validation on Bazzite:
- `bash -n` passed.
- Sandboxed shell startup completed without the errors.
- Stubbed checks confirmed that non-interactive shells skip the query and interactive shells still invoke it.
- Also manual-checked that if I add an echo below this case statement, I see the echo when I open an interactive terminal in Konsole.


KyleGospo:
Thanks!

KyleGospo (approved):

PR #5866, opened 2026-09-20
fix(cardwire): allow non-admin users via DBus policy override

vorosdev:
On bazzite-deck systems, the gamescope-session-plus session fails to launch the Wayland compositor and Steam when the active user is not a member of `wheel` or `sudo` (typical for single-user installs). The `cardwire` binary rejects the compositor's DBus calls with `AccessDenied` because the upstream `cardwire 0.12+` package ships a restrictive DBus policy that only allows `root`, `sudo`, and `wheel`. gamescope-session-plus counts this as `ogui-steam failed` and after 5 retries executes `steamosctl switch-to-desktop-mode`, dropping the user into the Plasma/KDE session instead of Gaming Mode.

Reported in [#5746](https://github.com/ublue-os/bazzite/issues/5746) after the 44.20260820 → 44.20260902 update.

## Fix

Adds an override at `system_files/deck/shared/etc/dbus-1/system.d/org.opengamingcollective.cardwire.bazzite.conf` that grants `<policy context="default">` with `send_destination` and `receive_sender` to `org.opengamingcollective.cardwire`. The file lands under `/etc/dbus-1/system.d/` via the already-existing `COPY system_files/deck/shared system_files/deck/${
BASE_IMAGE_NAME} /` directive in the Containerfile, so it survives `rpm-ostree upgrade` of the upstream `cardwire` package.

## Tested on

- Image: `ostree-image-signed:docker://ghcr.io/ublue-os/bazzite-deck:stable`
- Digest: `sha256:050572c864322f567a223741922f70932d2a888f34f6b839f12e06ebd8e08f64`
- Version: `44.20260919 (2026-09-19T20:09:54Z)`


KyleGospo:
Thanks!

